PR #106, opened 2022-07-05
created client package to abstract client related code

miladz68:


xlab (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum and @ysv)

wojtek-coreum (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 5 unresolved discussions (waiting on @miladz68 and @ysv)

___
*[`pkg/client/context.go` line 9 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3yzYElr-p4o9pZof:-N6H3yzYElr-p4o9pZog:b-p2cyio) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L9)):*
> ```Go
> 
> // NewClientContext returns a new cosmos client context
> func NewClientContext(opts ...Option) client.Context {
> ```

Could the `opts` magic be removed?
- it's not used anywhere
- if some args are needed I would prefer to add them explicitly as arguments rather than magic functions doing magic things

It also adds weirdness because by using this syntax `opts` are optional, but it seems that setting chain ID and rpc client is required to make it working, so they are not so optional.
___
*[`pkg/client/context.go` line 11 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H4FVT9W4LsiLGI3Vb:-N6H4FVT9W4LsiLGI3Vc:b-4gdjfh) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L11)):*
> ```Go
> func NewClientContext(opts ...Option) client.Context {
> 	encodingConfig := NewEncodingConfig()
> 	ctx := client.Context{}.
> ```

Minor: Wouldn't it be better to set fields on `client.Context` directly instead of calling `With*` functions? Those functions add more magic without giving any benefits.
___
*[`pkg/client/context.go` line 15 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H2kTD2RFmbyNTDXEH:-N6H2kTD2RFmbyNTDXEI:b-6znjk) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L15)):*
> ```Go
> 		WithInterfaceRegistry(encodingConfig.InterfaceRegistry).
> 		WithTxConfig(encodingConfig.TxConfig).
> 		WithLegacyAmino(encodingConfig.Amino)
> ```

This legacy amino should not be used by any module now. If something still uses this codec it's better to panic. Could you remove it?
We also have a piece of code in `app.go` supporting legacy amino which should be removed too, but that's different story.
___
*[`pkg/client/encoding.go` line 10 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3A_d9HFCTktYVHR4:-N6H3A_d9HFCTktYVHR5:boe05c4) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/encoding.go#L10)):*
> ```Go
> // NewEncodingConfig returns the encoding config
> func NewEncodingConfig() cosmoscmd.EncodingConfig {
> 	return cosmoscmd.MakeEncodingConfig(app.ModuleBasics)
> ```

Is separate function needed for this one-liner?
___
*[`pkg/config/genesis.go` line 52 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3WUTEEm5V4RmA4vm:-N6H3WUU79ZUdRLYGHJD:bmghf7c) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/config/genesis.go#L52)):*
> ```Go
> // TokenSymbol returns the governance token symbol. This is different
> // for each network(i.e mainnet, testnet, etc)
> func (g *Genesis) TokenSymbol() string {
> ```

What's the point of having this function?
Genesis is always produced from Network:

```
n.Genesis()
```

so why do you need to do sth like this:

```
n.Genesis().TokenSymbol()
```

instead of taking it directly from the network:

```
n.TokenSymbol()
```

I don't see a reason to copy the data around

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`pkg/client/context.go` line 9 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3yzYElr-p4o9pZof:-N6HBUNr3A_yJ_poMZPt:b-fufsva) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L9)):*

Could the `opts` magic be removed?
- it's not used anywhere
- if some args are needed I would prefer to add them explicitly as arguments rather than magic functions doing magic things

It also adds weirdness because by using this syntax `opts` are optional, but it seems that setting chain ID and rpc client is required to make it working, so they are not so optional.
</blockquote></details>

removed.
actually chain ID and rpc client are not needed eveywhere as I have used client context without them in the current pr.
___
*[`pkg/client/context.go` line 11 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H4FVT9W4LsiLGI3Vb:-N6HByV291ux1gwZmn8q:bs8h84p) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L11)):*

Minor: Wouldn't it be better to set fields on `client.Context` directly instead of calling `With*` functions? Those functions add more magic without giving any benefits.
</blockquote></details>

removed opts.
___
*[`pkg/client/context.go` line 15 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H2kTD2RFmbyNTDXEH:-N6HDTxz2dRd03z-weL5:bcetswc) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L15)):*

This legacy amino should not be used by any module now. If something still uses this codec it's better to panic. Could you remove it?
We also have a piece of code in `app.go` supporting legacy amino which should be removed too, but that's different story.
</blockquote></details>

I am only following the cosmoscmd code, I am not sure that all legacy amino is removed from all the modules. how can we know for sure that none of the module are using legacy amino ?
___
*[`pkg/client/encoding.go` line 10 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3A_d9HFCTktYVHR4:-N6HDlBLECND8a5P918U:b-ckt0jn) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/encoding.go#L10)):*

Is separate function needed for this one-liner?
</blockquote></details>

It is a wrapper to provide abstraction
encoding config will be used in different places and people have to remember to pull 2 different packages and combine them to get an encoding config. It is not about number of lines but mental over head (remembering stuff).
Also changes will go to a single place in future. 
___
*[`pkg/config/genesis.go` line 52 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3WUTEEm5V4RmA4vm:-N6HECVL8GUDZ2FUO_WF:b-efit9h) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/config/genesis.go#L52)):*

What's the point of having this function?
Genesis is always produced from Network:

```
n.Genesis()
```

so why do you need to do sth like this:

```
n.Genesis().TokenSymbol()
```

instead of taking it directly from the network:

```
n.TokenSymbol()
```

I don't see a reason to copy the data around
</blockquote></details>

there are functions in crust which have access to genesis but not to network. one solution is to pass network there, but why do that when genesis knows the token symbol and can return it.

wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 4 unresolved discussions (waiting on @miladz68, @xlab, and @ysv)

___
*[`pkg/client/context.go` line 11 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H4FVT9W4LsiLGI3Vb:-N6HLZun4Z0h9eQNPUxr:b-3fet37) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L11)):*

removed opts.
</blockquote></details>

I meant `With*` functions used in the lines below. You could set fields of `client.Context` directly
___
*[`pkg/client/context.go` line 15 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H2kTD2RFmbyNTDXEH:-N6HLqmaD9LlehuPN9bS:b-rdsfgh) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L15)):*

I am only following the cosmoscmd code, I am not sure that all legacy amino is removed from all the modules. how can we know for sure that none of the module are using legacy amino ?
</blockquote></details>

https://docs.cosmos.network/main/core/encoding.html

Now I see that:
```
Note, this process of using Protocol Buffers over Amino is still an ongoing process.
```

@xlab do you know if their standard modules still use amino?

Anyway, I'm changing this to non-blocking.
___
*[`pkg/client/context.go` line 8 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6HOvVU4wOkC0dm7vOt:-N6HOvVV5aO_pT5pluCU:b-l0qhzu) ([raw file](https://github.com/coreumfoundation/coreum/blob/5965b8c77ac1d1ea441a344dc63cd32c0e3e6aee/pkg/client/context.go#L8)):*
> ```Go
> 
> // NewDefaultClientContext returns a new cosmos client context
> func NewDefaultClientContext() client.Context {
> ```

I know that this is how cosmos sdk names it, but for me this is weird that a config object which can't be really used for connecting to some external service has `client` in it's name. This config object is used only for offline operations: signing transactions, generating genesis. Could you find a better name maybe? We don't need to use the exact same function to generate config for offline and online operations.

It's a shame that they created one single global container for everything. And it's a game of errors to find what fields must be set for particular operation (like signing) to succeed without panics.
___
*[`pkg/config/genesis.go` line 52 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3WUTEEm5V4RmA4vm:-N6HN-nA1asoltg2BFJn:b-u99kq8) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/config/genesis.go#L52)):*

there are functions in crust which have access to genesis but not to network. one solution is to pass network there, but why do that when genesis knows the token symbol and can return it.
</blockquote></details>

It looks weird to me that `TokenSymbol` is the property of `Genesis`. I would say that if sth in crust needs to know token symbol it should receive it directly as a parameter (or part of some config object) or receive entire `Network` as parameter.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 4 unresolved discussions (waiting on @miladz68, @xlab, and @ysv)

___
*[`pkg/client/context.go` line 11 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H4FVT9W4LsiLGI3Vb:-N6HYsRCBg1pYnPCR9c7:bnul9fl) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L11)):*

there will be no way to pass config to this function, the only way that I can think about is via options pattern, other wise it will be a mess.
</blockquote></details>

I don't know if you get my point. Instead of doing `client.Context{}.WithCodec(...)` you could do `client.Context{Codec: ..., JSONCodec: ...}`. Isn't it simpler and more straightforward? I'm not blocking on this so if you don't care much then ignore it.
___
*[`pkg/client/context.go` line 8 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6HOvVU4wOkC0dm7vOt:-N6HY4Pi1AO2XfAbORaK:b-nejkq7) ([raw file](https://github.com/coreumfoundation/coreum/blob/5965b8c77ac1d1ea441a344dc63cd32c0e3e6aee/pkg/client/context.go#L8)):*

I agree, but we get what we get.
</blockquote></details>

Yep, but I mean that the config returned by this function could be used exclusively for offline operations, like the ones required to prepare genesis. It doesn't need to support remote calls. At the moment you use it to prepare gentx transactions. In that case I believe the name should be changed to not contain "Client" word because it's not used in the context where client connects to server.
___
*[`pkg/config/genesis.go` line 52 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3WUTEEm5V4RmA4vm:-N6HZcm66mWuftvDo4JC:bb8lwr8) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/config/genesis.go#L52)):*

we use token symbol to generate genesis, so it is not that weird to me tbh.
</blockquote></details>

IMO `Genesis` should be just a type used to generate `genesis.json`, not a config container - this is what the original idea was. In my idea `Network` was the only source of network-related config. @ysv what do you think

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 4 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`pkg/client/context.go` line 11 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H4FVT9W4LsiLGI3Vb:-N6HbCMh2r5HRkDvpc6U:bsjy8hz) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L11)):*

I don't know if you get my point. Instead of doing `client.Context{}.WithCodec(...)` you could do `client.Context{Codec: ..., JSONCodec: ...}`. Isn't it simpler and more straightforward? I'm not blocking on this so if you don't care much then ignore it.
</blockquote></details>

I don't think it would be better. the `withs` approach is designed for a reason, they might contains some checks or sanitizatins. 
___
*[`pkg/client/context.go` line 8 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6HOvVU4wOkC0dm7vOt:-N6HaYAu2SX1XhA75bfb:bgjpp4w) ([raw file](https://github.com/coreumfoundation/coreum/blob/5965b8c77ac1d1ea441a344dc63cd32c0e3e6aee/pkg/client/context.go#L8)):*

Yep, but I mean that the config returned by this function could be used exclusively for offline operations, like the ones required to prepare genesis. It doesn't need to support remote calls. At the moment you use it to prepare gentx transactions. In that case I believe the name should be changed to not contain "Client" word because it's not used in the context where client connects to server.
</blockquote></details>

I don't think we should add another abstraction, I am just following cosmos sdk way of doing things, I put all the common code in this function and the caller can add rpc client to this when they need.
another way would be to always have rpc client as input and let caller pas nil if they don't need rpc.
another way also would be to use options which you disagreed with.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 2 unresolved discussions (waiting on @xlab and @ysv)

___
*[`pkg/config/genesis.go` line 52 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3WUTEEm5V4RmA4vm:-N6HgApLEJh9MEwXf-EA:bu3e10i) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/config/genesis.go#L52)):*
Adding more to the discussion:

> There are functions in crust which have access to genesis

In case of this code: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L49
I proposed in another PR that this code should be changed to set appropriate stuff (accounts, validators) on `Network` object, and then genesis should be derived from the network.

In case of this code: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L108
ChainID should be returned from `Network` object, not `Genesis`

In case of this code: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L119
Wallet should be added to `Network`, not `Genesis`

And this line: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L119
Could be changed to sth like `c.network.Genesis().Save(homeDir)`

By making all those changes cored wrapper could work on the `Network` type, not `Genesis` which is better IMO

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 2 unresolved discussions (waiting on @xlab and @ysv)

___
*[`pkg/config/genesis.go` line 52 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3WUTEEm5V4RmA4vm:-N6Hhjv1DkB4J7uJrKBq:b-hpljko) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/config/genesis.go#L52)):*

Adding more to the discussion:

> There are functions in crust which have access to genesis

In case of this code: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L49
I proposed in another PR that this code should be changed to set appropriate stuff (accounts, validators) on `Network` object, and then genesis should be derived from the network.

In case of this code: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L108
ChainID should be returned from `Network` object, not `Genesis`

In case of this code: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L119
Wallet should be added to `Network`, not `Genesis`

And this line: https://github.com/CoreumFoundation/crust/blob/master/infra/apps/cored/cored.go#L119
Could be changed to sth like `c.network.Genesis().Save(homeDir)`

By making all those changes cored wrapper could work on the `Network` type, not `Genesis` which is better IMO



</blockquote></details>

And to be honest after making all those changes I'm not sure if `Genesis` type would still be required. All it's methods could be moved to `Network`

xlab (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 2 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/client/context.go` line 15 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H2kTD2RFmbyNTDXEH:-N6I17jUD-k8Dpc8XjNm:b-y7vxdj) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L15)):*

https://docs.cosmos.network/main/core/encoding.html

Now I see that:
```
Note, this process of using Protocol Buffers over Amino is still an ongoing process.
```

@xlab do you know if their standard modules still use amino?

Anyway, I'm changing this to non-blocking.
</blockquote></details>

Yes, amino is used for signing

xlab:
*[`pkg/client/context.go` line 15 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H2kTD2RFmbyNTDXEH:-N6I2AEN7XX0uXlpU3vN:blp4a72) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/client/context.go#L15)):*
<details><summary><i>Previously, xlab (Max Kupriianov) wrote…</i></summary><blockquote>

Yes, amino is used for signing
</blockquote></details>

I think the need for this is reduced to almost none in the latest years, so if it doesn't break, it's safe to remove.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`pkg/config/genesis.go` line 52 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6H3WUTEEm5V4RmA4vm:-N6NN181CiBb4mRQGAco:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b3e6a7d8d1639b9b63176868d919a1fdd8374eb8/pkg/config/genesis.go#L52)):*

**Please don't take my comment as granted because I might not have enough context & understanding on this topic but my opinion is**:

It makes sense for me to include Genesis inside Network & use Network to initialize Cored app. I believe that genesis is mostly about genesis.json & its generation while network could contain other config on top of this related to devnet/testnet/mainnet etc (I think it is applicable). 

As for tokenSumbol it makes sense to have it in Network. As for Genesis I would also keep tokenSumbol there only if it is realted direct genesis functionality (eg. required to generate genesis.json) to avoid using Network from Genesis. I would still avoid having Genesis.TokenSymbol func exposed I believe there is no use case for it from outside except Network which already does expose.

> And to be honest after making all those changes I'm not sure if Genesis type would still be required. All it's methods could be moved to Network

I still think it is required but It might become internal thingy where all Genesis methods are exposed via Network e.g:

```
func (n *Network) FundAccount(...) {
  n.Genesis.funcAccount(...)
}
```
</blockquote></details>

Done.

wojtek-coreum (changes requested):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 8 unresolved discussions (waiting on @miladz68)

___
*[`pkg/config/genesis_test.go` line 101 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NLflZ03W2IXduQ6Lu:-N6NLflZ03W2IXduQ6Lv:b9qmwke) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/genesis_test.go#L101)):*
> ```Go
> 	accountAddress2 := sdk.AccAddress(key2.Address())
> 
> 	requireT.Len(n.fundedAccounts, 2)
> ```

I would say it's better to create separate network for all the unit tests (that network should not be part of `networksList` slice, just a `var` declared somewhere in this file). Otherwise you will have to update all of this once we e.g. decide to add real accounts to devnet to support faucet. 
___
*[`pkg/config/genesis_test.go` line 188 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NNCpy2eL_ubcjzClt:-N6NNCpz5O-lBflLVLXg:bjngbtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/genesis_test.go#L188)):*
> ```Go
> }
> 
> func TestNetworkFundedAccountsNotMutable(t *testing.T) {
> ```

Same test is needed for `genTxs`, I think both might be checked in the same unit test
___
*[`pkg/config/networks.go` line 43 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NNNDM2EAlpaez0zFP:-N6NNNDM2EAlpaez0zFQ:b5wx20g) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L43)):*
> ```Go
> func init() {
> 	networksList := []Network{
> 		New(
> ```

I would prefer to have public fields in the `Network` struct and set them directly here instead of using positional arguments of constructor. To keep funded accounts and gentx protected it could be done like this:

```go
type NetworkConfig struct {
  // same fields which are in Network - just public
}

func New(nConfig NetworkConfig) Network {
  // copy data from public fields of NetworkConfig to private fields of Network
}

New(NetworkConfig{
  ChainID: 
  GenesisTimeUTC: 
  // ...
})
```
___
*[`pkg/config/networks.go` line 76 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NN0dR3w1zF8olKEju:-N6NN0dR3w1zF8olKEjv:butlab1) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L76)):*
> ```Go
> 	fundedAccounts []FundedAccount
> 	genTxs         []json.RawMessage
> 	mu             *sync.Mutex
> ```

move `mu` above fields which need protection like

```
... some fields ...

mu
fundedAccounts
genTxs
```
___
*[`pkg/config/networks.go` line 157 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NOpbg4OWnhnafimcS:-N6NOpbh9QCpoG5G1mQk:bjbsmtk) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L157)):*
> ```Go
> 	genutilState := genutiltypes.GetGenesisStateFromAppState(codec, appState)
> 	bankState := banktypes.GetGenesisStateFromAppState(codec, appState)
> ```

you may move `n.mu.Lock(); defer n.mu.Unlock()` here
___
*[`pkg/config/networks.go` line 225 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NPBRm73yj9EpJaYY3:-N6NPBRm73yj9EpJaYY4:byi4k59) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L225)):*
> ```Go
> 
> // ChainID returns the chain ID used in network config
> func (n Network) ChainID() string {
> ```

why not returning `ChainID` type here?
___
*[`pkg/config/tx.go` line 18 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NPS1T12ZReb5-zbW_:-N6NPS1U7lD9ylmaBtpy:bh2nv2x) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/tx.go#L18)):*
> ```Go
> 
> // GenerateAddValidatorTx generates transaction of type MsgCreateValidator
> func GenerateAddValidatorTx(
> ```

Could you make it compatible with naming schema of `PrepareTxBankSend` used by `znet`?

The idea is like this:
- `Prepare` is a constant prefix
- the rest is taken from CLI interface of client, for transfer CLI looks like this: `cored tx bank send [arguments]` that's why method is named `TxBankSend`

Please check how CLI looks for transaction adding a validator
___
*[`pkg/config/tx.go` line 58 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NQafjEUihraXpqHkq:-N6NQafjEUihraXpqHkr:b-r37ti7) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/tx.go#L58)):*
> ```Go
> 	privKey := &cosmossecp256k1.PrivKey{Key: signerKey}
> 	txBuilder := clientCtx.TxConfig.NewTxBuilder()
> 	txBuilder.SetGasLimit(200000)
> ```

Please keep in mind there is a PR https://reviewable.io/reviews/coreumfoundation/crust/13
adding `txBuilder.SetMemo(memo)` here and redoing other parts a bit.

You may copy that logic here.

BTW: please integrate new location of the code into `znet` asap because we might eventually go out of sync

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

___
*[`pkg/config/genesis_test.go` line 101 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NLflZ03W2IXduQ6Lu:-N6NYgys6E0flr3YHQf9:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/genesis_test.go#L101)):*

I would say it's better to create separate network for all the unit tests (that network should not be part of `networksList` slice, just a `var` declared somewhere in this file). Otherwise you will have to update all of this once we e.g. decide to add real accounts to devnet to support faucet. 
</blockquote></details>

Done.
___
*[`pkg/config/genesis_test.go` line 188 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NNCpy2eL_ubcjzClt:-N6NZ5Wo8JwW-sYMi-ui:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/genesis_test.go#L188)):*

Same test is needed for `genTxs`, I think both might be checked in the same unit test
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 76 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NN0dR3w1zF8olKEju:-N6NbZsc6Ff4WPrYyQlb:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L76)):*

move `mu` above fields which need protection like

```
... some fields ...

mu
fundedAccounts
genTxs
```
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 157 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NOpbg4OWnhnafimcS:-N6NbhsYCL_FR6YQ5TWV:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L157)):*

you may move `n.mu.Lock(); defer n.mu.Unlock()` here
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 225 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NPBRm73yj9EpJaYY3:-N6NbnyN0V1nKu0n3hiV:b-a2ujfl) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L225)):*

why not returning `ChainID` type here?
</blockquote></details>

Changed.
___
*[`pkg/config/tx.go` line 18 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NPS1T12ZReb5-zbW_:-N6NcPRP1UX1GvP4nOpa:byagd4e) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/tx.go#L18)):*

Could you make it compatible with naming schema of `PrepareTxBankSend` used by `znet`?

The idea is like this:
- `Prepare` is a constant prefix
- the rest is taken from CLI interface of client, for transfer CLI looks like this: `cored tx bank send [arguments]` that's why method is named `TxBankSend`

Please check how CLI looks for transaction adding a validator
</blockquote></details>

rename to `PrepareTxStakingCreateValidator`
___
*[`pkg/config/tx.go` line 58 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NQafjEUihraXpqHkq:-N6NcZAQAuZy6YRBAfH_:b-r6rbd1) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/tx.go#L58)):*

Please keep in mind there is a PR https://reviewable.io/reviews/coreumfoundation/crust/13
adding `txBuilder.SetMemo(memo)` here and redoing other parts a bit.

You may copy that logic here.

BTW: please integrate new location of the code into `znet` asap because we might eventually go out of sync
</blockquote></details>

will do this in the next PR where I integrate this into znet

ysv (changes requested):
> Reviewed 1 of 5 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: 3 of 6 files reviewed, 10 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

___
*[`pkg/config/genesis_test.go` line 10 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NdLOHCJ4vKkY6cKvY:-N6NdLOHCJ4vKkY6cKvZ:bb34qv4) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/genesis_test.go#L10)):*
> ```Go
> 
> 	"github.com/CoreumFoundation/coreum/app"
> 	"github.com/CoreumFoundation/coreum/pkg/client"
> ```

so the file should be named network_test.go now since logic we test is in network ?
___
*[`pkg/config/networks.go` line 43 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NNNDM2EAlpaez0zFP:-N6Ncs8UC5vZbPksUIgg:b-ouvb5c) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L43)):*

I would prefer to have public fields in the `Network` struct and set them directly here instead of using positional arguments of constructor. To keep funded accounts and gentx protected it could be done like this:

```go
type NetworkConfig struct {
  // same fields which are in Network - just public
}

func New(nConfig NetworkConfig) Network {
  // copy data from public fields of NetworkConfig to private fields of Network
}

New(NetworkConfig{
  ChainID: 
  GenesisTimeUTC: 
  // ...
})
```
</blockquote></details>

for such a complex initialization I agree it is better
___
*[`pkg/config/networks.go` line 130 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NdXTUCDNSEKfkl9rm:-N6NdXTUCDNSEKfkl9rn:b-j36sdi) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L130)):*
> ```Go
> 
> // EncodeGenesis returns the json encoded representation of the genesis file
> func (n Network) EncodeGenesis() ([]byte, error) {
> ```

minor: I don't mind having it implemented as it is done now just this method seems to have become too long. Is there any possibility to spliti it by simple refactoring ?
___
*[`pkg/config/networks.go` line 132 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6Nc4KzBh5IUdOMkyin:-N6Nc4KzBh5IUdOMkyio:b-ahdg0a) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L132)):*
> ```Go
> func (n Network) EncodeGenesis() ([]byte, error) {
> 	n.mu.Lock()
> 	defer n.mu.Unlock()
> ```

minor: I prefer to have newline after mutex lock+defer unlock
just mentioned it a few times

wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 8 unresolved discussions (waiting on @miladz68)

___
*[`pkg/config/genesis_test.go` line 197 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NkYQ3D-zW7TA6V2iG:-N6NkYQ3D-zW7TA6V2iH:b5l9r3c) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/genesis_test.go#L197)):*
> ```Go
> 	requireT := require.New(t)
> 
> 	n, err := NetworkByChainID(Devnet)
> ```

what about `testNetwork()` here?
___
*[`pkg/config/networks.go` line 79 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NlLtrBzy7oUnhRCJC:-N6NlLtrBzy7oUnhRCJD:b-bqwgbg) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L79)):*
> ```Go
> 	addressPrefix  string
> 	tokenSymbol    string
> 	mu             *sync.Mutex
> ```

and empty line above `mu` please
___
*[`pkg/config/networks.go` line 86 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NlSXhDhihq4tJUtKu:-N6NlSXhDhihq4tJUtKv:bc21wnl) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L86)):*
> ```Go
> // New returns a new instance of Network
> func New(
> 	c NetworkConfig,
> ```

Empty lines around this one are not needed anymore
___
*[`pkg/config/networks.go` line 93 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6Nmht22YwLvXB7iV2C:-N6Nmht22YwLvXB7iV2D:b-v8557q) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L93)):*
> ```Go
> 		addressPrefix:  c.AddressPrefix,
> 		tokenSymbol:    c.TokenSymbol,
> 		fundedAccounts: c.FundedAccounts,
> ```

`fundedAccounts` and `genTxs` should be copied instead of just being assigned.
Otherwise if you create two`Network` objects from the same config and then add new funded account to one of them, it will be added to the second network too.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 11 unresolved discussions (waiting on @miladz68)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6R4xjBAHclFEdaM5yn:-N6R4xjC0X_2Va6DeRCy:bwr2eab) (no related file):*
After sleeping, I've got some ideas how this code could be simplified. I mentioned in another comment that funded accounts and gentxs must be copied in `New` function. It means that you would neeed to do it twice (in `New` and in `NetworkByChainID`). This is a possible solution:

1. Change type of `networksList` from `[]Network` to `[]NetworkConfig` because methods defined by `Network` type are not needed there:

```go
	networksList := []NetworkConfig{
		{
			ChainID:       Mainnet,
			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
			AddressPrefix: "core",
			TokenSymbol:   TokenSymbolMain,
		},
		{
			ChainID:       Devnet,
			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
			AddressPrefix: "devcore",
			TokenSymbol:   TokenSymbolDev,
		},
	}
```

2. Also change `var networks = map[ChainID]Network{}` to `var networks = map[ChainID]NetworkConfig{}`
3. Then `NetworkByChainID` could be as simple as

```go
func NetworkByChainID(id ChainID) (Network, error) {
	config, found := networks[id]
	if !found {
		return Network{}, errors.Errorf("chainID %s not found", id)
	}

	return New(config), nil
}
```

By doing it this way, copying data (funded accounts and gentxs) might be done only once inside `New`.


4. Because both `Network` and `NetworkConfig` contain the same list of fields, instead of reassigning them I would store config inside `Network`, changing this:


```go
type Network struct {
	chainID        ChainID
	genesisTime    time.Time
	addressPrefix  string
	tokenSymbol    string

	mu             *sync.Mutex
	fundedAccounts []FundedAccount
	genTxs         []json.RawMessage
}
```

to this:

```go
type Network struct {
	mu      *sync.Mutex
	config NetworkConfig
}
```

and then all the methods of `Network` could read and modify fields of `nw.config`.

5. And then, `New` could be implemented like this:

```go
func New(c NetworkConfig) Network {
        c.FundedAccounts = append([]FundedAccount{}, c.FundedAccounts...)
        c.GenTxs = append([]json.RawMessage{}, c.GenTxs...)
	return Network{
		mu:      &sync.Mutex{},
                config: c,
	}
}
```
```
```


___
___
*[`pkg/config/networks.go` line 240 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6R7hck2E8rS626cfUZ:-N6R7hcl2Q_jo2bqYfWx:b-ffzy9s) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L240)):*
> ```Go
> 	nw, found := networks[id]
> 	if !found {
> 		return Network{}, errors.Errorf("chainID %s not found", nw.chainID)
> ```

`nw` was not found so `nw.chainID` --> `id`
___
*[`pkg/config/networks.go` line 243 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6R7GbB5yK25TEjhN_O:-N6R7GbB5yK25TEjhN_P:b-m8w4io) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L243)):*
> ```Go
> 	}
> 
> 	return nw, nil
> ```

This function should return copy of the network to make changes made inside funded accounts and gentxs local to the caller

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 12 unresolved discussions (waiting on @miladz68)

___
*[`pkg/config/genesis_test.go` line 32 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6RzGND6dv6V_3VkpPH:-N6RzGNEEI1I8otg-QnG:bf0d7u1) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/genesis_test.go#L32)):*
> ```Go
> 
> func testNetwork() Network {
> 	return New(NetworkConfig{
> ```

I think this test network config should contain one funded account and one gentx to verify in tests that items are correctly moved from `NetworkConfig` object to the `Network` returned by `New`.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 13 unresolved discussions (waiting on @miladz68)

___
*[`pkg/config/genesis_test.go` line 10 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6T2vCm0PldbXQCZek1:-N6T2vCm0PldbXQCZek2:bsob9xq) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/genesis_test.go#L10)):*
> ```Go
> 	"time"
> 
> 	"github.com/CoreumFoundation/coreum/app"
> ```

I started using this code in my another branch and just found that because `app` is imported here, it's not possible to import `config` there to use `Network` type in the constructor of the `cored` app.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 11 unresolved discussions (waiting on @wojtek-coreum and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6R4xjBAHclFEdaM5yn:-N6gJV_b6KpNL4gNkOhq:bwt42v8) (no related file):*

After sleeping, I've got some ideas how this code could be simplified. I mentioned in another comment that funded accounts and gentxs must be copied in `New` function. It means that you would neeed to do it twice (in `New` and in `NetworkByChainID`). This is a possible solution:

1. Change type of `networksList` from `[]Network` to `[]NetworkConfig` because methods defined by `Network` type are not needed there:

```go
	networksList := []NetworkConfig{
		{
			ChainID:       Mainnet,
			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
			AddressPrefix: "core",
			TokenSymbol:   TokenSymbolMain,
		},
		{
			ChainID:       Devnet,
			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
			AddressPrefix: "devcore",
			TokenSymbol:   TokenSymbolDev,
		},
	}
```

2. Also change `var networks = map[ChainID]Network{}` to `var networks = map[ChainID]NetworkConfig{}`
3. Then `NetworkByChainID` could be as simple as

```go
func NetworkByChainID(id ChainID) (Network, error) {
	config, found := networks[id]
	if !found {
		return Network{}, errors.Errorf("chainID %s not found", id)
	}

	return New(config), nil
}
```

By doing it this way, copying data (funded accounts and gentxs) might be done only once inside `New`.


4. Because both `Network` and `NetworkConfig` contain the same list of fields, instead of reassigning them I would store config inside `Network`, changing this:


```go
type Network struct {
	chainID        ChainID
	genesisTime    time.Time
	addressPrefix  string
	tokenSymbol    string

	mu             *sync.Mutex
	fundedAccounts []FundedAccount
	genTxs         []json.RawMessage
}
```

to this:

```go
type Network struct {
	mu      *sync.Mutex
	config NetworkConfig
}
```

and then all the methods of `Network` could read and modify fields of `nw.config`.

5. And then, `New` could be implemented like this:

```go
func New(c NetworkConfig) Network {
        c.FundedAccounts = append([]FundedAccount{}, c.FundedAccounts...)
        c.GenTxs = append([]json.RawMessage{}, c.GenTxs...)
	return Network{
		mu:      &sync.Mutex{},
                config: c,
	}
}
```
```
```

</blockquote></details>

I like your first suggestion (bullet points 1-3). 
I don't like the second suggestion (bullet points 4-5). I believe that network config and network are different concepts and although network config is a subset of network now, it might not so in future. I think we should copy fileds, and although one might argue that copying is error prone when we add new fields, I think it is better to keep the two concerns separate.

___
___
*[`pkg/config/genesis_test.go` line 10 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NdLOHCJ4vKkY6cKvY:-N6gBRfRBbhNahaTeroo:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/genesis_test.go#L10)):*

so the file should be named network_test.go now since logic we test is in network ?
</blockquote></details>

Done.
___
*[`pkg/config/genesis_test.go` line 10 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6T2vCm0PldbXQCZek1:-N6gI9EuEXALUwyc3lkw:b-w7ghrr) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/genesis_test.go#L10)):*

I started using this code in my another branch and just found that because `app` is imported here, it's not possible to import `config` there to use `Network` type in the constructor of the `cored` app.
</blockquote></details>

I need more context to resolve this. 
we can either pass this PR and attend to the problem in another PR or you could give me the link to you PR so I can try to fix it.
___
*[`pkg/config/genesis_test.go` line 32 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6RzGND6dv6V_3VkpPH:-N6gEv6OArX_rAJWO_g4:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/genesis_test.go#L32)):*

I think this test network config should contain one funded account and one gentx to verify in tests that items are correctly moved from `NetworkConfig` object to the `Network` returned by `New`.
</blockquote></details>

Done.
___
*[`pkg/config/genesis_test.go` line 197 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NkYQ3D-zW7TA6V2iG:-N6NwplK4ngYp1qlW9Ew:bx315f3) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/genesis_test.go#L197)):*

what about `testNetwork()` here?
</blockquote></details>

we are testing `NetworkByChainID` itslef here. 
___
*[`pkg/config/networks.go` line 130 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NdXTUCDNSEKfkl9rm:-N6gGcfTAI_L2nytsz25:b-sogoup) ([raw file](https://github.com/coreumfoundation/coreum/blob/e3de288eebe7108856fab2ea12a18b9a97f45215/pkg/config/networks.go#L130)):*

minor: I don't mind having it implemented as it is done now just this method seems to have become too long. Is there any possibility to spliti it by simple refactoring ?
</blockquote></details>

reduced the function lines from 80 to 59. more reduction will reduce redablitiy and we will have to re-introduce genesis type as an abstraction.
___
*[`pkg/config/networks.go` line 79 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NlLtrBzy7oUnhRCJC:-N6Nx35g1ml2MFRr9rgW:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L79)):*

and empty line above `mu` please
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 86 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6NlSXhDhihq4tJUtKu:-N6NxH-c8CPkC55aqiv2:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L86)):*

Empty lines around this one are not needed anymore
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 93 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6Nmht22YwLvXB7iV2C:-N6gH98K9hmzOciqF5uF:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L93)):*

`fundedAccounts` and `genTxs` should be copied instead of just being assigned.
Otherwise if you create two`Network` objects from the same config and then add new funded account to one of them, it will be added to the second network too.
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 240 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6R7hck2E8rS626cfUZ:-N6gH2RsBiXsSSX6Bel8:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L240)):*

`nw` was not found so `nw.chainID` --> `id`
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 243 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6R7GbB5yK25TEjhN_O:-N6gH28l2g6mLHLitWjS:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/networks.go#L243)):*

This function should return copy of the network to make changes made inside funded accounts and gentxs local to the caller
</blockquote></details>

Done.

wojtek-coreum (changes requested):
> Reviewed 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 8 unresolved discussions (waiting on @miladz68 and @ysv)

___
*[`pkg/config/networks.go` line 95 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6lvDeaD4RFprNzSwfa:-N6lvDebBxZqpbTf2NwH:b-icxh2q) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/networks.go#L95)):*
> ```Go
> 		mu:            &sync.Mutex{},
> 	}
> 	n.fundedAccounts = append(n.fundedAccounts, c.FundedAccounts...)
> ```

you could do it inside `Network{}`:
```go
Network {
    // ...
    fundedAccounts: append([]fundedAccount{}, c.FundedAccounts...)
}
```

same thing for next line
___
*[`pkg/config/networks.go` line 151 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6lvyaJ4u1Znl68QTNW:-N6lvyaJ4u1Znl68QTNX:b-d3mr1f) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/networks.go#L151)):*
> ```Go
> 	)
> 	bankState.Supply = bankState.Supply.Add(coins...)
> 	return accountState, bankState, nil
> ```

`bankState` is passed by pointer to this function so it doesn't need to be returned because it is already modified
___
*[`pkg/config/genesis_test.go` line 10 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6T2vCm0PldbXQCZek1:-N6ltex6E2EQnV2KEjR_:b-5f86cg) ([raw file](https://github.com/coreumfoundation/coreum/blob/b01c93ab09bdec70e9be1bc710f70db4421bf2e8/pkg/config/genesis_test.go#L10)):*

I need more context to resolve this. 
we can either pass this PR and attend to the problem in another PR or you could give me the link to you PR so I can try to fix it.
</blockquote></details>

Take a look at this line in my branch: https://github.com/CoreumFoundation/coreum/blob/deterministic-gas/app/app.go#L548
This import: https://github.com/CoreumFoundation/coreum/blob/deterministic-gas/app/app.go#L99
causes circular dependency.
___
*[`pkg/config/network_test.go` line 145 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6luJKh2-ivvT3DHFwl:-N6luJKh2-ivvT3DHFwm:b-kzt57a) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/network_test.go#L145)):*
> ```Go
> 	requireT.NoError(err)
> 
> 	assertT.Subset(state.Bank.Balances, []balance{
> ```

You don't check `requireT.Len(state.Bank.Balances, 3)`
___
*[`pkg/config/network_test.go` line 145 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6luTzNCuHPeHSWt5sm:-N6luTzNCuHPeHSWt5sn:b-3dapps) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/network_test.go#L145)):*
> ```Go
> 	requireT.NoError(err)
> 
> 	assertT.Subset(state.Bank.Balances, []balance{
> ```

Could you test it for the third one too?
___
*[`pkg/config/network_test.go` line 165 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6luX2jA8MtQL4Ao44K:-N6luX2jA8MtQL4Ao44L:b-3dapps) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/network_test.go#L165)):*
> ```Go
> 	)
> 	requireT.Len(state.Auth.Accounts, 3)
> 	assertT.Subset(state.Auth.Accounts, []account{
> ```

Could you test it for the third one too?

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: 4 of 6 files reviewed, 7 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/config/networks.go` line 95 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6lvDeaD4RFprNzSwfa:-N6mCkmS6NeFc4LGCtAP:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/networks.go#L95)):*

you could do it inside `Network{}`:
```go
Network {
    // ...
    fundedAccounts: append([]fundedAccount{}, c.FundedAccounts...)
}
```

same thing for next line
</blockquote></details>

Done.
___
*[`pkg/config/networks.go` line 151 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6lvyaJ4u1Znl68QTNW:-N6mD6VN2IGO5Vodi-4l:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/networks.go#L151)):*

`bankState` is passed by pointer to this function so it doesn't need to be returned because it is already modified
</blockquote></details>

Done.
___
*[`pkg/config/network_test.go` line 145 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6luJKh2-ivvT3DHFwl:-N6m7V0t7ZGM1yVOTr3z:bfxoqcr) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/network_test.go#L145)):*

You don't check `requireT.Len(state.Bank.Balances, 3)`
</blockquote></details>

I would say it is not needed. the tests adds 2 balances and checks to see those 2 balances are there. In other words this test should avoid making assumptions about how many entries were in testConfig if possible.
___
*[`pkg/config/network_test.go` line 145 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6luTzNCuHPeHSWt5sm:-N6mBWUD4e7tZKqPjvxv:bx9jflf) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/network_test.go#L145)):*

Could you test it for the third one too?
</blockquote></details>

provided an explanation above.
___
*[`pkg/config/network_test.go` line 165 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6luX2jA8MtQL4Ao44K:-N6mBduo-E-FFp62PAik:bcarmmo) ([raw file](https://github.com/coreumfoundation/coreum/blob/290dc3318ac5459fe3cc3182227cfaa3e8cf6a66/pkg/config/network_test.go#L165)):*

Could you test it for the third one too?
</blockquote></details>

provided explanation above.

wojtek-coreum (changes requested):
> Reviewed 10 of 10 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 2 unresolved discussions (waiting on @miladz68 and @ysv)

___
*[`app/network.go` line 7 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6nAfLnD2BLkKUVkSKE:-N6nAfLnD2BLkKUVkSKF:b-rzc7ay) ([raw file](https://github.com/coreumfoundation/coreum/blob/14caf2f7d64089f549e06c861e02d8a522f1ff31/app/network.go#L7)):*
> ```Go
> 	_ "embed"
> 	"encoding/json"
> 	"html/template"
> ```

use `text/template` instead as we are not dealing with html

wojtek-coreum (approved):
:lgtm:

> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106#-:-N6qPkhj0gm9CFmsptkh:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @ysv)

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 1 unresolved discussion (waiting on @ysv)

___
*[`app/network.go` line 7 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/106#-N6nAfLnD2BLkKUVkSKE:-N6qPlRd-5jBlDRn6pES:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/14caf2f7d64089f549e06c861e02d8a522f1ff31/app/network.go#L7)):*

use `text/template` instead as we are not dealing with html
</blockquote></details>

Done.

wojtek-coreum (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: all files reviewed, 1 unresolved discussion (waiting on @ysv)

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/106)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68)

PR #109, opened 2022-07-13
Refuse to add transaction to mempool if gas price is below minimum specified for network

wojtek-coreum:
By default, in cosmos sdk, each node independently define the minimum gas price for accepted transactions.
In this PR per-network, node-independent, global minimum gas price is introduced by modifying ante handler.

- standard ante handler contains `MempoolFeeDecorator` decorator which is responsible for checking if fee provided with transaction is above the level required by the node
- in this PR ante handler constructor and `MempoolFeeDecorator` are copied
- `MempoolFeeDecorator` is reimplemented (`x/auth/ante/fee.go`) to check if fee provided by transaction meets requirement defined by network
- ante handler constructor (`x/auth/ante/ante.go`) is adjusted to inject the modified decorator instead of the default one.

---
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -20,7 +20,6 @@ import (
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/version"
 	"github.com/cosmos/cosmos-sdk/x/auth"
-	"github.com/cosmos/cosmos-sdk/x/auth/ante"
 	authrest "github.com/cosmos/cosmos-sdk/x/auth/client/rest"
 	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
 	authsims "github.com/cosmos/cosmos-sdk/x/auth/simulation"
@@ -84,21 +83,20 @@ import (
 	ibcporttypes "github.com/cosmos/ibc-go/v3/modules/core/05-port/types"
 	ibchost "github.com/cosmos/ibc-go/v3/modules/core/24-host"
 	ibckeeper "github.com/cosmos/ibc-go/v3/modules/core/keeper"
+	"github.com/ignite-hq/cli/ignite/pkg/cosmoscmd"
+	"github.com/ignite-hq/cli/ignite/pkg/openapiconsole"
 	"github.com/spf13/cast"
+	monitoringp "github.com/tendermint/spn/x/monitoringp"
+	monitoringpkeeper "github.com/tendermint/spn/x/monitoringp/keeper"
+	monitoringptypes "github.com/tendermint/spn/x/monitoringp/types"
 	abci "github.com/tendermint/tendermint/abci/types"
 	tmjson "github.com/tendermint/tendermint/libs/json"
 	"github.com/tendermint/tendermint/libs/log"
 	tmos "github.com/tendermint/tendermint/libs/os"
 	dbm "github.com/tendermint/tm-db"
 
-	"github.com/ignite-hq/cli/ignite/pkg/cosmoscmd"
-	"github.com/ignite-hq/cli/ignite/pkg/openapiconsole"
-
-	monitoringp "github.com/tendermint/spn/x/monitoringp"
-	monitoringpkeeper "github.com/tendermint/spn/x/monitoringp/keeper"
-	monitoringptypes "github.com/tendermint/spn/x/monitoringp/types"
-
 	"github.com/CoreumFoundation/coreum/docs"
+	"github.com/CoreumFoundation/coreum/x/auth/ante"
 	// this line is used by starport scaffolding # stargate/app/moduleImport
 )
 
@@ -545,7 +543,7 @@ func New(
 			BankKeeper:      app.BankKeeper,
 			SignModeHandler: encodingConfig.TxConfig.SignModeHandler(),
 			FeegrantKeeper:  app.FeeGrantKeeper,
-			SigGasConsumer:  ante.DefaultSigVerificationGasConsumer,
+			MinGasPrice:     sdk.NewCoin(DefaultNetwork.TokenSymbol(), sdk.NewIntFromBigInt(DefaultNetwork.MinGasPrice())),
 		},
 	)
 	if err != nil {
--- a/app/app.go
+++ b/app/app.go
@@ -543,7 +543,7 @@ func New(
 			BankKeeper:      app.BankKeeper,
 			SignModeHandler: encodingConfig.TxConfig.SignModeHandler(),
 			FeegrantKeeper:  app.FeeGrantKeeper,
-			MinGasPrice:     sdk.NewCoin(DefaultNetwork.TokenSymbol(), sdk.NewIntFromBigInt(DefaultNetwork.MinGasPrice())),
+			MinGasPrice:     sdk.NewCoin(DefaultNetwork.TokenSymbol(), sdk.NewIntFromBigInt(DefaultNetwork.MinDiscountedGasPrice())),
 		},
 	)
 	if err != nil {
--- a/app/network.go
+++ b/app/network.go
@@ -54,9 +54,9 @@ func init() {
 			AddressPrefix: "core",
 			TokenSymbol:   TokenSymbolMain,
 			Fees: FeeConfig{
-				InitialGasPrice: big.NewInt(1500),
-				MinGasPrice:     big.NewInt(1000),
-				TxBankSendGas:   120000,
+				InitialGasPrice:       big.NewInt(1500),
+				MinDiscountedGasPrice: big.NewInt(1000),
+				TxBankSendGas:         120000,
 			},
 		},
 		{
@@ -65,9 +65,9 @@ func init() {
 			AddressPrefix: "devcore",
 			TokenSymbol:   TokenSymbolDev,
 			Fees: FeeConfig{
-				InitialGasPrice: big.NewInt(1500),
-				MinGasPrice:     big.NewInt(1000),
-				TxBankSendGas:   120000,
+				InitialGasPrice:       big.NewInt(1500),
+				MinDiscountedGasPrice: big.NewInt(1000),
+				TxBankSendGas:         120000,
 			},
 		},
 	}
@@ -87,9 +87,9 @@ var networks = map[ChainID]NetworkConfig{}
 
 // FeeConfig is the part of network config defining parameters of our fee model
 type FeeConfig struct {
-	InitialGasPrice *big.Int
-	MinGasPrice     *big.Int
-	TxBankSendGas   uint64
+	InitialGasPrice       *big.Int
+	MinDiscountedGasPrice *big.Int
+	TxBankSendGas         uint64
 }
 
 // NetworkConfig helps initialize Network instance
@@ -105,13 +105,13 @@ type NetworkConfig struct {
 
 // Network holds all the configuration for different predefined networks
 type Network struct {
-	chainID         ChainID
-	genesisTime     time.Time
-	addressPrefix   string
-	tokenSymbol     string
-	initialGasPrice *big.Int
-	minGasPrice     *big.Int
-	txBankSendGas   uint64
+	chainID               ChainID
+	genesisTime           time.Time
+	addressPrefix         string
+	tokenSymbol           string
+	initialGasPrice       *big.Int
+	minDiscountedGasPrice *big.Int
+	txBankSendGas         uint64
 
 	mu             *sync.Mutex
 	fundedAccounts []FundedAccount
@@ -121,16 +121,16 @@ type Network struct {
 // NewNetwork returns a new instance of Network
 func NewNetwork(c NetworkConfig) Network {
 	n := Network{
-		genesisTime:     c.GenesisTime,
-		chainID:         c.ChainID,
-		addressPrefix:   c.AddressPrefix,
-		tokenSymbol:     c.TokenSymbol,
-		initialGasPrice: big.NewInt(0).Set(c.Fees.InitialGasPrice),
-		minGasPrice:     big.NewInt(0).Set(c.Fees.MinGasPrice),
-		txBankSendGas:   c.Fees.TxBankSendGas,
-		mu:              &sync.Mutex{},
-		fundedAccounts:  append([]FundedAccount{}, c.FundedAccounts...),
-		genTxs:          append([]json.RawMessage{}, c.GenTxs...),
+		genesisTime:           c.GenesisTime,
+		chainID:               c.ChainID,
+		addressPrefix:         c.AddressPrefix,
+		tokenSymbol:           c.TokenSymbol,
+		initialGasPrice:       big.NewInt(0).Set(c.Fees.InitialGasPrice),
+		minDiscountedGasPrice: big.NewInt(0).Set(c.Fees.MinDiscountedGasPrice),
+		txBankSendGas:         c.Fees.TxBankSendGas,
+		mu:                    &sync.Mutex{},
+		fundedAccounts:        append([]FundedAccount{}, c.FundedAccounts...),
+		genTxs:                append([]json.RawMessage{}, c.GenTxs...),
 	}
 
 	return n
@@ -292,9 +292,9 @@ func (n Network) InitialGasPrice() *big.Int {
 	return big.NewInt(0).Set(n.initialGasPrice)
 }
 
-// MinGasPrice returns minimum gas price after giving maximum discount
-func (n Network) MinGasPrice() *big.Int {
-	return big.NewInt(0).Set(n.minGasPrice)
+// MinDiscountedGasPrice returns minimum gas price after giving maximum discount
+func (n Network) MinDiscountedGasPrice() *big.Int {
+	return big.NewInt(0).Set(n.minDiscountedGasPrice)
 }
 
 // TxBankSendGas returns gas required by tx bank send
--- a/app/network.go
+++ b/app/network.go
@@ -53,10 +53,12 @@ func init() {
 			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
 			AddressPrefix: "core",
 			TokenSymbol:   TokenSymbolMain,
-			Fees: FeeConfig{
+			Fee: FeeConfig{
 				InitialGasPrice:       big.NewInt(1500),
 				MinDiscountedGasPrice: big.NewInt(1000),
-				TxBankSendGas:         120000,
+				DeterministicGas: DeterministicGasConfig{
+					BankSend: 120000,
+				},
 			},
 		},
 		{
@@ -64,10 +66,12 @@ func init() {
 			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
 			AddressPrefix: "devcore",
 			TokenSymbol:   TokenSymbolDev,
-			Fees: FeeConfig{
+			Fee: FeeConfig{
 				InitialGasPrice:       big.NewInt(1500),
 				MinDiscountedGasPrice: big.NewInt(1000),
-				TxBankSendGas:         120000,
+				DeterministicGas: DeterministicGasConfig{
+					BankSend: 120000,
+				},
 			},
 		},
 	}
@@ -85,11 +89,16 @@ func init() {
 
 var networks = map[ChainID]NetworkConfig{}
 
+// DeterministicGasConfig keeps config about deterministic gas for some message types
+type DeterministicGasConfig struct {
+	BankSend uint64
+}
+
 // FeeConfig is the part of network config defining parameters of our fee model
 type FeeConfig struct {
 	InitialGasPrice       *big.Int
 	MinDiscountedGasPrice *big.Int
-	TxBankSendGas         uint64
+	DeterministicGas      DeterministicGasConfig
 }
 
 // NetworkConfig helps initialize Network instance
@@ -98,20 +107,18 @@ type NetworkConfig struct {
 	GenesisTime    time.Time
 	AddressPrefix  string
 	TokenSymbol    string
-	Fees           FeeConfig
+	Fee            FeeConfig
 	FundedAccounts []FundedAccount
 	GenTxs         []json.RawMessage
 }
 
 // Network holds all the configuration for different predefined networks
 type Network struct {
-	chainID               ChainID
-	genesisTime           time.Time
-	addressPrefix         string
-	tokenSymbol           string
-	initialGasPrice       *big.Int
-	minDiscountedGasPrice *big.Int
-	txBankSendGas         uint64
+	chainID       ChainID
+	genesisTime   time.Time
+	addressPrefix string
+	tokenSymbol   string
+	fee           FeeConfig
 
 	mu             *sync.Mutex
 	fundedAccounts []FundedAccount
@@ -120,17 +127,18 @@ type Network struct {
 
 // NewNetwork returns a new instance of Network
 func NewNetwork(c NetworkConfig) Network {
+	fee := c.Fee
+	fee.InitialGasPrice = big.NewInt(0).Set(c.Fee.InitialGasPrice)
+	fee.MinDiscountedGasPrice = big.NewInt(0).Set(c.Fee.MinDiscountedGasPrice)
 	n := Network{
-		genesisTime:           c.GenesisTime,
-		chainID:               c.ChainID,
-		addressPrefix:         c.AddressPrefix,
-		tokenSymbol:           c.TokenSymbol,
-		initialGasPrice:       big.NewInt(0).Set(c.Fees.InitialGasPrice),
-		minDiscountedGasPrice: big.NewInt(0).Set(c.Fees.MinDiscountedGasPrice),
-		txBankSendGas:         c.Fees.TxBankSendGas,
-		mu:                    &sync.Mutex{},
-		fundedAccounts:        append([]FundedAccount{}, c.FundedAccounts...),
-		genTxs:                append([]json.RawMessage{}, c.GenTxs...),
+		genesisTime:    c.GenesisTime,
+		chainID:        c.ChainID,
+		addressPrefix:  c.AddressPrefix,
+		tokenSymbol:    c.TokenSymbol,
+		fee:            fee,
+		mu:             &sync.Mutex{},
+		fundedAccounts: append([]FundedAccount{}, c.FundedAccounts...),
+		genTxs:         append([]json.RawMessage{}, c.GenTxs...),
 	}
 
 	return n
@@ -289,17 +297,17 @@ func (n Network) TokenSymbol() string {
 
 // InitialGasPrice returns initial gas price used by the first block
 func (n Network) InitialGasPrice() *big.Int {
-	return big.NewInt(0).Set(n.initialGasPrice)
+	return big.NewInt(0).Set(n.fee.InitialGasPrice)
 }
 
 // MinDiscountedGasPrice returns minimum gas price after giving maximum discount
 func (n Network) MinDiscountedGasPrice() *big.Int {
-	return big.NewInt(0).Set(n.minDiscountedGasPrice)
+	return big.NewInt(0).Set(n.fee.MinDiscountedGasPrice)
 }
 
-// TxBankSendGas returns gas required by tx bank send
-func (n Network) TxBankSendGas() uint64 {
-	return n.txBankSendGas
+// DeterministicGas returns deterministic gas amounts required by some message types
+func (n Network) DeterministicGas() DeterministicGasConfig {
+	return n.fee.DeterministicGas
 }
 
 // NetworkByChainID returns config for a predefined config.
--- a/app/network.go
+++ b/app/network.go
@@ -53,14 +53,22 @@ func init() {
 			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
 			AddressPrefix: "core",
 			TokenSymbol:   TokenSymbolMain,
-			MinGasPrice:   big.NewInt(1000),
+			Fees: FeeConfig{
+				InitialGasPrice: big.NewInt(1500),
+				MinGasPrice:     big.NewInt(1000),
+				TxBankSendGas:   120000,
+			},
 		},
 		{
 			ChainID:       Devnet,
 			GenesisTime:   time.Date(2022, 6, 27, 12, 0, 0, 0, time.UTC),
 			AddressPrefix: "devcore",
 			TokenSymbol:   TokenSymbolDev,
-			MinGasPrice:   big.NewInt(1000),
+			Fees: FeeConfig{
+				InitialGasPrice: big.NewInt(1500),
+				MinGasPrice:     big.NewInt(1000),
+				TxBankSendGas:   120000,
+			},
 		},
 	}
 
@@ -77,24 +85,33 @@ func init() {
 
 var networks = map[ChainID]NetworkConfig{}
 
+// FeeConfig is the part of network config defining parameters of our fee model
+type FeeConfig struct {
+	InitialGasPrice *big.Int
+	MinGasPrice     *big.Int
+	TxBankSendGas   uint64
+}
+
 // NetworkConfig helps initialize Network instance
 type NetworkConfig struct {
 	ChainID        ChainID
 	GenesisTime    time.Time
 	AddressPrefix  string
 	TokenSymbol    string
-	MinGasPrice    *big.Int
+	Fees           FeeConfig
 	FundedAccounts []FundedAccount
 	GenTxs         []json.RawMessage
 }
 
 // Network holds all the configuration for different predefined networks
 type Network struct {
-	chainID       ChainID
-	genesisTime   time.Time
-	addressPrefix string
-	tokenSymbol   string
-	minGasPrice   *big.Int
+	chainID         ChainID
+	genesisTime     time.Time
+	addressPrefix   string
+	tokenSymbol     string
+	initialGasPrice *big.Int
+	minGasPrice     *big.Int
+	txBankSendGas   uint64
 
 	mu             *sync.Mutex
 	fundedAccounts []FundedAccount
@@ -104,14 +121,16 @@ type Network struct {
 // NewNetwork returns a new instance of Network
 func NewNetwork(c NetworkConfig) Network {
 	n := Network{
-		genesisTime:    c.GenesisTime,
-		chainID:        c.ChainID,
-		addressPrefix:  c.AddressPrefix,
-		tokenSymbol:    c.TokenSymbol,
-		minGasPrice:    big.NewInt(0).Set(c.MinGasPrice),
-		mu:             &sync.Mutex{},
-		fundedAccounts: append([]FundedAccount{}, c.FundedAccounts...),
-		genTxs:         append([]json.RawMessage{}, c.GenTxs...),
+		genesisTime:     c.GenesisTime,
+		chainID:         c.ChainID,
+		addressPrefix:   c.AddressPrefix,
+		tokenSymbol:     c.TokenSymbol,
+		initialGasPrice: big.NewInt(0).Set(c.Fees.InitialGasPrice),
+		minGasPrice:     big.NewInt(0).Set(c.Fees.MinGasPrice),
+		txBankSendGas:   c.Fees.TxBankSendGas,
+		mu:              &sync.Mutex{},
+		fundedAccounts:  append([]FundedAccount{}, c.FundedAccounts...),
+		genTxs:          append([]json.RawMessage{}, c.GenTxs...),
 	}
 
 	return n
@@ -268,8 +287,19 @@ func (n Network) TokenSymbol() string {
 	return n.tokenSymbol
 }
 
+// InitialGasPrice returns initial gas price used by the first block
+func (n Network) InitialGasPrice() *big.Int {
+	return big.NewInt(0).Set(n.initialGasPrice)
+}
+
+// MinGasPrice returns minimum gas price after giving maximum discount
 func (n Network) MinGasPrice() *big.Int {
 	return big.NewInt(0).Set(n.minGasPrice)
+}
+
+// TxBankSendGas returns gas required by tx bank send
+func (n Network) TxBankSendGas() uint64 {
+	return n.txBankSendGas
 }
 
 // NetworkByChainID returns config for a predefined config.
--- a/app/network.go
+++ b/app/network.go
@@ -5,13 +5,15 @@ import (
 	_ "embed"
 	"encoding/json"
 	"io/ioutil"
+	"math/big"
 	"os"
 	"sync"
 	"text/template"
 	"time"
 
-	"github.com/CoreumFoundation/coreum/pkg/types"
 	"github.com/pkg/errors"
+
+	"github.com/CoreumFoundation/coreum/pkg/types"
 
 	cosmossecp256k1 "github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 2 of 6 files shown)


ysv (changes requested):
I will really appreciate if you could make more extensive description to this PR because for me it is hard to understand how this was achieved

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109#-:-N6rn_vU3Cb02Pzjzp_x:bebwljh)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @miladz68, @oliverasistemas, and @xlab)

oliverasistemas (commented):
CI is failing for this PR

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109#-:-N6sl-rH-1izW6KDj_Pn:b-bwkis0)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @miladz68 and @xlab)

wojtek-coreum (commented):
@oliverasistemas yes, it will stop once this one is merged: https://reviewable.io/reviews/CoreumFoundation/crust/20#- into crust

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109#-:-N6vJ91k1GkVBotIZOxy:b-xe2jct)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @miladz68 and @xlab)

miladz68 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @wojtek-coreum and @xlab)

___
*[`app/network.go` line 58 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FFxDQEG5sVwJ_RQyE:-N7FFxDQEG5sVwJ_RQyF:bgyxp37) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L58)):*
> ```Go
> 			Fees: FeeConfig{
> 				InitialGasPrice: big.NewInt(1500),
> 				MinGasPrice:     big.NewInt(1000),
> ```

I can see `InitialGasPrice` can be a network config which will later be decided by governance. But I don't get the `MinGasPrice` field.
___
*[`x/auth/ante/ante.go` line 1 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FM1s670iOjo2Viuqt:-N7FM1s670iOjo2Viuqu:b-w8cshh) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/ante.go#L1)):*
> ```Go
> package ante
> ```

Do we really need to copy this package in order to achieve our use case ? The way that I understand it we just need to add a ante handler which checks provided fee vs network-calculated fee and either rejects or accepts. it should be achievable without copying this package.

miladz68 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @wojtek-coreum and @xlab)

___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7FPgdf1qI0FjL6Cmya:bv3zbi2) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*
> ```Go
> 
> // AnteHandle handles transaction in ante decorator
> func (mfd MempoolFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
> ```

I can see improvements to adapt this code better to our use cases, but I am not seeing any functionality added (i.e the original code can get the same result). This goes with my argument of why copy this ?

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @miladz68 and @xlab)

___
*[`app/network.go` line 58 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FFxDQEG5sVwJ_RQyE:-N7FSqiGEOdEg89Z6rMu:bl1ra0h) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L58)):*

I can see `InitialGasPrice` can be a network config which will later be decided by governance. But I don't get the `MinGasPrice` field.
</blockquote></details>

Both parameters come from the equation describing our fee model:
- `InitialGasPrice` - gas price if previous blocks were empty 
- `MinGasPrice` - gas price after applying maximum discount
___
*[`x/auth/ante/ante.go` line 1 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FM1s670iOjo2Viuqt:-N7FTKNI78xft6FW8fmE:bnnluh9) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/ante.go#L1)):*
> it should be achievable without copying this package

I agree, it should be. Welcome to Cosmos SDK. Thankfully, I must have copied only the constructor of ante handler which is only one function so not a big deal.
___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7FU1kn0kNgp5rIIqeX:bfm1nyx) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*

I can see improvements to adapt this code better to our use cases, but I am not seeing any functionality added (i.e the original code can get the same result). This goes with my argument of why copy this ?
</blockquote></details>

I don't understand.
The original code took the minimum fee from the `app.toml` so owner of each node could change it to any value at any time causing the problem we discussed earlier - user doesn't know what minimum fee he should offer.

In this PR, minimum fee is set to the same value on entire network and node owner can't override it. In the future it will be stored as part of blockchain state once fee model is fully implemented.

miladz68 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N6rbcrs3VWhdmaMPkbj:-N7FuTNZ-FbXClljPj2_:b-q8ilnv) (no related file):*

PR related to this one exists in `crust` repo: https://reviewable.io/reviews/CoreumFoundation/crust/20#-
</blockquote></details>

gotcha

___
___
*[`app/network.go` line 58 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FFxDQEG5sVwJ_RQyE:-N7FsCVw7oP5fHERN5Ia:b-2j0tkp) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L58)):*

Both parameters come from the equation describing our fee model:
- `InitialGasPrice` - gas price if previous blocks were empty 
- `MinGasPrice` - gas price after applying maximum discount
</blockquote></details>

I think we should find a better name for `MinGasPrice` so it is not confused with a variable with the same name used by cosmos sdk itself. 
maybe `MaxDiscountedGasPrice` or `MaxDiscountPercentage` ?
___
*[`x/auth/ante/ante.go` line 1 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FM1s670iOjo2Viuqt:-N7Fsn_G-RGZpHkSuD9C:bzfqbg2) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/ante.go#L1)):*

> it should be achievable without copying this package

I agree, it should be. Welcome to Cosmos SDK. Thankfully, I must have copied only the constructor of ante handler which is only one function so not a big deal.
</blockquote></details>

I think if it is achievable without copying, then let's not copy. 
@ysv @xlab would you like to comment ?
___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7FtCKNAyruBdfSC08K:b-na5nti) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*

I don't understand.
The original code took the minimum fee from the `app.toml` so owner of each node could change it to any value at any time causing the problem we discussed earlier - user doesn't know what minimum fee he should offer.

In this PR, minimum fee is set to the same value on entire network and node owner can't override it. In the future it will be stored as part of blockchain state once fee model is fully implemented.
</blockquote></details>

I understand now.
I now see what you are trying to achieve here. but I would argue that we should not adapt this code for our gas price functionality. Maybe leave it as is and implement another handler for our gasPrices and not use `MinGasPrice` for that.
@ysv @xlab would you care to comment ?

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`x/auth/ante/ante.go` line 1 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FM1s670iOjo2Viuqt:-N7G5QRN4zP9W2Ge09m0:b-4dpehw) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/ante.go#L1)):*
> I think if it is achievable without copying, then let's not copy.

But is it?
___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7G5qUj9Ke4RXd1cahJ:bnfqdlx) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*

I understand now.
I now see what you are trying to achieve here. but I would argue that we should not adapt this code for our gas price functionality. Maybe leave it as is and implement another handler for our gasPrices and not use `MinGasPrice` for that.
@ysv @xlab would you care to comment ?
</blockquote></details>

But this is how Cosmos SDK is designed. Anything related to enforcing gas & fees policy is done by ante decorator added to ante handler. There is default ante handler with predefined decorators. If you want to modify that logic you need to copy ante handler constructor and add decorators of your selection there. This is how it works. Cosmos SDK implemented "stackoverflow-driven development" (all rights to this term reserved) - whatever you want to modify or add, you copy the code first.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @xlab and @ysv)

___
*[`app/network.go` line 58 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FFxDQEG5sVwJ_RQyE:-N7G70kpBvuv5qpUx2il:b338xjp) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L58)):*

I think we should find a better name for `MinGasPrice` so it is not confused with a variable with the same name used by cosmos sdk itself. 
maybe `MaxDiscountedGasPrice` or `MaxDiscountPercentage` ?
</blockquote></details>

I can rename, but actually the meaning of `MinGasPrice` I defined is exactly the same as defined by cosmos - don't add txes declaring lower gas price to mempool. In our case `MinGasPrice` is just equal to `MaxDiscountedGasPrice` because according to our fee model it's technically impossible that transaction declaring gas price below `MaxDiscountedGasPrice` is ever processed. So both names are correct from technical point of view, the only question is which one is more clear for us.

@ysv Any opinion?

ysv (changes requested):
> standard ante handler contains MempoolFeeDecorator decorator which is responsible for checking if fee provided with transaction is below the level required by the node

"below the level required by the node" - but shouldn’t fee be above required level ?

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109#-:-N7FvpWE1_mWn7TBIHDn:bib1m4c)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @miladz68, @xlab, and @ysv)

___
*[`app/network.go` line 58 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FFxDQEG5sVwJ_RQyE:-N7G8TO68eyxuLpQJuJM:bfnaov4) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L58)):*

I can rename, but actually the meaning of `MinGasPrice` I defined is exactly the same as defined by cosmos - don't add txes declaring lower gas price to mempool. In our case `MinGasPrice` is just equal to `MaxDiscountedGasPrice` because according to our fee model it's technically impossible that transaction declaring gas price below `MaxDiscountedGasPrice` is ever processed. So both names are correct from technical point of view, the only question is which one is more clear for us.

@ysv Any opinion?
</blockquote></details>

both of you points are clear for me

`MinGasPrice` was also confusing for me initially and I wanted to put exactly the same comment as Milad but while reading I got used to it.

However I still prefer to rename it to `DiscountedGasPrice` it plays nicely in pair with `InitialGasPrice` (like you had full price and discounted). While in `ante` I would keep `MinGasPrice` because it seems more natural in that place and `MinGasPrice` might not be always equal to `DiscountedGasPrice` (Lets say we want to always have 10% overhead on top).

So for me it seems that these 2 variables a slightly different by their purpose

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 58 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FFxDQEG5sVwJ_RQyE:-N7GGKO707_mTxH38EKa:b2ulp42) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L58)):*

both of you points are clear for me

`MinGasPrice` was also confusing for me initially and I wanted to put exactly the same comment as Milad but while reading I got used to it.

However I still prefer to rename it to `DiscountedGasPrice` it plays nicely in pair with `InitialGasPrice` (like you had full price and discounted). While in `ante` I would keep `MinGasPrice` because it seems more natural in that place and `MinGasPrice` might not be always equal to `DiscountedGasPrice` (Lets say we want to always have 10% overhead on top).

So for me it seems that these 2 variables a slightly different by their purpose



</blockquote></details>

Renamed to `MinDiscountedGasPrice`.

Why not `DiscountedGasPrice`? Because any price with any discount is `DiscountedGasPrice` but the minimum price (the one after applying max discount) is `MinDiscountedGasPrice`.

Why not `MaxDiscountedGasPrice`? Because price after applying max discount is the min price, so having "min" next to "gas price" seems more accurate to me than having "max" there.

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @xlab)

___
*[`app/network.go` line 58 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FFxDQEG5sVwJ_RQyE:-N7GPtyE2cA3z1647SLX:b50r1qx) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L58)):*

Renamed to `MinDiscountedGasPrice`.

Why not `DiscountedGasPrice`? Because any price with any discount is `DiscountedGasPrice` but the minimum price (the one after applying max discount) is `MinDiscountedGasPrice`.

Why not `MaxDiscountedGasPrice`? Because price after applying max discount is the min price, so having "min" next to "gas price" seems more accurate to me than having "max" there.
</blockquote></details>

that is why I decided to not put neither min nor max. But you comment also makes sense
___
*[`app/network.go` line 59 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GBvrz8isU9OvJILaR:-N7GBvrz8isU9OvJILaS:b-udtkew) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L59)):*
> ```Go
> 				InitialGasPrice: big.NewInt(1500),
> 				MinGasPrice:     big.NewInt(1000),
> 				TxBankSendGas:   120000,
> ```

So here we have static fee defined for Bank Send Tx type do I understand it correctly ? But it is never used for now and added just as an exampl ?
___
*[`app/network.go` line 101 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GGJwZ9gdKgInPCPZF:-N7GGJwZ9gdKgInPCPZG:b39uvr6) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L101)):*
> ```Go
> 	AddressPrefix  string
> 	TokenSymbol    string
> 	Fees           FeeConfig
> ```

Fees -> Fee WDYT ?
___
*[`app/network.go` line 112 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GG85E5eob_M3Wz9_V:-N7GG85E5eob_M3Wz9_W:b8tij3e) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L112)):*
> ```Go
> 	addressPrefix   string
> 	tokenSymbol     string
> 	initialGasPrice *big.Int
> ```

shouln't we just use FeeConfig struct instead of 3 separate fields here ?
___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7GHEIn-JFPMM3yIakR:bbf4br) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*
> ```Go
> func (n Network) MinGasPrice() *big.Int {
> 	return big.NewInt(0).Set(n.minGasPrice)
> }
> ```

as I see in cosmos they use sdk.Coin type for gasPrice (it is actually `type DecCoins []DecCoin`)
I prefer we also return sdk.Coin here because of multiple benefits





_Code quote:_
```Go
func (n Network) InitialGasPrice() *big.Int {
	return big.NewInt(0).Set(n.initialGasPrice)
}

// MinGasPrice returns minimum gas price after giving maximum discount
func (n Network) MinGasPrice() *big.Int {
	return big.NewInt(0).Set(n.minGasPrice)
}
```
___
*[`app/network_test.go` line 276 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GIibJ2vM_eROg9_Rr:-N7GIibJ2vM_eROg9_Rs:b2q9935) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network_test.go#L276)):*
> ```Go
> 	n1.MinGasPrice().Set(big.NewInt(100))
> 
> 	assertT.Equal("2", n1.InitialGasPrice().String())
> ```

minor: I would still compare big Int here not strings



_Code quote:_
```Go
	assertT.Equal("2", n1.InitialGasPrice().String())
```
___
*[`x/auth/ante/ante.go` line 1 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FM1s670iOjo2Viuqt:-N7GIr1OCw7xvH-fxEOE:b-5dvtmz) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/ante.go#L1)):*

> I think if it is achievable without copying, then let's not copy.

But is it?
</blockquote></details>

initially I was also purely against copying it but then I went through cosmos and now I'm not sure.


The problems are:
1. decorator list is hardcoded there and you can you decorator only on top or bottom of all decorators described in standard cosmos ante
2. you can't remove/replace one of the default decorators


Since we copied very limited amount of code I'm more or less ok with it only because there is almost no other way at least I don't see it but if you have an idea you are welcome. 

@wojtek-coreum my concern is that we should not just copy but put explicit comments from where it was copied and that it is not original code. The same comment needs to be added in fee.go because it is also very similar to ante/fee.go
___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7GOcuy4CR2m_s5lFMh:b-8wpbwk) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*
> I now see what you are trying to achieve here. but I would argue that we should not adapt this code for our gas price functionality. Maybe leave it as is and implement another handler for our gasPrices and not use MinGasPrice for that.
I now see what you are trying to achieve here. but I would argue that we should not adapt this code for our gas price functionality. Maybe leave it as is and implement another handler for our gasPrices and not use MinGasPrice for that.

@miladz68  I'm not sure I understand you 100%. 

because another option I see here is to use default `MempoolFeeDecorator` and just set `MinGasPrices` in ctx

Lets discuss on the call

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @miladz68, @wojtek-coreum, @xlab, and @ysv)

___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7GZQyI8tTmNSsWwQk2:b-fbzkdu) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*

> because another option I see here is to use default MempoolFeeDecorator and just set MinGasPrices in ctx

1. Which means you would have to copy ante handler and override `ContextDecorator` or write another decorator which sets it.
2. Or you would have to copy half of the cosmos sdk to modify the place where it sets this based on `app.toml`

(2) would be a temporary solution anyway because in the future parameters of the fee model will be managed by governance module (it means they must be read from KVStore)
</blockquote></details>

Now I see that we could set it inside `func (app *App) BeginBlocker(...)` but it's still not a good solution, because then user would be able to set fees like `Xcore,YsomeToken`.
Which could be acceptable but I prefer to accept only cores for fees and reject transactions offering anything else than cores.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on @miladz68, @xlab, and @ysv)

___
*[`app/network.go` line 101 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GGJwZ9gdKgInPCPZF:-N7H03rwAsSs7icYhGdi:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L101)):*

Fees -> Fee WDYT ?
</blockquote></details>

Done.
___
*[`app/network.go` line 112 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GG85E5eob_M3Wz9_V:-N7H1vPu-9MUvzPBVu1G:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L112)):*

Yes, in next PR I'm working on I already did that. I even proposed before that entire `NetworkConfig` should be referenced directly as a field of `Network` but @miladz68 was agains doing this.

I will add `FeeConfig` here
</blockquote></details>

Done.
___
*[`x/auth/ante/ante.go` line 1 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FM1s670iOjo2Viuqt:-N7H5HhnC0nkW0TVlvnj:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/ante.go#L1)):*

initially I was also purely against copying it but then I went through cosmos and now I'm not sure.


The problems are:
1. decorator list is hardcoded there and you can you decorator only on top or bottom of all decorators described in standard cosmos ante
2. you can't remove/replace one of the default decorators


Since we copied very limited amount of code I'm more or less ok with it only because there is almost no other way at least I don't see it but if you have an idea you are welcome. 

@wojtek-coreum my concern is that we should not just copy but put explicit comments from where it was copied and that it is not original code. The same comment needs to be added in fee.go because it is also very similar to ante/fee.go


</blockquote></details>

Done.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 101 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GGJwZ9gdKgInPCPZF:-N7KMHge0RZ92ghDxsR6:b-jozd95) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L101)):*

Done.
</blockquote></details>

Both Fees and Fee are good with me, but why not FeeConfig ? isn't it what it really is ?
___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7KLSk4DIjiCCqkTMO2:bni5ndw) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*

This is the question...

My initial idea was to use non-cosmos types to define the network. Otherwise entire cosmos-sdk legacy must be imported in every module which wants to use `Network`. That's why I use `big.Int` here and then convert it to `sdk.Coin` only i n the place where I pass it to cosmos-related parts.
</blockquote></details>

if we want to abstract away cosmos sdk code then we should do it completely and properly, doing it randomly here and there is worse than not doing it. so it means complete commitment to it. 
Since I do not know the ins and outs of cosmos sdk I cannot say how feasible this is, just putting it out there. My initial judgment is that it will be too much to wrap the entirety of the cosmos API.
___
*[`app/network.go` line 92 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7KMtqg0SkHvixjcu3V:-N7KMtqg0SkHvixjcu3W:b-t14pz) ([raw file](https://github.com/coreumfoundation/coreum/blob/954ed2c6762695699cbfedf8cece6fed1ea25ccc/app/network.go#L92)):*
> ```Go
> 	InitialGasPrice       *big.Int
> 	MinDiscountedGasPrice *big.Int
> 	TxBankSendGas         uint64
> ```

Just a point to consider, should the TxBankSendGas be a constant ? I mean I see no reason to have it differently for each network.
___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7KL3D17SWrB9pj14eb:b-7sxuv9) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*

And what's most important: I will modify ante handler anyway to support deterministic gas requirements and fee model so ante will be copied anyway.
</blockquote></details>

This is now resolved based on the live discussion we had

miladz68 (commented):
> Reviewed 1 of 6 files at r1, 1 of 3 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7KQIgf3Hrm9lM-h2Ee:bfkc4r7) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*

if we want to abstract away cosmos sdk code then we should do it completely and properly, doing it randomly here and there is worse than not doing it. so it means complete commitment to it. 
Since I do not know the ins and outs of cosmos sdk I cannot say how feasible this is, just putting it out there. My initial judgment is that it will be too much to wrap the entirety of the cosmos API.
</blockquote></details>

It's not done randomly. There is no single place inside this config using cosmos type. So it is done consistently.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7KWKvzER3hdN0Qc0BK:b-tb3jhg) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*

It's not done randomly. There is no single place inside this config using cosmos type. So it is done consistently.
</blockquote></details>

I meant all over the code base, not in your PR.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7KWTWY6tBLOQ6MUAW2:b-uobjfd) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*

I meant all over the code base, not in your PR. 
</blockquote></details>

the question is where do you draw the line on where you want to have you own abstractions and where you cross over to cosmos world. in a traditional web service it is simple, you put all db types inside the db layer and so on. but here, I don't how we would draw the lines between abstractions here.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7KaREY5rQV5y5KWlqV:b-sf9pto) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*

the question is where do you draw the line on where you want to have you own abstractions and where you cross over to cosmos world. in a traditional web service it is simple, you put all db types inside the db layer and so on. but here, I don't how we would draw the lines between abstractions here.
</blockquote></details>

We may discuss it after daily. To keep the discussion specific to this PR, preparing genesis should be the only place where cosmos types are used. But now I see that because of that logic, cosmos legacy must be imported by any module using `Network` anyway, even if that module doesn't need to use that logic. I'm not happy about that. So maybe I shouldn't try to save the world. Let's talk about it during the meeting.

ysv (changes requested):
> Reviewed 1 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 3 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @xlab)

___
*[`app/network.go` line 59 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GBvrz8isU9OvJILaR:-N7Ga_T192I4RK65zRWq:b-6cwo49) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L59)):*

Static gas amount - not fee, fee = gas price * gas amount.

It's not used physically but is used logically, wherever I left fixme in the other PR like this:
```
// FIXME (wojtek): Take this value from Network.TxBankSendGas() once Milad integrates it into crust
```
</blockquote></details>

yes gas amount that is what I meant

I mean that it is not used in cored for now anywhere (you don't calculate fee based on this value)

And in the future PRs we will start determining gas based base on this value & tx type
___
*[`app/network.go` line 101 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GGJwZ9gdKgInPCPZF:-N7KbOFX-Jn6YeZjtdCZ:bc4xj1p) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L101)):*

Both Fees and Fee are good with me, but why not FeeConfig ? isn't it what it really is ?
</blockquote></details>

not FeeConfig probably because it is already inside NetworkConfig
But I still don't mind Fee or FeeConfig
___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7Kcmp0AkBjQ5drxuHT:b-qh4ygd) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*

We may discuss it after daily. To keep the discussion specific to this PR, preparing genesis should be the only place where cosmos types are used. But now I see that because of that logic, cosmos legacy must be imported by any module using `Network` anyway, even if that module doesn't need to use that logic. I'm not happy about that. So maybe I shouldn't try to save the world. Let's talk about it during the meeting.
</blockquote></details>

makes sense for me to discuss this because it is important concept
___
*[`app/network.go` line 92 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7KMtqg0SkHvixjcu3V:-N7Kb7YG3Ad6Ij4msu72:b-bb377g) ([raw file](https://github.com/coreumfoundation/coreum/blob/954ed2c6762695699cbfedf8cece6fed1ea25ccc/app/network.go#L92)):*

Just a point to consider, should the TxBankSendGas be a constant ? I mean I see no reason to have it differently for each network.
</blockquote></details>

maybe
I personally don't mind. I see this definition as preliminary for now and I think we are likely to restructure this fee definition in the future
___
*[`x/auth/ante/fee.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7FPgdf1qI0FjL6Cmy_:-N7KcZng5ZGOkOrM0AF0:bw350ni) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/x/auth/ante/fee.go#L26)):*

This is now resolved based on the live discussion we had
</blockquote></details>

yes ok for me also

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 3 unresolved discussions (waiting on @miladz68, @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 59 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GBvrz8isU9OvJILaR:-N7KiIxl57XaHGLhY3XL:b82di46) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L59)):*

yes gas amount that is what I meant

I mean that it is not used in cored for now anywhere (you don't calculate fee based on this value)

And in the future PRs we will start determining gas based base on this value & tx type
</blockquote></details>

No, in `cored` it is not used yet - will be in the next PR. Here I just placed it to be referenced in `FIXME` comments specified in the other PR.
___
*[`app/network.go` line 92 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7KMtqg0SkHvixjcu3V:-N7KiZOL0tHDJWnPREXk:bs8uhx3) ([raw file](https://github.com/coreumfoundation/coreum/blob/954ed2c6762695699cbfedf8cece6fed1ea25ccc/app/network.go#L92)):*

maybe
I personally don't mind. I see this definition as preliminary for now and I think we are likely to restructure this fee definition in the future
</blockquote></details>

Probably `DeterministicGas` could be set once and then referenced in all the networks. It could be even hardcoded in some places without being included in `NetworkConfig` but personally I feel it's cleaner if it's specified here.

miladz68 (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 2 unresolved discussions (waiting on @miladz68, @wojtek-coreum, @xlab, and @ysv)

___
*[`app/network.go` line 92 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7KMtqg0SkHvixjcu3V:-N7KmEGTC_w-dVMjs6q4:b-xr0fd7) ([raw file](https://github.com/coreumfoundation/coreum/blob/954ed2c6762695699cbfedf8cece6fed1ea25ccc/app/network.go#L92)):*

Probably `DeterministicGas` could be set once and then referenced in all the networks. It could be even hardcoded in some places without being included in `NetworkConfig` but personally I feel it's cleaner if it's specified here.
</blockquote></details>

I will resolve this then.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109)* status: all files reviewed, 3 unresolved discussions (waiting on @xlab and @ysv)

___
*[`app/network.go` line 298 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7GHEIn-JFPMM3yIakQ:-N7M9f281GMBpqmgTxwT:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/e0e16e63706706b149d2fe52ddce8003aa2bd332/app/network.go#L298)):*

makes sense for me to discuss this because it is important concept
</blockquote></details>

Done.
___
*[`x/auth/ante/ante.go` line 21 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/109#-N7Ls0Ba92w527yCN97D:-N7M9gEvBPysYIJSjDyr:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/edb885ca63f4447b0dccf36fbae518393e65c7cc/x/auth/ante/ante.go#L21)):*

also wanted to discuss this 
</blockquote></details>

Done.

miladz68 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/109#-:-N7PJr-OAbst5BIuLxl-:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @xlab and @ysv)

PR #128, opened 2022-08-08
Introduce `Int` wrapper around `*big.Int` for easier math operations

wojtek-coreum:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/128)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -11,7 +11,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/client/grpc/tmservice"
 	"github.com/cosmos/cosmos-sdk/client/rpc"
 	"github.com/cosmos/cosmos-sdk/codec"
-	"github.com/cosmos/cosmos-sdk/codec/types"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	"github.com/cosmos/cosmos-sdk/server/api"
 	"github.com/cosmos/cosmos-sdk/server/config"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
@@ -96,6 +96,7 @@ import (
 
 	"github.com/CoreumFoundation/coreum/cmd/cored/cosmoscmd"
 	"github.com/CoreumFoundation/coreum/docs"
+	"github.com/CoreumFoundation/coreum/pkg/types"
 	"github.com/CoreumFoundation/coreum/x/auth/ante"
 	// this line is used by starport scaffolding # stargate/app/moduleImport
 )
@@ -199,7 +200,7 @@ type App struct {
 
 	cdc               *codec.LegacyAmino
 	appCodec          codec.Codec
-	interfaceRegistry types.InterfaceRegistry
+	interfaceRegistry codectypes.InterfaceRegistry
 
 	invCheckPeriod uint
 
@@ -552,7 +553,7 @@ func New(
 			BankKeeper:      app.BankKeeper,
 			SignModeHandler: encodingConfig.TxConfig.SignModeHandler(),
 			FeegrantKeeper:  app.FeeGrantKeeper,
-			MinGasPrice:     sdk.NewCoin(ChosenNetwork.TokenSymbol(), sdk.NewIntFromBigInt(ChosenNetwork.MinDiscountedGasPrice().BigInt())),
+			MinGasPrice:     sdk.NewCoin(ChosenNetwork.TokenSymbol(), types.IntToSDK(ChosenNetwork.MinDiscountedGasPrice())),
 			GasRequirements: ante.DeterministicGasRequirements{
 				BankSend: ChosenNetwork.DeterministicGas().BankSend,
 			},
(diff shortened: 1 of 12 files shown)


wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/128)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @begmaroman, @miladz68, @oliverasistemas, @xlab, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/128#-N8xBtKn4xqHTtAO83pO:-N8xBtKo6qauqC1dk8Vk:b-m8s8fl) (no related file):*
It's better to review this PR commit by commit. Start by checking `pkg/types/int.go`

___

xlab:
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/128#-N8xKtnc7mnRZte4cGzQ:-N8xKtnc7mnRZte4cGzR:b-svuzc1) (no related file):*
Cosmos SDK already provides many wrappers to work with Ints for balances, e.g. `sdk.Coin` backed by `sdk.Int`, `sdk.DecCoin` backed by `sdk.Dec` etc. Why introduce an incompatible wrapper?

The sdk.Int and other types also provide necessary rules for protobuf encoding, etc. Why reimplementing Cosmos SDK there again?

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/128)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @xlab, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/128#-N8xKtnc7mnRZte4cGzQ:-N8xP7Aa6D4rsXCpPoRG:b-kasrk9) (no related file):*
<details><summary><i>Previously, xlab (Max Kupriianov) wrote…</i></summary><blockquote>

Cosmos SDK already provides many wrappers to work with Ints for balances, e.g. `sdk.Coin` backed by `sdk.Int`, `sdk.DecCoin` backed by `sdk.Dec` etc. Why introduce an incompatible wrapper?

The sdk.Int and other types also provide necessary rules for protobuf encoding, etc. Why reimplementing Cosmos SDK there again?



</blockquote></details>

To make math around `big.Int` easier without introducing dependency on sdk. This wrapper is so simple that it can be easily taken out of sdk and used as an independent library. We don't need sdk to make big.Int to be easier to use.

___

xlab:
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/128#-N8xKtnc7mnRZte4cGzQ:-N8xRp5v8fW4pYt1tvld:b9vhq3q) (no related file):*

To make math around `big.Int` easier without introducing dependency on sdk. This wrapper is so simple that it can be easily taken out of sdk and used as an independent library. We don't need sdk to make big.Int to be easier to use.
</blockquote></details>

And what is the problem of having a dependency on SDK? Right now I am seeing that we just reimplementing everything SDK provides.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/128)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @xlab, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/128#-N8xKtnc7mnRZte4cGzQ:-N8xSZhQDl60gsYH6D6b:bj3bwki) (no related file):*
We decided to use cosmos SDK as a bottom layer for communication inside our client.
It means this bottom layer is used only internally. That's why `Client`, `NetworkConfig` and `Network` expose their own types to the user.

> we just reimplementing everything SDK provides

You are exaggerating. Only two types have been copied so far: `Coin` and `Int`. I believe that more or less this is the end of copying code from there.

`Int` shouldn't even be a part of Cosmos SDK  because this is how `big.Int` should be implemented by default. This is very generic, project-independent wrapper which removes hassle from math operations on `big.Int`s without adding any business logic there. So it is very natural that we don't expose entire SDK just to use that wrapper.

___

miladz68 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/128)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @oliverasistemas, @wojtek-coreum, @xlab, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/128#-N90NLPRCnAv1vC-Hvkd:-N90NLPRCnAv1vC-Hvke:bfneak3) (no related file):*
I don't think wrapping `big.Int` has much utility for us. there are some convenience functions introduced, but I think the cost outweighs the benefits (i.e adding a new abstraction just to get a few convenience functions is not justified). 

I think those convenience functions should be moved on the Coin type. IMO people should try to use coins not numbers, and we should encourage that, as it simplifies our abstraction. 

Moreover I think `InitialGasPrice` and other price configs should return coins and not `big.Int` or `types.Int`.  

As for the convenience functions, we just need to add `AddCoin/SubCoin` and `AddAmount/SubAmount` methods. We can have the first couple validate whether Token Symbols match and the second may just add the amount. 

___

PR #142, opened 2022-08-18
Playground: Implement coin freezing mechanism

begmaroman:
- `freeze` module - can freeze and unfreeze coins by any account. Haven't implemented token ownership within this PR. 
- `bank` - forked module from the original repo with an additional check for coin freezeness in send operations by importing the keeper from `freeze` module.

Feel free to suggest more logic to add so I could play around it and get more understanding of how things work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)


dzmitryhil (changes requested):
I know that there are a lot of code generated by default, but It would be good to review only task specific code, so all redundant/templated code could be removed next time :-)

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142#-:-N9q37Y9-ZlQu5QyQBLc:b-70w06z)* status: 0 of 79 files reviewed, 5 unresolved discussions (waiting on @begmaroman)

___
*[`proto/freeze/query.proto` line 13 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q28em8WdeWHGiLKNp:-N9q28em8WdeWHGiLKNq:b-7kkune) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/proto/freeze/query.proto#L13)):*
> ```protobuf
> 
> // Query defines the gRPC querier service.
> service Query {
> ```

Haven't you implementer the Query for the Frozen coins? Those are generic for params.
___
*[`x/bank/keeper/msg_server.go` line 48 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q5TBo4nBXh3xzS8n5:-N9q5TBo4nBXh3xzS8n6:b-l0tdli) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/bank/keeper/msg_server.go#L48)):*
> ```Go
> 	}
> 
> 	if err = k.isFrozenCoins(ctx, from, msg.Amount...); err != nil {
> ```

This is the one point of the integration. If any of the modules use the BankKeeper as a dependency they will ignore that frozen coins, which might lead to unpredictable state, you should use the Bank keeper as that point
___
*[`x/freeze/genesis.go` line 11 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q2aYME-1bPN7J6UCG:-N9q2aYME-1bPN7J6UCH:bv9pwrj) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/freeze/genesis.go#L11)):*
> ```Go
> // InitGenesis initializes the capability module's state from a provided genesis
> // state.
> func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
> ```

If your module is able to keep a state it should be possible to import that state and export, here you do it for the params only.
___
*[`x/freeze/keeper/keeper.go` line 16 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q0d2nEbzDMEzQzmm8:-N9q0d2nEbzDMEzQzmm9:bos9bx3) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/freeze/keeper/keeper.go#L16)):*
> ```Go
> )
> 
> type Keeper interface {
> ```

What was the reason to copy the enite module instead of overriding it?
___
*[`x/freeze/keeper/keeper.go` line 65 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q1-AQCSdRGpHWeEtV:-N9q1-AQCSdRGpHWeEtW:b-2oshqp) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/freeze/keeper/keeper.go#L65)):*
> ```Go
> 
> func (k *BaseKeeper) FreezeCoin(ctx sdk.Context, holder sdk.AccAddress, denom string) error {
> 	store := prefix.NewStore(ctx.KVStore(k.storeKey), holder.Bytes())
> ```

1) From what I see here, you freeze the entire denom, not Coins.
The Coins are: []Coin where each Coin{denom + amount}, so the logic is wrong.
Even if it would be right - here you freeze the demon, but don't validate whether it exists or not, what if the denom doesn't exist at all? 
Just for future - there is a demon metadata store (in the bank module) so you should register new or use the existing.

2) It's bad practice to save the key that way, you save it as root key, just imagine that the store is a json and each root item is an logical entity, so in you case it would be better to add a root key and then add child entities to it.

miladz68 (changes requested):
> Reviewed 27 of 79 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 27 of 79 files reviewed, 6 unresolved discussions (waiting on @begmaroman)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA3ItiADge1_yBH6_f1:-NA3ItiB5bxqGkqROMfj:bv6camm) (no related file):*
I tried to review it, but it is almost impossible to review in a timely manner, maybe it makes sense to have a branch with generated and copied code, and make a PR against that branch with changes, so we would know which are the changes. 

___

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 19 of 82 files reviewed, 6 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @miladz68)

___
*[`proto/freeze/query.proto` line 13 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q28em8WdeWHGiLKNp:-NA4FjgC4bV_k6GQuRF3:bpn17gu) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/proto/freeze/query.proto#L13)):*

Haven't you implementer the Query for the Frozen coins? Those are generic for params.
</blockquote></details>

No queries were implemented since we don't need them. Should this file be removed? 
___
*[`x/bank/keeper/msg_server.go` line 48 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q5TBo4nBXh3xzS8n5:-NA4ougd4GpSEtBbn6va:b-bprc3) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/bank/keeper/msg_server.go#L48)):*

This is the one point of the integration. If any of the modules use the BankKeeper as a dependency they will ignore that frozen coins, which might lead to unpredictable state, you should use the Bank keeper as that point
</blockquote></details>

Agree, done
___
*[`x/freeze/genesis.go` line 11 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q2aYME-1bPN7J6UCG:-NA4oo_l06FfQ8Xp7arP:bb6un1j) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/freeze/genesis.go#L11)):*

If your module is able to keep a state it should be possible to import that state and export, here you do it for the params only.
</blockquote></details>

Done
___
*[`x/freeze/keeper/keeper.go` line 16 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q0d2nEbzDMEzQzmm8:-NA4GGI75X3I4-Sojc4V:bjb7qr7) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/freeze/keeper/keeper.go#L16)):*

What was the reason to copy the enite module instead of overriding it?
</blockquote></details>

Not really the entire module. A lot of parts from the native bank module are re-used here. The keeper and module are forked just to make it easier to work with. 
___
*[`x/freeze/keeper/keeper.go` line 65 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q1-AQCSdRGpHWeEtV:-NA4ySgO8C97VcL99yxY:btp43em) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/freeze/keeper/keeper.go#L65)):*

1) From what I see here, you freeze the entire denom, not Coins.
The Coins are: []Coin where each Coin{denom + amount}, so the logic is wrong.
Even if it would be right - here you freeze the demon, but don't validate whether it exists or not, what if the denom doesn't exist at all? 
Just for future - there is a demon metadata store (in the bank module) so you should register new or use the existing.

2) It's bad practice to save the key that way, you save it as root key, just imagine that the store is a json and each root item is an logical entity, so in you case it would be better to add a root key and then add child entities to it. 

</blockquote></details>

1. Done
2. Will update soon

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 19 of 82 files reviewed, 6 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @miladz68)

___
*[`x/freeze/keeper/keeper.go` line 65 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q1-AQCSdRGpHWeEtV:-NA4zXN_3H4Y12PtD8zz:badvrkx) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/x/freeze/keeper/keeper.go#L65)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

1. Done
2. Will update soon
</blockquote></details>

2. Done

dzmitryhil (changes requested):
> Reviewed 79 of 79 files at r1, 2 of 42 files at r2.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 41 of 82 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @miladz68)

___
*[`proto/freeze/query.proto` line 13 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q28em8WdeWHGiLKNp:-NA9QmUsCAKY0rfPwMHA:b-4bzadr) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/proto/freeze/query.proto#L13)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

No queries were implemented since we don't need them. Should this file be removed? 
</blockquote></details>

Why you don't need them? It's obvious that users might need to get the locked coins

dzmitryhil (commented):
> Reviewed 39 of 42 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman)

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: all files reviewed, 7 unresolved discussions (waiting on @begmaroman)

___
*[`x/freeze/keeper/keeper.go` line 63 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9RPn22yo94gFoYGRa:-NA9RPn22yo94gFoYGRb:bvo90ta) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L63)):*
> ```Go
> 
> func (k *BaseKeeper) FreezeCoin(ctx sdk.Context, holder sdk.AccAddress, coin sdk.Coin) {
> 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.FrozenCoinKey))
> ```

You don't need go get the store and then store from store, you can just build the key as FrozenCoinKey + holder.
For the next method as well.
___
*[`x/freeze/keeper/keeper.go` line 66 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9RrpR7dwRNxGQOPs2:-NA9RrpR7dwRNxGQOPs3:b-2st7bz) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L66)):*
> ```Go
> 	store = prefix.NewStore(store, holder.Bytes())
> 
> 	key := []byte(coin.Denom)
> ```

It's a bad practive to generate the kes in the code directly better to to have the specific function for it (check SDK modules as an example)



_Suggestion:_
```Go
amount := big.NewInt(0).SetBytes(store.Get(key))
Please check how the decoder "k.cdc.MustUnmarshal" marshals the data and use it.
```
___
*[`x/freeze/keeper/keeper.go` line 68 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9TQa208Q4UJy2mO_9:-NA9TQa208Q4UJy2mO_A:b-pswedf) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L68)):*
> ```Go
> 	key := []byte(coin.Denom)
> 
> 	if store.Has(key) {
> ```

I might be wrong about the business logic but I would assume to increase it rather then set.
___
*[`x/freeze/keeper/keeper.go` line 73 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9TBxzEmbISx4IuQPH:-NA9TBxzEmbISx4IuQPI:bdng4ny) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L73)):*
> ```Go
> 	}
> 
> 	store.Set(key, coin.Amount.BigInt().Bytes())
> ```

Is it possible to freeze the coins which I don't have? If yes - why :-)
___
*[`x/freeze/keeper/keeper.go` line 98 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9YgHW-0aLg3khNH0K:-NA9YgHW-0aLg3khNH0L:bj2juu6) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L98)):*
> ```Go
> }
> 
> func (k *BaseKeeper) GetFrozenCoin(ctx sdk.Context, holder sdk.AccAddress, denom string) sdk.Coin {
> ```

Is it used?

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 68 of 85 files reviewed, 7 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @miladz68)

___
*[`proto/freeze/query.proto` line 13 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/142#-N9q28em8WdeWHGiLKNp:-NA9vuBC0uvMyloiJ54S:b-2q2ied) ([raw file](https://github.com/coreumfoundation/coreum/blob/5c923ad54cbfedd9ab1bf0549323ad86e9f62b3d/proto/freeze/query.proto#L13)):*

Why you don't need them? It's obvious that users might need to get the locked coins
</blockquote></details>

Added a query to list frozen coins per account

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 68 of 85 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @miladz68)

___
*[`x/freeze/keeper/keeper.go` line 68 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9TQa208Q4UJy2mO_9:-NAA8zXm8roOKSYGZrMm:bfsetl6) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L68)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

That is correct. Should I just set the given amount? 
</blockquote></details>

now it's ok 
___
*[`x/freeze/keeper/keeper.go` line 73 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9TBxzEmbISx4IuQPH:-NAA93WV3Q25D7T-Utop:b-oc4n6h) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L73)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

Added an additional check for this 
</blockquote></details>

Now you check whether the account has the demon at all. But should check where it has enough amount.

dzmitryhil (commented):
> Reviewed 6 of 9 files at r4, 1 of 1 files at r5, 7 of 10 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 82 of 85 files reviewed, 3 unresolved discussions (waiting on @begmaroman)

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: 81 of 85 files reviewed, 2 unresolved discussions (waiting on @begmaroman and @dzmitryhil)

___
*[`x/freeze/keeper/keeper.go` line 73 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/142#-NA9TBxzEmbISx4IuQPH:-NAAdlcx4Lvdh3lG_iJd:b-886gsx) ([raw file](https://github.com/coreumfoundation/coreum/blob/537688785ab1f37a7cf1c7c97633494e4e7f08e8/x/freeze/keeper/keeper.go#L73)):*

Now you check whether the account has the demon at all. But should check where it has enough amount.
</blockquote></details>

Fixed

dzmitryhil (approved):
> Reviewed 3 of 10 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/142)* status: all files reviewed, 1 unresolved discussion (waiting on @begmaroman)

PR #149, opened 2022-08-30
Store fee model params on chain

wojtek-coreum:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)


dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 0 of 19 files reviewed, 5 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/feemodel/v1/genesis.proto` line 12 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmwiMV8WL9gMREgoNn:-NAmwiMV8WL9gMREgoNo:b-oval9v) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/genesis.proto#L12)):*
> ```protobuf
> message GenesisState {
>   // params defines all the paramaters of the module.
>   Model params = 1 [(gogoproto.nullable) = false];
> ```

I don't like the idea to use the FeeModel as params here. The params might include the FeeModel related params and other params as well, so you can add the fee_model param with the Model type inside the params {  fee_model: Model}. 
___
*[`proto/coreum/feemodel/v1/model.proto` line 18 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmx2WdAJwHqAsf7y-y:-NAmx2WdAJwHqAsf7y-z:b-cup5vf) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/model.proto#L18)):*
> ```protobuf
> // The input (x value) for that function is calculated by taking short block gas average.
> // Price (y value) being an output of the fee model is used as the minimum gas price for next block.
> message Model {
> ```

Let's add the prefix to it, to generic to understand the it's a FeeModel. 
___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NAmvfMz5DCq4KKJxI-D:b-rn726c) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*
> ```Go
> // module.
> func (amb AppModuleBasic) DefaultGenesis(cdc codec.JSONCodec) json.RawMessage {
> 	return cdc.MustMarshalJSON(&types.GenesisState{Params: amb.feeModel})
> ```

You can set here any params you want since we repance them in the genesis template, so no need to overcomplicate that part. For example the values for the mainnet.
BTW you already have the method for it: 'DefaultModel()' ?
___
*[`x/feemodel/module.go` line 137 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmzjE76Q9rn5_mzTjl:-NAmzjE76Q9rn5_mzTjm:b-5ebue2) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L137)):*
> ```Go
> // InitGenesis performs genesis initialization for the fee module. It returns
> // no validator updates.
> func (am AppModule) InitGenesis(ctx sdk.Context, cdc codec.JSONCodec, data json.RawMessage) []abci.ValidatorUpdate {
> ```

I don't see the Import/Export unti tests?
___
*[`x/feemodel/module.go` line 148 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmw6xeC-ViD-aHKM4k:-NAmw6xf6RIf-NagCt7Y:bail4ln) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L148)):*
> ```Go
> 	}
> 
> 	am.keeper.SetModel(ctx, genesis.Params)
> ```

It seems wierd that SetModel but consume all params. Better to use the SetParams to save the params. And the model is just a part of the params.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 0 of 19 files reviewed, 5 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/feemodel/v1/model.proto` line 18 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmx2WdAJwHqAsf7y-y:-NAn8DQI6PumO1TzSKsH:bna1gi8) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/model.proto#L18)):*

Let's add the prefix to it, to generic to understand the it's a FeeModel. 
</blockquote></details>

it's already inside `feemodel` namespace
___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NAn8MnB97QmKuKgMsri:b-ynr2d8) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*

You can set here any params you want since we repance them in the genesis template, so no need to overcomplicate that part. For example the values for the mainnet.
BTW you already have the method for it: 'DefaultModel()' ?
</blockquote></details>

That's why I was againt introducing `DefaultModel` when @ysv proposed it.

Our config for all the networks go through `NetworkConfig` type. This is the ground truth for the network settings. The fact that `FeeModel` config inside `NetworkConfig` is set by using `feemodel.DefaultModel` is another step of getting values common to all the networks.

I was very sure that adding `feemodel.DefaultModel()` will create a lot of confusion. The facts are like this:
- config of the network used to configure genesis comes from `NetworkConfig`
- `feemodel.DefaultModel` is a weird place to store values common to all our network configs. It is weird place because module package (which is a generic code to be reused by any blockchain) should not contain network specific settings.
___
*[`x/feemodel/module.go` line 137 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmzjE76Q9rn5_mzTjl:-NAn9jPn6vpLwfG8DHQB:bwubhhm) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L137)):*

I don't see the Import/Export unti tests?
</blockquote></details>

import/export is not implemented in this PR - it's a separate task

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 0 of 19 files reviewed, 5 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/feemodel/v1/genesis.proto` line 12 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmwiMV8WL9gMREgoNn:-NAnI5BL2iEyrVcOqMOL:b-t5ogbj) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/genesis.proto#L12)):*

I don't like the idea to use the FeeModel as params here. The params might include the FeeModel related params and other params as well, so you can add the fee_model param with the Model type inside the params {  fee_model: Model}. 
</blockquote></details>

When I saw this comment I thought it might be a good idea but after thinking about it more I disagree. `feemodel` module is responsible exclusively for managing fee models - nothing more. It means that the only params managed by this module are those related to fee model, and all of them are stored inside `Model` structure. There won't be any no-fee params delivered by this module and managed by governance. Saying that, I don't see any reason to introduce additional level in this structure.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 0 of 19 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/feemodel/v1/genesis.proto` line 12 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmwiMV8WL9gMREgoNn:-NAo8m957Dvsqb11wtIh:bvschla) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/genesis.proto#L12)):*

When I saw this comment I thought it might be a good idea but after thinking about it more I disagree. `feemodel` module is responsible exclusively for managing fee models - nothing more. It means that the only params managed by this module are those related to fee model, and all of them are stored inside `Model` structure. There won't be any no-fee params delivered by this module and managed by governance. Saying that, I don't see any reason to introduce additional level in this structure.
</blockquote></details>

Disagree, this with that point or view, the params if more generic abstraction, and here the Model is either the param, or all models params should be directly in the Params. 
___
*[`proto/coreum/feemodel/v1/model.proto` line 18 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmx2WdAJwHqAsf7y-y:-NAo9iDm98RPjNlQrEn9:b-3fdmgx) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/model.proto#L18)):*

it's already inside `feemodel` namespace
</blockquote></details>

I'm ok, to live with it, but still not comfortable name. 
___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NAo9uOwBH4wRUr6mldL:bq7l3gi) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*

That's why I was againt introducing `DefaultModel` when @ysv proposed it.

Our config for all the networks go through `NetworkConfig` type. This is the ground truth for the network settings. The fact that `FeeModel` config inside `NetworkConfig` is set by using `feemodel.DefaultModel` is another step of getting values common to all the networks.

I was very sure that adding `feemodel.DefaultModel()` will create a lot of confusion. The facts are like this:
- config of the network used to configure genesis comes from `NetworkConfig`
- `feemodel.DefaultModel` is a weird place to store values common to all our network configs. It is weird place because module package (which is a generic code to be reused by any blockchain) should not contain network specific settings.
</blockquote></details>

But that aproach is different from the approaches of other modules we use. 
Now we generate the defaults and then replace with the values we need in the template.
But now you generate with the values you want and then replace to the same values, which I wired and different from the approach we wuse for other modules.

dzmitryhil (commented):
> Reviewed 17 of 19 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/feemodel/v1/genesis.proto` line 12 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmwiMV8WL9gMREgoNn:-NAoAwPW6D0oippHkLEe:bfnhvix) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/genesis.proto#L12)):*

Disagree, this with that point or view, the params if more generic abstraction, and here the Model is either the param, or all models params should be directly in the Params. 
</blockquote></details>

I may do it but I don't think it's needed and I don't like to add logic to the code which is not really used. @ysv @miladz68 WDYT?
___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NAoBMB2CGKpREPYcD8N:b-8rxunw) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*

But that aproach is different from the approaches of other modules we use. 
Now we generate the defaults and then replace with the values we need in the template.
But now you generate with the values you want and then replace to the same values, which I wired and different from the approach we wuse for other modules.
</blockquote></details>

Ah, now I understand your point. BEcause this code is for generating default genesis, not the one used by our network. I will change it.
BTW we never use this functionality (generating default genesis) because we reimplemented `cored init` command and it just dumps predefined genesis.json for selected network.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NAoLWrg7jiigxRq1TPg:b-knras4) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*

Ah, now I understand your point. BEcause this code is for generating default genesis, not the one used by our network. I will change it.
BTW we never use this functionality (generating default genesis) because we reimplemented `cored init` command and it just dumps predefined genesis.json for selected network.
</blockquote></details>

Right and the defauild come from the template.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NAoLlpTEsmUJRVKB5KX:buk430t) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*

Right and the defauild come from the template. 
</blockquote></details>

* defaults

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 17 of 19 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NArfd5E83NfMgYEZPpS:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*

* defaults
</blockquote></details>

Done.

dzmitryhil (commented):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/feemodel/v1/model.proto` line 18 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmx2WdAJwHqAsf7y-y:-NAwgp4TCVhXYKg_ke4g:bvcwn6z) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/model.proto#L18)):*

I'm ok, to live with it, but still not comfortable name. 
</blockquote></details>

Let's rename to Params, and I'll resolve the thread
___
*[`x/feemodel/module.go` line 148 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmw6xeC-ViD-aHKM4k:-NAwgt9w1AusbRr-VkgG:b-xj6gf) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L148)):*
> Let's rename to Params, and I'll resolve the thread
Let's rename to SetParams (+ GetParams), and I'll resolve the thread.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 6 of 19 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, and @ysv)

___
*[`x/feemodel/module.go` line 148 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmw6xeC-ViD-aHKM4k:-NAxuU65BWuKwRd9ecP3:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L148)):*

> Let's rename to Params, and I'll resolve the thread
Let's rename to SetParams (+ GetParams), and I'll resolve the thread.
</blockquote></details>

Done.
___
*[`proto/coreum/feemodel/v1/model.proto` line 18 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmx2WdAJwHqAsf7y-y:-NAxuOET3h0_pdn0Sa-L:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/model.proto#L18)):*

Let's rename to Params, and I'll resolve the thread
</blockquote></details>

Done.

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 6 of 19 files reviewed, all discussions resolved (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, and @ysv)

ysv (changes requested):
> Reviewed 2 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 6 of 19 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, and @wojtek-coreum)

___
*[`proto/coreum/feemodel/v1/genesis.proto` line 12 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmwiMV8WL9gMREgoNn:-NAy84dCAUoggv34sIn9:b-cswr64) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/proto/coreum/feemodel/v1/genesis.proto#L12)):*

I may do it but I don't think it's needed and I don't like to add logic to the code which is not really used. @ysv @miladz68 WDYT?
</blockquote></details>

ok
(discussed in slack)
___
*[`x/feemodel/module.go` line 63 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAmvfMz5DCq4KKJxI-C:-NAy7M_o5akmRc95mAmr:bchwhjp) ([raw file](https://github.com/coreumfoundation/coreum/blob/ffe0eaca2876f022ff9b5108f94584f5995ebb40/x/feemodel/module.go#L63)):*

Done.
</blockquote></details>

ok

That is why I wanted to have these default params
I didn't mean these values to be used as defaults for our network. 
But our module should self-contain everything needed like all other modules & have reasonable default values
___
*[`x/feemodel/types/model_sim_test.go` line 57 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyqifpBdjwaXmTYecH:-NAyqifpBdjwaXmTYecI:b-c9n6rz) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/model_sim_test.go#L57)):*
> ```Go
> 
> 	for i, gas := range blockGas {
> 		shortAverage = CalculateEMA(shortAverage, gas, feeModelSim.ShortAverageBlockLength)
> ```

at some places we call it EMA but in other we try to avoid using specific average kind and just use Average
I think we should stick to a single version everywhere: 
- EMA (func CalculateEMA, fields in struct: LongEMABlockLength, ShortEMABlockLength, func args: longEMA, shortEMA ...)
- MovingAverage (func CalculateMovingAverage, LongeMovingAverageBlockLength ...)

I prefer first version
___
*[`x/feemodel/types/params.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyjzmdC8a_AefsC_PE:-NAyjzmdC8a_AefsC_PF:b-tzip4y) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L11)):*
> ```Go
> 
> // String implements the stringer interface.
> func (m Params) String() string {
> ```

as discussed on the call:
```
type FeeModel {
  params Params
}

// business logic funcs
func (model FeeModel) CalculateNextGasPrice(params, ...)
```
(names are subject for changes)

so the idea is to just wrap params into other struct and define all business logic on top of this struct. Because currently we have mix of business logic & cosmos-sdk type in a single place
___
*[`x/feemodel/types/params.go` line 111 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyPYMk8BQqmBgWs2YB:-NAyPYMk8BQqmBgWs2YC:bfp7eb5) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L111)):*
> ```Go
> }
> 
> func (m Params) calculateNextGasPriceInEscalationRegion(shortEMA int64) sdk.Int {
> ```

receiver name is `m` but struct is named Params
___
*[`x/feemodel/types/params_test.go` line 20 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAys7eq9VDQJ6OT65aH:-NAys7erBec10_1XFOQO:b6onkic) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params_test.go#L20)):*
> ```Go
> 		EscalationStartBlockGas: 700,
> 		MaxBlockGas:             1000,
> 		ShortAverageBlockLength: 10,
> ```

non-blocking:
IDK is it only for me but `ShortAverageLengthInBlocks` seems more readable name

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 6 of 19 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

___
*[`x/feemodel/types/params.go` line 111 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyPYMk8BQqmBgWs2YB:-NBBGQ4HC9ezeBvEYpYZ:b-aspym4) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L111)):*

receiver name is `m` but struct is named Params
</blockquote></details>

this is how code is generated from proto and receiver name must be the same everyhere
___
*[`x/feemodel/types/params_test.go` line 20 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAys7eq9VDQJ6OT65aH:-NBBGbnaCBF7mOOzU26e:b8jcmhm) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params_test.go#L20)):*

non-blocking:
IDK is it only for me but `ShortAverageLengthInBlocks` seems more readable name
</blockquote></details>

I think we all agreed on the current names previously, it would be 4th time when we rename them

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 6 of 19 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: 5 of 27 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @oliverasistemas, and @ysv)

___
*[`x/feemodel/types/model_sim_test.go` line 57 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyqifpBdjwaXmTYecH:-NBBdJDD1rRyEndfaid0:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/model_sim_test.go#L57)):*

at some places we call it EMA but in other we try to avoid using specific average kind and just use Average
I think we should stick to a single version everywhere: 
- EMA (func CalculateEMA, fields in struct: LongEMABlockLength, ShortEMABlockLength, func args: longEMA, shortEMA ...)
- MovingAverage (func CalculateMovingAverage, LongeMovingAverageBlockLength ...)

I prefer first version
</blockquote></details>

Done.
___
*[`x/feemodel/types/params.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyjzmdC8a_AefsC_PE:-NBBPGAlB7Mh7DF5xrr_:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L11)):*

as discussed on the call:
```
type FeeModel {
  params Params
}

// business logic funcs
func (model FeeModel) CalculateNextGasPrice(params, ...)
```
(names are subject for changes)

so the idea is to just wrap params into other struct and define all business logic on top of this struct. Because currently we have mix of business logic & cosmos-sdk type in a single place


</blockquote></details>

Done.

dzmitryhil (dismissed):
> Reviewed 1 of 1 files at r4, 2 of 14 files at r5, 19 of 19 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, and @ysv)

ysv (changes requested):
> Reviewed 3 of 21 files at r1, 1 of 1 files at r4, 2 of 14 files at r5, 19 of 19 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, and @wojtek-coreum)

___
*[`x/feemodel/types/model_sim_test.go` line 57 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyqifpBdjwaXmTYecH:-NBC0JaV7oqUTD8O94zG:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/model_sim_test.go#L57)):*

Done.
</blockquote></details>

ok
___
*[`x/feemodel/types/params.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyjzmdC8a_AefsC_PE:-NBCGxLR9KBrsGmepvWd:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L11)):*

Done.
</blockquote></details>

ok
___
*[`x/feemodel/types/params.go` line 111 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyPYMk8BQqmBgWs2YB:-NBC0lDNDCJckCQQMwmE:b-5bjfja) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L111)):*

this is how code is generated from proto and receiver name must be the same everyhere
</blockquote></details>

That is strange because for me `p` seems much more logical shortcut for Params

I think that it is leftover because of initial struct name `Model` -> `m`.
And I assume that they do some magic when generating from to keep name if it is already defined.
Maybe try to remove this file fully and regenerate ?

I also see in std sdk module that they have `p`

Screenshot 2022-09-05 at 11.43.30

Same for distribution cosmos-sdk module
___
*[`x/feemodel/types/params.go` line 29 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NBCGhD11Hu75Xb-6ANN:-NBCGhD11Hu75Xb-6ANO:b-77w6ut) ([raw file](https://github.com/coreumfoundation/coreum/blob/3056c6d0993a11c5115a6fa8d28b1b48aeab0206/x/feemodel/types/params.go#L29)):*
> ```Go
> 		paramtypes.NewParamSetPair([]byte("EscalationStartBlockGas"), &m.EscalationStartBlockGas, modelValidator),
> 		paramtypes.NewParamSetPair([]byte("MaxBlockGas"), &m.MaxBlockGas, modelValidator),
> 		paramtypes.NewParamSetPair([]byte("ShortEmaBlockLength"), &m.ShortEmaBlockLength, modelValidator),
> ```

can we somehow define EMA as acronym in proto ?
I prefer we had it named `ShortEMABlockLength`

Because we use this acronym everywhere except here
___
*[`x/feemodel/types/params_test.go` line 20 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAys7eq9VDQJ6OT65aH:-NBCH3iK9KZzxIgT2lni:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params_test.go#L20)):*

I think we all agreed on the current names previously, it would be 4th time when we rename them
</blockquote></details>

ok

dzmitryhil (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, and @wojtek-coreum)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @oliverasistemas, and @ysv)

___
*[`x/feemodel/types/params.go` line 111 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyPYMk8BQqmBgWs2YB:-NBCZZ8aE9FM2lLfbkdV:b-owck8p) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L111)):*

That is strange because for me `p` seems much more logical shortcut for Params

I think that it is leftover because of initial struct name `Model` -> `m`.
And I assume that they do some magic when generating from to keep name if it is already defined.
Maybe try to remove this file fully and regenerate ?

I also see in std sdk module that they have `p`

Screenshot 2022-09-05 at 11.43.30

Same for distribution cosmos-sdk module


</blockquote></details>

But in all the generated `*.pb.go` files `m` is used (it is taken from "message" word). We have a linter which checks that the same receiver is used in all places. That's why I need to follow this convention.
___
*[`x/feemodel/types/params.go` line 29 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NBCGhD11Hu75Xb-6ANN:-NBC_HNR2kaHwZbBja7l:b-bsy1c3) ([raw file](https://github.com/coreumfoundation/coreum/blob/3056c6d0993a11c5115a6fa8d28b1b48aeab0206/x/feemodel/types/params.go#L29)):*

can we somehow define EMA as acronym in proto ?
I prefer we had it named `ShortEMABlockLength`

Because we use this acronym everywhere except here
</blockquote></details>

According to this discussion: https://github.com/golang/protobuf/issues/53

It seems to be impossible

ysv (approved):
> Reviewed 8 of 8 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/149)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @begmaroman, @miladz68, and @oliverasistemas)

___
*[`x/feemodel/types/params.go` line 111 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NAyPYMk8BQqmBgWs2YB:-NBDBcx42sc0X27ZF81d:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/ee73d7925a0f3a24648677d57f890d4d417cfb22/x/feemodel/types/params.go#L111)):*

But in all the generated `*.pb.go` files `m` is used (it is taken from "message" word). We have a linter which checks that the same receiver is used in all places. That's why I need to follow this convention.
</blockquote></details>

ok
___
*[`x/feemodel/types/params.go` line 29 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/149#-NBCGhD11Hu75Xb-6ANN:-NBDBf0J8NHCOTtOI_He:b-72jvvb) ([raw file](https://github.com/coreumfoundation/coreum/blob/3056c6d0993a11c5115a6fa8d28b1b48aeab0206/x/feemodel/types/params.go#L29)):*

According to this discussion: https://github.com/golang/protobuf/issues/53

It seems to be impossible
</blockquote></details>

unfortunately
ok

PR #153, opened 2022-08-31
refactored new tx package

miladz68:
The new tx package is supposed to provide helper functions to broadcast and await on tx inclusion in the block, and will replace the client package completely down the line.

- - -
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)


miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArlT3t2pl73_fIzixj:-NArlT3t2pl73_fIzixk:bu1hzuz) (no related file):*
I should mention that this PR contains some changes to the client package as well. I was hoping to remove the client package usage but realised it would take too long so stopped mid way. I have also tried to remove the client code usage in the integration tests as much as possible. 

___

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 13 files reviewed, 8 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 107 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArmE5f6Lbn1N9CmGrP:-NArmE5f6Lbn1N9CmGrQ:b-es6xjn) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/bank/transfer.go#L107)):*
> ```Go
> }
> 
> func queryBankBalances(ctx context.Context, clientCtx client.Context, wallet types.Wallet) (map[string]sdk.Coin, error) {
> ```

Is the "types" removal is out of scope for that PR ?
___
*[`integration-tests/bank/transfer.go` line 110 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArllaRBgQcwlX4ap01:-NArllaRBgQcwlX4ap02:b-oe6pqo) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/bank/transfer.go#L110)):*
> ```Go
> 	requestCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
> 	defer cancel()
> 	bankQueryClient := banktypes.NewQueryClient(clientCtx)
> ```

Does it mean that for each integration test we will have to create a new instance of the client that way? 



_Code quote (from integration-tests/auth/sequence.go):_
```Go
ecp256k1.PrivKey{Key: sender.Key}
```
___
*[`integration-tests/testing/types.go` line 27 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArn8TdEAqznzsUlPBt:-NArn8TdEAqznzsUlPBu:b-1xb20m) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/testing/types.go#L27)):*
> ```Go
> type Chain struct {
> 	NetworkConfig app.NetworkConfig
> 	ClientCtx     cosmosclient.Context
> ```

Did you check what is inside the "cosmosclient.Context" ?  
There are some helper functions which might be used. 
For example:
1)
```
ctx.AccountRetriever
```
And use can use it like this:
```
WithAccountRetriever(authtypes.AccountRetriever{}).
```
Which you can use to get the Account info.
2) 
Also here is:
```
BroadcastMode     string
```
And use can use it like this:
```
WithBroadcastMode(flags.BroadcastBlock).
```
Which can be used to await the block. 
-----
So I don't actually understand why we need the all those functions. 

Also there are a lot for other .With** useful functions.
___
*[`pkg/tx/account.go` line 13 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArka5t0L6tjFk0HDtQ:-NArka5t0L6tjFk0HDtR:bsu1sh1) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/account.go#L13)):*
> ```Go
> 
> // GetAccountInfo returns account number and account sequence for provided address
> func GetAccountInfo(
> ```

Shouild it be exported?
___
*[`pkg/tx/broadcast.go` line 1 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArk_OqEz_caeW7v1C_:-NArk_OqEz_caeW7v1Ca:b-l0xlo0) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/broadcast.go#L1)):*
> ```Go
> package tx
> ```

What is the purpose to keep on function in file? From what I see, we can merge them all into one file.
___
*[`pkg/tx/broadcast.go` line 30 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArkCQH4TTGs2rV1wEN:-NArkCQH4TTGs2rV1wEO:bu7v2fj) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/broadcast.go#L30)):*
> ```Go
> // BroadcastAsync sends transaction to chain, ensuring it passes CheckTx.
> // Doesn't await for Tx being included in a block.
> func BroadcastAsync(
> ```

What about the Sync version? Based on the use-cases we have now, most of them need the sync version.
___
*[`pkg/tx/types.go` line 14 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArl0UGA2c5e_N1oro1:-NArl0UGA2c5e_N1oro2:btkhvjt) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/types.go#L14)):*
> ```Go
> 	GasLimit    uint64
> 	GasPrice    sdk.Coin
> 	Memo        string
> ```

Do you use it anywhere?

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 13 files reviewed, 8 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 107 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArmE5f6Lbn1N9CmGrP:-NArr4RJ8Lewriwf6jhb:b32wsm) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/bank/transfer.go#L107)):*

Is the "types" removal is out of scope for that PR ?
</blockquote></details>

no
___
*[`integration-tests/bank/transfer.go` line 110 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArllaRBgQcwlX4ap01:-NArrBQ9-fPLE8nPd2Kf:bpvvt36) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/bank/transfer.go#L110)):*

Does it mean that for each integration test we will have to create a new instance of the client that way? 
</blockquote></details>

yes, but I think you quoted the wrong code.
___
*[`integration-tests/testing/types.go` line 27 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArn8TdEAqznzsUlPBt:-NArrzI3A2iwdCbl39Aq:b-asx4f) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/testing/types.go#L27)):*

Did you check what is inside the "cosmosclient.Context" ?  
There are some helper functions which might be used. 
For example:
1)
```
ctx.AccountRetriever
```
And use can use it like this:
```
WithAccountRetriever(authtypes.AccountRetriever{}).
```
Which you can use to get the Account info.
2) 
Also here is:
```
BroadcastMode     string
```
And use can use it like this:
```
WithBroadcastMode(flags.BroadcastBlock).
```
Which can be used to await the block. 
-----
So I don't actually understand why we need the all those functions. 

Also there are a lot for other .With** useful functions.
</blockquote></details>


I see that AccountRetriver could be used and GetAccoundInfo is redundant. the `flag.BroadcastBlock` flag is broken and may timeout, it is commented in the code and they encourage not to use it. so wojtek and max agreed to have this await function to be on the safe side. Moreover the broadcast function does not just broadcast, but also signs and encodes the message, so it is 3 steps in one. 
___
*[`pkg/tx/account.go` line 13 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArka5t0L6tjFk0HDtQ:-NArtMsCEkwkf4BTD_R1:b-dxku32) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/account.go#L13)):*

Shouild it be exported?
</blockquote></details>

yes, it is used on integration tests. however it will be removed according to your previous comment.
___
*[`pkg/tx/broadcast.go` line 1 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArk_OqEz_caeW7v1C_:-NArtcRo3mTRaKKKtSKB:b1wxz3a) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/broadcast.go#L1)):*

What is the purpose to keep on function in file? From what I see, we can merge them all into one file.
</blockquote></details>

I prefer to have them separated by the functionality they do so we will not have one large file. 
___
*[`pkg/tx/broadcast.go` line 30 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArkCQH4TTGs2rV1wEN:-NArtndE4QZNNq3AC4Sg:b-3ahjxl) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/broadcast.go#L30)):*

What about the Sync version? Based on the use-cases we have now, most of them need the sync version.
</blockquote></details>

the sync version is just 2 calls to broadcast async and await tx. that is why I did not add it. 
___
*[`pkg/tx/types.go` line 14 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArl0UGA2c5e_N1oro1:-NAru4QW3JK6oq-ZsUaO:b3n6mc2) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/types.go#L14)):*

Do you use it anywhere?
</blockquote></details>

yes, in tests.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 13 files reviewed, 9 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 117 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAs1ZkzCoELAhJ5dEAL:-NAs1ZkzCoELAhJ5dEAM:b-t2aepc) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/bank/transfer.go#L117)):*
> ```Go
> 	}
> 
> 	balances := map[string]sdk.Coin{}
> ```

Use the sdk.Coins instead. It contains the .AmountOf function which you can use.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 14 files reviewed, 9 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 117 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAs1ZkzCoELAhJ5dEAL:-NAs3m-b9Q2YQOYszmaQ:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/integration-tests/bank/transfer.go#L117)):*

Use the sdk.Coins instead. It contains the .AmountOf function which you can use.
</blockquote></details>

Done.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 14 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/broadcast.go` line 30 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArkCQH4TTGs2rV1wEN:-NAsfv-uAn-pp8G2VbkB:b-i94ij6) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/broadcast.go#L30)):*

the sync version is just 2 calls to broadcast async and await tx. that is why I did not add it. 
</blockquote></details>

I know, but we do it almost every time we call the BroadcastAsync

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/types.go` line 18 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAsguIgCo9Z0ySVM00_:-NAsguIgCo9Z0ySVM00a:bwocazf) ([raw file](https://github.com/coreumfoundation/coreum/blob/e92b96f02e8e7ba16b04bd4a1e8eba4100da398d/pkg/tx/types.go#L18)):*
> ```Go
> 
> // AccountInfo contains account number and sequence used to sign a tx
> type AccountInfo struct {
> ```

I don't realy like the idea to keep in the "SignInput", because the only use case of it, form what I see is to tests the "TestUnexpectedSequenceNumber".
It's possible to rewite that test, and remove it here.



_Code quote:_
```Go
AccountInfo
```

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/broadcast.go` line 30 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NArkCQH4TTGs2rV1wEN:-NAshow2DzZ8Fs6tcHya:b-jtqfpe) ([raw file](https://github.com/coreumfoundation/coreum/blob/b175097d7dbe3939c909c1c2057d2d2aee519c07/pkg/tx/broadcast.go#L30)):*

I know, but we do it almost every time we call the BroadcastAsync
</blockquote></details>

that's ok, I will add the sync verision, it should be simple.
___
*[`pkg/tx/types.go` line 18 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAsguIgCo9Z0ySVM00_:-NAshdGjEMYc3FwWF-a2:b-ofj50l) ([raw file](https://github.com/coreumfoundation/coreum/blob/e92b96f02e8e7ba16b04bd4a1e8eba4100da398d/pkg/tx/types.go#L18)):*

I don't realy like the idea to keep in the "SignInput", because the only use case of it, form what I see is to tests the "TestUnexpectedSequenceNumber".
It's possible to rewite that test, and remove it here.
</blockquote></details>

it is actually an optimization step, so we can manually increment this and not ask for it via query on every request.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/types.go` line 18 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAsguIgCo9Z0ySVM00_:-NAspsmR3xAK5cmw3Fzz:becbh35) ([raw file](https://github.com/coreumfoundation/coreum/blob/e92b96f02e8e7ba16b04bd4a1e8eba4100da398d/pkg/tx/types.go#L18)):*

it is actually an optimization step, so we can manually increment this and not ask for it via query on every request.
</blockquote></details>

But we don't have a use-case for it now, so nothing to optimize, right?

dzmitryhil (commented):
> Reviewed 6 of 13 files at r1, 3 of 5 files at r2, 1 of 1 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

integration-tests/integration_test.go
```diff
@@ -157,35 +175,45 @@ func (tf *testingFaucet) FundAccounts(ctx context.Context, accountsToFund ...cor
...
+		msgList = append(msgList, msg)
+	}
+
+	signInput := tx.SignInput{
+		PrivateKey: fundingPrivateKey,
+		GasLimit:   cfg.NetworkConfig.Fee.DeterministicGas.BankSend * uint64(len(accountsToFund)),
+		GasPrice:   gasPrice,
+	}
+
+	txHash, err := tx.BroadcastAsync(ctx, cfg.ClientContext, signInput, msgList...)
```
dzmitryhil, line 206: We can use the "BroadcastSync" here.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/errors.go` line 27 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAt17I4CvQpXCYcH7uY:-NAt17I4CvQpXCYcH7uZ:bifjmrm) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcce5017a0d607bca806d2888ff3f788a82d7a67/pkg/tx/errors.go#L27)):*
> ```Go
> 
> // ExpectedSequenceFromError checks if error is related to account sequence mismatch, and returns expected account sequence
> func ExpectedSequenceFromError(err error) (uint64, bool, error) {
> ```

You don't use the "ExpectedSequenceFromError", "IsInsufficientFeeError", "asSDKError" funcs.
And the "expectedSequenceRegExp" as well - if you remove them.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`integration-tests/integration_test.go` line 206 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAoRqXRQC2uqRr525kq-r4-206:-NAtGxiT1QF3cx0Wh1ii:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcce5017a0d607bca806d2888ff3f788a82d7a67/integration-tests/integration_test.go#L206)):*

We can use the "BroadcastSync" here.
</blockquote></details>

Done.
___
*[`pkg/tx/errors.go` line 27 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAt17I4CvQpXCYcH7uY:-NAtG9Rm1LCm47SjQfoM:b-rxpny0) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcce5017a0d607bca806d2888ff3f788a82d7a67/pkg/tx/errors.go#L27)):*

You don't use the "ExpectedSequenceFromError", "IsInsufficientFeeError", "asSDKError" funcs.
And the "expectedSequenceRegExp" as well - if you remove them. 
</blockquote></details>

yes there are not used here, @wojtek-coreum are you using them anywhere ? or is it OK to remove these ?
___
*[`pkg/tx/types.go` line 18 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAsguIgCo9Z0ySVM00_:-NAtGOqr672fWfrRrtAO:b-qemhfu) ([raw file](https://github.com/coreumfoundation/coreum/blob/e92b96f02e8e7ba16b04bd4a1e8eba4100da398d/pkg/tx/types.go#L18)):*

But we don't have a use-case for it now, so nothing to optimize, right? 
</blockquote></details>

I think we used them in zstress, @wojtek-coreum any objections on removing these ?

wojtek-coreum (changes requested):
> Reviewed 4 of 13 files at r1, 2 of 5 files at r2, 1 of 1 files at r3, 1 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/153)* status: 13 of 14 files reviewed, 19 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`app/context.go` line 15 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxGFnU6JsblU1br69A:-NAxGFnU6JsblU1br69B:b-vjpxi3) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/app/context.go#L15)):*
> ```Go
> 		WithInterfaceRegistry(encodingConfig.InterfaceRegistry).
> 		WithTxConfig(encodingConfig.TxConfig).
> 		WithAccountRetriever(authtypes.AccountRetriever{}).
> ```

this is bas, as described in https://reviewable.io/reviews/CoreumFoundation/coreum/153#-NAxG2LY9YdGNJkF4gzv
___
*[`integration-tests/auth/sequence.go` line 35 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxG2LY9YdGNJkF4gzv:-NAxG2LY9YdGNJkF4gzw:b-s230gd) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/integration-tests/auth/sequence.go#L35)):*
> ```Go
> 	privateKey := secp256k1.PrivKey{Key: sender.Key}
> 	senderAddress := sdk.AccAddress(privateKey.PubKey().Address())
> 	info, err := chain.ClientCtx.AccountRetriever.GetAccount(chain.ClientCtx, senderAddress)
> ```

this is bad because it does not take a `ctx` as an argument. That's why it was implemented using the other client instead of `AccountRetriever`
___
*[`pkg/client/client.go` line 126 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxGZF81RIdW6LMgVGc:-NAxGZF81RIdW6LMgVGd:b9wecu2) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/client/client.go#L126)):*
> ```Go
> 
> // ExpectedSequenceFromError checks if error is related to account sequence mismatch, and returns expected account sequence
> func ExpectedSequenceFromError(err error) (uint64, bool, error) {
> ```

it might be removed together with other functions around
___
*[`pkg/tx/broadcast.go` line 34 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxClQgB986ErgyHLpP:-NAxClQgB986ErgyHLpQ:b-m8wisu) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L34)):*
> ```Go
> 	input SignInput,
> 	msgs ...sdk.Msg,
> ) (resultTx *coretypes.ResultTx, err error) {
> ```

no pointer needed
___
*[`pkg/tx/broadcast.go` line 63 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAx8n912-LZsznkYjzu:-NAx8n92-UbvorO7yKiT:b-tbxai1) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L63)):*
> ```Go
> 	if err != nil {
> 		if errors.Is(err, requestCtx.Err()) {
> 			return txHash, errors.WithStack(err)
> ```

here and everywhere: `txHash` should not be returned in case of error, this is Max's legacy
___
*[`pkg/tx/broadcast.go` line 71 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAx89EN5xlkhy0dbHAA:-NAx89EN5xlkhy0dbHAB:b-8t7gjb) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L71)):*
> ```Go
> 		}
> 	} else if res.Code != 0 {
> 		err := errors.Wrapf(sdkerrors.New(res.Codespace, res.Code, res.Log),
> ```

As @dzmitryhil said in another PR, `sdkerrors.New` shoudl be replaced with `sdkerrors.ABCIError`
___
*[`pkg/tx/broadcast.go` line 76 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxCUKo8n7c7ISMGFrc:-NAxCUKo8n7c7ISMGFrd:b-ijl9ku) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L76)):*
> ```Go
> 	}
> 
> 	return res.Hash.String(), nil
> ```

`txHash` might be reused to not recompute the hash
___
*[`pkg/tx/broadcast.go` line 84 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxCcGu1sLtZZ4NpgmG:-NAxCcGu1sLtZZ4NpgmH:b-m8wisu) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L84)):*
> ```Go
> 	clientCtx client.Context,
> 	txHash string,
> ) (resultTx *coretypes.ResultTx, err error) {
> ```

no pointer needed
___
*[`pkg/tx/broadcast.go` line 105 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxCvUy0fDoYVWAOMlR:-NAxCvUy0fDoYVWAOMlS:b-7ropln) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L105)):*
> ```Go
> 			}
> 
> 			if errRes := checkTendermintError(err, txHash); errRes != nil {
> ```

ditto
___
*[`pkg/tx/broadcast.go` line 117 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAx8OMR9Xdg1MhpyM-M:-NAx8OMR9Xdg1MhpyM-N:b-8t7gjb) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L117)):*
> ```Go
> 		if resultTx.TxResult.Code != 0 {
> 			res := resultTx.TxResult
> 			return errors.Wrapf(sdkerrors.New(res.Codespace, res.Code, res.Log), "transaction '%s' failed", txHash)
> ```

As @dzmitryhil said in another PR, `sdkerrors.New` shoudl be replaced with `sdkerrors.ABCIError`
___
*[`pkg/tx/broadcast.go` line 139 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxD0nG8GmvAkq5bnqc:-NAxD0nG8GmvAkq5bnqd:bbrw3w6) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L139)):*
> ```Go
> ) ([]byte, error) {
> 	fromAddress := sdk.AccAddress(config.PrivateKey.PubKey().Address())
> 	if config.AccountInfo.Number == 0 {
> ```

nope, 0 is a perfect valid number for first transaction executed by the account. It means that if someone sets it to 0, this step here should be skipped. that's why in the original code `config.AccountInfo` was nillable to distinguish between setting values to 0 or not setting them.
___
*[`pkg/tx/broadcast.go` line 165 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxDQmd2aF48klRZQ3S:-NAxDQmd2aF48klRZQ3T:b77vi62) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/broadcast.go#L165)):*
> ```Go
> }
> 
> func isTxInMempool(errRes *sdk.TxResponse) bool {
> ```

After applying changes described in other comments this function might be removed
___
*[`pkg/tx/errors.go` line 27 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAt17I4CvQpXCYcH7uY:-NAxBUX2A_FS398-HI_X:bee92g) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcce5017a0d607bca806d2888ff3f788a82d7a67/pkg/tx/errors.go#L27)):*

yes there are not used here, @wojtek-coreum are you using them anywhere ? or is it OK to remove these ?
</blockquote></details>

it was used by `zstress`, it's no longer in master so it might be removed
___
*[`pkg/tx/errors.go` line 46 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxBa0u40KwjR9vmVbX:-NAxBa0u40KwjR9vmVbY:b-u1j3c) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/errors.go#L46)):*
> ```Go
> 
> // IsInsufficientFeeError returns true if error was caused by insufficient fee provided with the transaction
> func IsInsufficientFeeError(err error) bool {
> ```

Once `sdkerrors.ABCIError` is used instead of `sdkerrors.New` (as described in other comments), it could be simplified to `sdkerrors.ErrInsufficientFee.Is(err)`.
Because we decided to use cosmos sdk types directly, it means this one liner does not deserve its own function here and might be used directly wherever needed.
It means, together with my another comment in `checkTendermintError` function, that functions `isSDKErrorResult` and `asSDKError` might be removed.
___
*[`pkg/tx/errors.go` line 65 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAx905u93DyhajU_7kX:-NAx905vDjBHr1k5M-my:b-b33xmh) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/errors.go#L65)):*
> ```Go
> 
> 	switch {
> 	case strings.Contains(errStr, strings.ToLower(mempool.ErrTxInCache.Error())):
> ```

this is the only relevant thing in this function, so whenever there is code like

```go
		errRes := checkTendermintError(err, txHash)
		if !isTxInMempool(errRes) {
			return txHash, errors.WithStack(err)
		}
```

it might be simplified to

```go
if !strings.Contains(strings.ToLower(err.Error()), strings.ToLower(mempool.ErrTxInCache.Error())) {
    return txHash, errors.WithStack(err)
}
```

Eventually `checkTendermintError` might be rewritten to `convertTendermintError` to provide sdk error (so they later might be compared if needed):

```go
func convertTendermintError(err error) error {
	if err == nil {
		return nil
	}
	errStr := strings.ToLower(err.Error())

	switch {
	case strings.Contains(errStr, strings.ToLower(mempool.ErrTxInCache.Error())):
                return sdkerrors.ErrTxInMempoolCache.Wrap(err.Error())
	case strings.Contains(errStr, "mempool is full"):
                return sdkerrors.ErrMempoolIsFull.Wrap(err.Error())
	case strings.Contains(errStr, "tx too large"):
                return sdkerrors.ErrTxTooLarge.Wrap(err.Error())
	default:
		return err
	}
}
```

and then, this code

```go
		errRes := checkTendermintError(err, txHash)
		if !isTxInMempool(errRes) {
			return txHash, errors.WithStack(err)
		}
```

could be changed to

```go
if !sdkerrors.ErrTxInMempoolCache.Is(err) {
    return txHash, errors.WithStack(err)
}
```

which is the preferred way IM
___
*[`pkg/tx/sign.go` line 21 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAxDafP2M1t_qGdVwSQ:-NAxDafP2M1t_qGdVwSR:b-jz0s5p) ([raw file](https://github.com/coreumfoundation/coreum/blob/cbee7da9b599ec422b8f2c38731cd9042fccfc2a/pkg/tx/sign.go#L21)):*
> ```Go
> 	txBuilder.SetMemo(input.Memo)
> 
> 	if !input.GasPrice.Amount.IsNil() {
> ```

I think it would be better to make `input.GasPrice` nillable
___
*[`pkg/tx/types.go` line 18 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/153#-NAsguIgCo9Z0ySVM00_:-NAxF7vuDyq0FFXr8gVe:btqb3mc) ([raw file](https://github.com/coreumfoundation/coreum/blob/e92b96f02e8e7ba16b04bd4a1e8eba4100da398d/pkg/tx/types.go#L18)):*

I think we used them in zstress, @wojtek-coreum any objections on removing these ?
</blockquote></details>

We use it in two tests: `TestUnexpectedSequenceNumber` and `TestTransferMaximumGas` so we have two use-cases for it.
It was also in `zstress` but it's no longer with us.

PR #155, opened 2022-09-02
Added new Broadcast function in tx package

miladz68:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)


dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 51 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBil5s2J0_VnRv7lAI:-NBBil5s2J0_VnRv7lAJ:bh474s7) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L51)):*
> ```Go
> 	// Create two random wallets
> 	clientCtx := chain.ClientContext
> 	txf, walletGen := setup(clientCtx)
> ```

The main problem in tearm of the complexity here is the support of the prev wallet struct.
___
*[`integration-tests/bank/transfer.go` line 84 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBjtOkBzJVAR6IeRA1:-NBBjtOkBzJVAR6IeRA2:b-3moexc) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L84)):*
> ```Go
> 	clCtx := clientCtx.
> 		WithFromName(sender.Name).
> 		WithFromAddress(senderAddress).
> ```

That seems redundant, just FromName shuold be enough. 

Also we can generate the clientCtx and txFactory for each account we use in the tests (instead of using just one clientCtx).
The  "BroadcastMode" might be set to the default as well.
___
*[`integration-tests/bank/transfer.go` line 91 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBhlZP5L7ibuYNeYnV:-NBBhlZP5L7ibuYNeYnW:b6pgl60) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L91)):*
> ```Go
> 	result, err := tx.BroadcastTx(ctx, clCtx, txf, msg)
> 	require.NoError(t, err)
> 	time.Sleep(4 * time.Second)
> ```

What is the purpose of that sleep?
___
*[`integration-tests/bank/transfer.go` line 141 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBOTVD2uMv-CefS9CT:-NBBOTVD2uMv-CefS9CU:b-e33y36) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L141)):*
> ```Go
> func setup(clientCtx cosmosclient.Context) (tx.Factory, walletGen) {
> 	kr := keyring.NewUnsafe(keyring.NewInMemory())
> 	txf := tx.Factory{}.
> ```

For that case better to create the NewFactory function, and return the new instance each time.
___
*[`integration-tests/bank/transfer.go` line 144 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBiCql80wRyWtnTKIK:-NBBiCql80wRyWtnTKIL:b-qdu3sf) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L144)):*
> ```Go
> 		WithKeybase(kr).
> 		WithChainID(clientCtx.ChainID).
> 		WithTxConfig(clientCtx.TxConfig)
> ```

From what I see you set it here, and additionally adds as an argument for the broadcast, what is the purpose?
___
*[`pkg/tx/tx.go` line 21 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBPK-W7_LaWlWVyMlx:-NBBPK-W7_LaWlWVyMly:bzerst7) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L21)):*
> ```Go
> // NOTE: copied from
> // https://github.com/cosmos/cosmos-sdk/blob/v0.45.2/client/tx/tx.go
> func BroadcastTx(ctx context.Context, clientCtx client.Context, txf tx.Factory, msgs ...sdk.Msg) (*sdk.TxResponse, error) {
> ```

If you desided to go with the "Factory" type as a wrapper under the SDK, why do you required the SDK type "tx.Factory" here ?
___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBBNzb55c6KNPv1gsDA:b-yeyln6) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*
> ```Go
> func prepareFactory(ctx context.Context, clientCtx client.Context, txf tx.Factory) (tx.Factory, error) {
> 	if txf.AccountNumber() == 0 || txf.Sequence() == 0 {
> 		acc, err := GetAccountInfo(ctx, clientCtx, clientCtx.GetFromAddress())
> ```

Since we've found the context issue for the queries as well, we don't need that custom retriever, and can use the default.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 51 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBil5s2J0_VnRv7lAI:-NBCLDpL8mjVSA198cjp:b-4867r0) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L51)):*

The main problem in tearm of the complexity here is the support of the prev wallet struct.
</blockquote></details>

the removing the old types is not in the scope of this PR, the changes here are introduced to show the reviewers how the new tx package will be used in the integration tests. a proper abstraction and a re-work must be introduces in a later PR. 
___
*[`integration-tests/bank/transfer.go` line 84 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBjtOkBzJVAR6IeRA1:-NBCM4OG8BIE39tzcjGI:b-hdxo3b) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L84)):*

That seems redundant, just FromName shuold be enough. 

Also we can generate the clientCtx and txFactory for each account we use in the tests (instead of using just one clientCtx).
The  "BroadcastMode" might be set to the default as well.


</blockquote></details>

FromName is not enough, since the signer and fromAddress can be different accounts. 
and for creating clientCtx for each account, yes we can do it, but it seems beyond the scope of this PR, it is not a proper abstraction but just a showcase of how the new tx package can be used in integration tests. a proper abstraction must be introduced in a separate PR, which also completely removes old tx, old tyeps and client package. 
I remove "BroadcastMode" to the setup function of the test. 
___
*[`integration-tests/bank/transfer.go` line 91 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBhlZP5L7ibuYNeYnV:-NBCINCp7_Sj7eVOxVrn:bxy38ua) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L91)):*

What is the purpose of that sleep?
</blockquote></details>

dirty code, removed.
___
*[`integration-tests/bank/transfer.go` line 141 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBOTVD2uMv-CefS9CT:-NBCJD-aEfHiKM7pVRh_:b1xyoci) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L141)):*

For that case better to create the NewFactory function, and return the new instance each time.
</blockquote></details>

but it is not reused, so putting it in a function does not add value. 
___
*[`integration-tests/bank/transfer.go` line 144 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBiCql80wRyWtnTKIK:-NBCJSdi66p_Xyntg7GO:b-m27mcj) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L144)):*

From what I see you set it here, and additionally adds as an argument for the broadcast, what is the purpose?
</blockquote></details>

can you elaborate more on your question, in sentence`you set it here`, `it` is not very clear to me. 
but to answer generally, we must setup tx factory correctly and some of needed info is available on clientCtx. and also need to pass both clientCtx and TxFactory to the broadcast method. 
___
*[`pkg/tx/tx.go` line 21 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBPK-W7_LaWlWVyMlx:-NBCKpKoEGJl69w3fla2:b-m0tkn0) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L21)):*

If you desided to go with the "Factory" type as a wrapper under the SDK, why do you required the SDK type "tx.Factory" here ?
</blockquote></details>

Done. we don't need the `tx.` here.
___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBCKwJV9WAaXk0xWE0E:b-domfnt) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

Since we've found the context issue for the queries as well, we don't need that custom retriever, and can use the default. 
</blockquote></details>

we need it to make the function signatures correct, so when we refactor the clientCtx, then all the correct Contexts will be used. if we use accountRetreiver then the signature does not support passing Context.

dzmitryhil (changes requested):
> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 3 of 6 files reviewed, 5 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 51 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBil5s2J0_VnRv7lAI:-NBCURmC5KmnjZCt9UbZ:b-dqaied) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L51)):*

the removing the old types is not in the scope of this PR, the changes here are introduced to show the reviewers how the new tx package will be used in the integration tests. a proper abstraction and a re-work must be introduces in a later PR. 
</blockquote></details>

Sure, I left that comment here to 
___
*[`integration-tests/bank/transfer.go` line 84 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBjtOkBzJVAR6IeRA1:-NBCUfrbCKapa_OAJCnQ:b-c6sffu) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L84)):*

FromName is not enough, since the signer and fromAddress can be different accounts. 
and for creating clientCtx for each account, yes we can do it, but it seems beyond the scope of this PR, it is not a proper abstraction but just a showcase of how the new tx package can be used in integration tests. a proper abstraction must be introduced in a separate PR, which also completely removes old tx, old tyeps and client package. 
I remove "BroadcastMode" to the setup function of the test. 
</blockquote></details>

I thought it's an example how other tests would look like. But of the team is OK to with that I'm ok as well.
___
*[`integration-tests/bank/transfer.go` line 141 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBOTVD2uMv-CefS9CT:-NBCVAaHDLjiNMIHJhhl:b-lzlwg5) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L141)):*

but it is not reused, so putting it in a function does not add value. 
</blockquote></details>

As for me the constructor function is better approach, since it garantee that the way of the struct instantiation is safe, especially for the structs with logic inside. But we can consider it as a matter of taste.
___
*[`integration-tests/bank/transfer.go` line 144 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBiCql80wRyWtnTKIK:-NBCVdFG2IBeESmvh8e0:b-2ejbqm) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L144)):*

can you elaborate more on your question, in sentence`you set it here`, `it` is not very clear to me. 
but to answer generally, we must setup tx factory correctly and some of needed info is available on clientCtx. and also need to pass both clientCtx and TxFactory to the broadcast method. 
</blockquote></details>

Got your point.
___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBCVmLJ6bOOwgqIeyIW:b-ie77ph) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

we need it to make the function signatures correct, so when we refactor the clientCtx, then all the correct Contexts will be used. if we use accountRetreiver then the signature does not support passing Context.
</blockquote></details>

Right, as well as other functions we don't see inside. So as for me, once we start the "upgrade" we will decide how to work with it. So for now I personally would work with the standard. But will agree with the team's decision.

dzmitryhil (commented):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 5 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

ysv (changes requested):
> Reviewed 2 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, and @wojtek-coreum)

___
*[`pkg/tx/tx.go` line 14 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCMNojBQE4pw9CmsU2:-NBCMNok1MJfmm4cO-o_:bv73dmt) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L14)):*
> ```Go
> 
> // Factory is a re-export of the cosmos sdk tx.Factory type, to make usage of this package more convenient.
> type Factory = tx.Factory
> ```

It makes sense to make type aliases if we have a few types of underlying pkg which we can hide

But it this case will anyways not be able to hide all cosmos types. E.g we still return *sdk.TxResponse use sdk.Msgs etc. 
So it doesn't make sense for me

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 84 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBjtOkBzJVAR6IeRA1:-NBCaTjY71rV_GxzysFn:bw6l4an) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L84)):*

I thought it's an example how other tests would look like. But of the team is OK to with that I'm ok as well.
</blockquote></details>

I see your point, I move this logic inside the generator, so it does not need to be set here. 
___
*[`integration-tests/bank/transfer.go` line 141 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBOTVD2uMv-CefS9CT:-NBCacmf6p935ElJXJ5h:b-yk7uoz) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/bank/transfer.go#L141)):*

As for me the constructor function is better approach, since it garantee that the way of the struct instantiation is safe, especially for the structs with logic inside. But we can consider it as a matter of taste.
</blockquote></details>

where would you rather have that `new` constructor, in integration test folder, or in tx package ? if in the integration test, then it will not be exported and is not really a constructor, if in the tx package, that means we are moving away from cosmos sdk way of doing things. 
___
*[`pkg/tx/tx.go` line 14 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCMNojBQE4pw9CmsU2:-NBCb6J07s0L8l8hUOGs:bfpddy1) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L14)):*

It makes sense to make type aliases if we have a few types of underlying pkg which we can hide

But it this case will anyways not be able to hide all cosmos types. E.g we still return *sdk.TxResponse use sdk.Msgs etc. 
So it doesn't make sense for me 
</blockquote></details>

I am not trying to hide cosmos sdk types here, but make the usage of this package easier. without this line, the user of the package will need to import 2 packages named tx, one is our package and another from cosmos sdk and there will be package name collisions and the user will need to use import name alias. so it saves a lot of ugly code on the user side of this package. 
___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBCb_QyEGJgCli_mAOj:b-qab57s) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

Right, as well as other functions we don't see inside. So as for me, once we start the "upgrade" we will decide how to work with it. So for now I personally would work with the standard. But will agree with the team's decision.
</blockquote></details>

@ysv @wojtek-coreum 
your opinions ?

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`integration-tests/integration_test.go` line 40 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCcRv21zmJ-CQ00uvn:-NBCfgRDCUYbNc_USnzh:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/integration_test.go#L40)):*

I suggest to scope out integration test changes from this PR
We saw example of usage with new Broadcast method and we see some issues with it

Lets revert integration-test part and do it separately
</blockquote></details>

Done.
___
*[`pkg/tx/tx.go` line 14 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCMNojBQE4pw9CmsU2:-NBCfhMt1-UrDjWq5T5E:bfmn1om) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L14)):*

I see 
then ok
Probably description needs to have what you described in comment
</blockquote></details>

Done. Added a little more comment

ysv (approved):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 5 of 6 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`integration-tests/integration_test.go` line 40 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCcRv21zmJ-CQ00uvn:-NBCfmJk68Qr45DJhCx4:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/integration-tests/integration_test.go#L40)):*

I suggest to scope out integration test changes from this PR
We saw example of usage with new Broadcast method and we see some issues with it

Lets revert integration-test part and do it separately
</blockquote></details>

ok
___
*[`pkg/tx/tx.go` line 19 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCv8Vr8VOSs59_Z_6R:-NBCv8Vr8VOSs59_Z_6S:bpxcwcu) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L19)):*
> ```Go
> // given set of messages. It will also simulate gas requirements if necessary.
> // It will return an error upon failure.
> // NOTE: copied from
> ```

minor:
but it is not just copied there are some changes done
___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBCh4SGDMs-yTgx_gTu:bnjile0) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

@ysv @wojtek-coreum 
your opinions ?
</blockquote></details>

as far as I remember another Wojtek's concern here was that it overrides AccountNumber & Sequence if they are 0 which are valid values

lets keep this code here

dzmitryhil (dismissed):
> Reviewed 1 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 1 unresolved discussion (waiting on @begmaroman, @miladz68, and @wojtek-coreum)

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 1 unresolved discussion (waiting on @begmaroman, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/tx.go` line 19 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCv8Vr8VOSs59_Z_6R:-NBCx1wn342TkFEEV2L-:b9y7pj4) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L19)):*

minor:
but it is not just copied there are some changes done
</blockquote></details>

added comment to mention that there are changes introduced.

ysv (changes requested):
> Reviewed 1 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, and @wojtek-coreum)

___
*[`pkg/tx/tx.go` line 22 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBD6enF-3NtVOCXFxcW:-NBD6enF-3NtVOCXFxcX:b-q5dz58) ([raw file](https://github.com/coreumfoundation/coreum/blob/4d9c33afa80d21455d1e0f7ba607f2b6998ee9a5/pkg/tx/tx.go#L22)):*
> ```Go
> // NOTE: copied from
> // https://github.com/cosmos/cosmos-sdk/blob/v0.45.2/client/tx/tx.go
> func BroadcastTx(ctx context.Context, clientCtx client.Context, txf Factory, msgs ...sdk.Msg) (*sdk.TxResponse, error) {
> ```

can you add TODO to add test to check if client respects ctx
___
*[`pkg/tx/tx.go` line 45 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBD6VrP6g31zvqmyoTC:-NBD6VrP6g31zvqmyoTD:b-nrjzfo) ([raw file](https://github.com/coreumfoundation/coreum/blob/4d9c33afa80d21455d1e0f7ba607f2b6998ee9a5/pkg/tx/tx.go#L45)):*
> ```Go
> 
> 	// broadcast to a Tendermint node
> 	return clientCtx.BroadcastTx(txBytes)
> ```

as @dzmitryhil mentioned on the call there should be method which accepts ctx and we should use it

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/tx.go` line 22 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBD6enF-3NtVOCXFxcW:-NBDEENO0ccDqb48ryHV:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/4d9c33afa80d21455d1e0f7ba607f2b6998ee9a5/pkg/tx/tx.go#L22)):*

can you add TODO to add test to check if client respects ctx
</blockquote></details>

Done.
___
*[`pkg/tx/tx.go` line 45 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBD6VrP6g31zvqmyoTC:-NBDECYGAlrFe_voXxrL:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/4d9c33afa80d21455d1e0f7ba607f2b6998ee9a5/pkg/tx/tx.go#L45)):*

as @dzmitryhil mentioned on the call there should be method which accepts ctx and we should use it
</blockquote></details>

Done.

wojtek-coreum (changes requested):
> Reviewed 2 of 6 files at r1, 3 of 3 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 6 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBGReAE5tYzEcrgwA1H:b-rt76pz) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

as far as I remember another Wojtek's concern here was that it overrides AccountNumber & Sequence if they are 0 which are valid values

lets keep this code here
</blockquote></details>

Yes, this is bad behavior IMO that in case of 0s, blockchain is queried. There are two situations:
- account number and/or account sequence is not set --> query
- both are set (even if 0s) -> don't query

That's why this part of design in tx factory is broken and I implemented it differently.
___
*[`pkg/tx/tx.go` line 24 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGSIjp8kdhnZbaiR_z:-NBGSIjp8kdhnZbaiRa-:b-vvfuhh) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L24)):*
> ```Go
> // https://github.com/cosmos/cosmos-sdk/blob/v0.45.2/client/tx/tx.go
> // TODO: add test to check if client respects ctx.
> func BroadcastTx(ctx context.Context, clientCtx client.Context, txf Factory, msgs ...sdk.Msg) (*sdk.TxResponse, error) {
> ```

I don't know the scope of this PR but awaiting tx using polling here or later because the built-in even-based method may cause troubles. Once it is done `flags.BroadcastBlock` is not needed.
___
*[`pkg/tx/tx.go` line 35 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGRN1LDuGIeTjnD1CZ:-NBGRN1LDuGIeTjnD1C_:b6t5hj8) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L35)):*
> ```Go
> 	}
> 
> 	unsignedTx.SetFeeGranter(clientCtx.GetFeeGranterAddress())
> ```

Is it sth we are going to have on our chain, it's related to `authz` module right?
___
*[`pkg/tx/tx.go` line 55 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGSmtl8Epsv9LsSb9c:-NBGSmtl8Epsv9LsSb9d:b-hrei3f) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L55)):*
> ```Go
> 		return sdk.NewResponseFormatBroadcastTx(res), nil
> 
> 	case flags.BroadcastAsync:
> ```

Do we need to include this? It's the method when even basic validation is not executed on transaction

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 6 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/tx.go` line 24 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGSIjp8kdhnZbaiR_z:-NBGwTRDBKqELaQp_i3Z:b2gem71) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L24)):*

I don't know the scope of this PR but awaiting tx using polling here or later because the built-in even-based method may cause troubles. Once it is done `flags.BroadcastBlock` is not needed.
</blockquote></details>

It was decided to keep conforming to cosmos SDK's signature so we need to use such flags. regarding the retry, I don't have a preference, but I think relying on grpc connections should be good enough for our use case (which is the faucet)
___
*[`pkg/tx/tx.go` line 35 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGRN1LDuGIeTjnD1CZ:-NBGxRRe3NdcPoJN4NwE:b2nign6) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L35)):*

Is it sth we are going to have on our chain, it's related to `authz` module right?
</blockquote></details>

yes
___
*[`pkg/tx/tx.go` line 55 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGSmtl8Epsv9LsSb9c:-NBGwBxTCKocaBLosw6G:b7j6j96) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L55)):*

Do we need to include this? It's the method when even basic validation is not executed on transaction
</blockquote></details>

we decided to conform to cosmos sdk flow. so we need to have this.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 6 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/tx.go` line 55 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGSmtl8Epsv9LsSb9c:-NBGyZUHB28-hAbiA3cH:b7btmhq) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L55)):*

we decided to conform to cosmos sdk flow. so we need to have this. 
</blockquote></details>

To keep the consistency, I agree we need it.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`pkg/tx/tx.go` line 24 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGSIjp8kdhnZbaiR_z:-NBGz8ZB0zHpKHbUWzOp:b-cgmct0) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L24)):*
>  relying on grpc connections should be good enough

nope, faucet will run on real internet, and a lot of shit happens there

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/tx.go` line 19 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBCv8Vr8VOSs59_Z_6R:-NBGtSK_D2DicDChukKg:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L19)):*

added comment to mention that there are changes introduced.
</blockquote></details>

ok
___
*[`pkg/tx/tx.go` line 22 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBD6enF-3NtVOCXFxcW:-NBGtPDBAEwIm2-imHfJ:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/4d9c33afa80d21455d1e0f7ba607f2b6998ee9a5/pkg/tx/tx.go#L22)):*

Done.
</blockquote></details>

ok
___
*[`pkg/tx/tx.go` line 45 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBD6VrP6g31zvqmyoTC:-NBGtLdRCOxpZBdAi2gU:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/4d9c33afa80d21455d1e0f7ba607f2b6998ee9a5/pkg/tx/tx.go#L45)):*

Done.
</blockquote></details>

ok
___
*[`pkg/tx/tx.go` line 24 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGSIjp8kdhnZbaiR_z:-NBH2hBnBBoT8DIlKac5:b1ld9wb) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L24)):*

>  relying on grpc connections should be good enough

nope, faucet will run on real internet, and a lot of shit happens there
</blockquote></details>

discussed & decided on call
___
*[`pkg/tx/tx.go` line 35 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBGRN1LDuGIeTjnD1CZ:-NBGt1-NBPSZebX-T2cn:bawzg8n) ([raw file](https://github.com/coreumfoundation/coreum/blob/521fce74d01d17f479dde1bbce0384c466fed710/pkg/tx/tx.go#L35)):*

Is it sth we are going to have on our chain, it's related to `authz` module right?
</blockquote></details>

yes the plan is to have it in the future

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBHDIvZ5--UGCWCGXKw:b-lo4vyf) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

Yes, this is bad behavior IMO that in case of 0s, blockchain is queried. There are two situations:
- account number and/or account sequence is not set --> query
- both are set (even if 0s) -> don't query

That's why this part of design in tx factory is broken and I implemented it differently.
</blockquote></details>

but here is what we have in our client now 
so we also have check for 0s there

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

dzmitryhil (commented):
> Reviewed 2 of 3 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

dzmitryhil (approved):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

___
*[`pkg/tx/tx.go` line 191 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBHabaM-l7qN8fX_JEi:-NBHbWns4EKD33ViUEH1:b-x01zec) ([raw file](https://github.com/coreumfoundation/coreum/blob/362ce8d5fe429a970c0dcfce204004af3cdc863a/pkg/tx/tx.go#L191)):*

It requires the original transaction to be passed to it. it would make the AwaitTx signature a little less convenient. 
</blockquote></details>

Ok, got it. Agree if are going to use it directory, better to use the hash. 



_Code quote:_
```Go
txHashBytes
```

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBHczYO8x2wDAojLq6k:b-8jxl3n) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

again, since we will introduce the GetAccountInfo function anyway (because `AccountRetriever` does not accept ctx as input) then we might as well have and use this function so when we fix the issue with queryClient, then all will be fixed.
In other words I am arguing that we will replace the default `AccountRetriever` at some point, And it is ok to do it here. 
</blockquote></details>

@ysv there is a newer implementation in the branch here where it was fixed: https://github.com/CoreumFoundation/coreum/blob/new-client/pkg/client2/client.go#L314
___
*[`pkg/tx/tx.go` line 78 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBHgsvz6nkgUb9KEmio:-NBHgsvz6nkgUb9KEmip:b-f67yvs) ([raw file](https://github.com/coreumfoundation/coreum/blob/362ce8d5fe429a970c0dcfce204004af3cdc863a/pkg/tx/tx.go#L78)):*
> ```Go
> 	case flags.BroadcastBlock:
> 		res, err := clientCtx.Client.BroadcastTxSync(ctx, txBytes)
> 		if err != nil {
> ```

Take a look at original implementation: https://github.com/CoreumFoundation/coreum/blob/new-client/pkg/client2/client.go#L153
There are more edge cases to handle.
___
*[`pkg/tx/tx.go` line 169 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBHfJY2BvrhpnzKxEJ8:-NBHfJY2BvrhpnzKxEJ9:b-xctnmx) ([raw file](https://github.com/coreumfoundation/coreum/blob/362ce8d5fe429a970c0dcfce204004af3cdc863a/pkg/tx/tx.go#L169)):*
> ```Go
> 		if resultTx.TxResult.Code != 0 {
> 			res := resultTx.TxResult
> 			return errors.Wrapf(sdkerrors.New(res.Codespace, res.Code, res.Log), "transaction '%s' failed", txHash)
> ```

sdkerrors.New -> sdkerrors.ABCIError
___
*[`pkg/tx/tx.go` line 199 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBHepBJ8s2RmbIY4azr:-NBHepBJ8s2RmbIY4azs:b-ovvl0y) ([raw file](https://github.com/coreumfoundation/coreum/blob/362ce8d5fe429a970c0dcfce204004af3cdc863a/pkg/tx/tx.go#L199)):*
> ```Go
> 
> 	switch {
> 	case strings.Contains(errStr, strings.ToLower(mempool.ErrTxInCache.Error())):
> ```

this is the only relevant thing in this function, so whenever there is code like

```go
		errRes := checkTendermintError(err, txHash)
		if !isTxInMempool(errRes) {
			return txHash, errors.WithStack(err)
		}
```

it might be simplified to

```go
if !strings.Contains(strings.ToLower(err.Error()), strings.ToLower(mempool.ErrTxInCache.Error())) {
    return txHash, errors.WithStack(err)
}
```

Eventually `checkTendermintError` might be rewritten to `convertTendermintError` to provide sdk error (so they later might be compared if needed):

```go
func convertTendermintError(err error) error {
	if err == nil {
		return nil
	}
	errStr := strings.ToLower(err.Error())

	switch {
	case strings.Contains(errStr, strings.ToLower(mempool.ErrTxInCache.Error())):
                return sdkerrors.ErrTxInMempoolCache.Wrap(err.Error())
	case strings.Contains(errStr, "mempool is full"):
                return sdkerrors.ErrMempoolIsFull.Wrap(err.Error())
	case strings.Contains(errStr, "tx too large"):
                return sdkerrors.ErrTxTooLarge.Wrap(err.Error())
	default:
		return err
	}
}
```


and then, this code

```go
		errRes := checkTendermintError(err, txHash)
		if !isTxInMempool(errRes) {
			return txHash, errors.WithStack(err)
		}
```

could be changed to

```go
if !sdkerrors.ErrTxInMempoolCache.Is(err) {
    return txHash, errors.WithStack(err)
}
```

which is the preferred way IMO

ysv (commented):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

dzmitryhil (commented):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

wojtek-coreum (commented):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 5 of 6 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/tx.go` line 49 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBBNzb48wymZ8uZuH-d:-NBLpR76BRbAlWers9HH:bng6bbd) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3203bd92b620cbc2ac209d6aeb4d09285cdf35e/pkg/tx/tx.go#L49)):*

@ysv there is a newer implementation in the branch here where it was fixed: https://github.com/CoreumFoundation/coreum/blob/new-client/pkg/client2/client.go#L314
</blockquote></details>

Discussed on a call and resolved
___
*[`pkg/tx/tx.go` line 78 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBHgsvz6nkgUb9KEmio:-NBLpO9Q2mEf4WnJvmUG:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/362ce8d5fe429a970c0dcfce204004af3cdc863a/pkg/tx/tx.go#L78)):*

Take a look at original implementation: https://github.com/CoreumFoundation/coreum/blob/new-client/pkg/client2/client.go#L153
There are more edge cases to handle.
</blockquote></details>

Done.
___
*[`pkg/tx/tx.go` line 169 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBHfJY2BvrhpnzKxEJ8:-NBLq21GC9MK_3K5wGGE:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/362ce8d5fe429a970c0dcfce204004af3cdc863a/pkg/tx/tx.go#L169)):*

sdkerrors.New -> sdkerrors.ABCIError
</blockquote></details>

Done.
___
*[`pkg/tx/tx.go` line 199 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBHepBJ8s2RmbIY4azr:-NBLq393AxT_a2vc8fF_:b-nq8p67) ([raw file](https://github.com/coreumfoundation/coreum/blob/362ce8d5fe429a970c0dcfce204004af3cdc863a/pkg/tx/tx.go#L199)):*

this is the only relevant thing in this function, so whenever there is code like

```go
		errRes := checkTendermintError(err, txHash)
		if !isTxInMempool(errRes) {
			return txHash, errors.WithStack(err)
		}
```

it might be simplified to

```go
if !strings.Contains(strings.ToLower(err.Error()), strings.ToLower(mempool.ErrTxInCache.Error())) {
    return txHash, errors.WithStack(err)
}
```

Eventually `checkTendermintError` might be rewritten to `convertTendermintError` to provide sdk error (so they later might be compared if needed):

```go
func convertTendermintError(err error) error {
	if err == nil {
		return nil
	}
	errStr := strings.ToLower(err.Error())

	switch {
	case strings.Contains(errStr, strings.ToLower(mempool.ErrTxInCache.Error())):
                return sdkerrors.ErrTxInMempoolCache.Wrap(err.Error())
	case strings.Contains(errStr, "mempool is full"):
                return sdkerrors.ErrMempoolIsFull.Wrap(err.Error())
	case strings.Contains(errStr, "tx too large"):
                return sdkerrors.ErrTxTooLarge.Wrap(err.Error())
	default:
		return err
	}
}
```


and then, this code

```go
		errRes := checkTendermintError(err, txHash)
		if !isTxInMempool(errRes) {
			return txHash, errors.WithStack(err)
		}
```

could be changed to

```go
if !sdkerrors.ErrTxInMempoolCache.Is(err) {
    return txHash, errors.WithStack(err)
}
```

which is the preferred way IMO
</blockquote></details>

Discussed and done

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @wojtek-coreum, and @ysv)

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

___
*[`pkg/tx/tx.go` line 102 at r8](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBM1eIf5m2ZFQJZSvNB:-NBM1eIg7tOld3iXsZpj:b-xctnmx) ([raw file](https://github.com/coreumfoundation/coreum/blob/c41c26af10a9dc8205e849e2a1f2bdf5db934ef4/pkg/tx/tx.go#L102)):*
> ```Go
> 		}
> 	} else if res.Code != 0 {
> 		return nil, errors.Wrapf(sdkerrors.New(res.Codespace, res.Code, res.Log),
> ```

sdkerrors.New -> sdkerrors.ABCIError
___
*[`pkg/tx/tx.go` line 192 at r8](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBM20f3DaSBAjDKBJTf:-NBM20f3DaSBAjDKBJTg:b2otsv5) ([raw file](https://github.com/coreumfoundation/coreum/blob/c41c26af10a9dc8205e849e2a1f2bdf5db934ef4/pkg/tx/tx.go#L192)):*
> ```Go
> }
> 
> // the idea behind this function is to map it similar to cosmos sdk does it in the link below
> ```

similar -> similarly

- the part around "does" not very english

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman and @miladz68)

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`pkg/tx/tx.go` line 102 at r8](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBM1eIf5m2ZFQJZSvNB:-NBMZpHVE1ca5WlaObWD:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/c41c26af10a9dc8205e849e2a1f2bdf5db934ef4/pkg/tx/tx.go#L102)):*

sdkerrors.New -> sdkerrors.ABCIError
</blockquote></details>

Done.
___
*[`pkg/tx/tx.go` line 192 at r8](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBM20f3DaSBAjDKBJTf:-NBM_YQJ6lZHDd6C1bL0:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/c41c26af10a9dc8205e849e2a1f2bdf5db934ef4/pkg/tx/tx.go#L192)):*

similar -> similarly

- the part around "does" not very english
</blockquote></details>

Done.

ysv (dismissed):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

wojtek-coreum (changes requested):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: all files reviewed, 1 unresolved discussion (waiting on @begmaroman and @miladz68)

___
*[`pkg/tx/tx.go` line 102 at r8](https://reviewable.io/reviews/coreumfoundation/coreum/155#-NBM1eIf5m2ZFQJZSvNB:-NBMe8L87yL55hRsUYRr:b-ju3kiw) ([raw file](https://github.com/coreumfoundation/coreum/blob/c41c26af10a9dc8205e849e2a1f2bdf5db934ef4/pkg/tx/tx.go#L102)):*

Done.
</blockquote></details>

I don't see the change

wojtek-coreum (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @begmaroman)

ysv (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/155)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @begmaroman)

PR #162, opened 2022-09-07
Add new helpers to simplify Broadcast func usage in integration tests

ysv:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)


ysv (commented):
Note: this PR contains helper for better usage of tx.Factory & client.Conext and example of helper usage in a single integration test
If everyone is fine with the approach other tests will be refactored to use the same approach

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162#-:-NBNYSQ7DD0DyG3AEzG-:bg8d41n)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`pkg/types/types.go` line 40 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBNY7WZAAolb-SwgEyW:-NBNY7WZAAolb-SwgEyX:b-sk3jp0) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/pkg/types/types.go#L40)):*
> ```Go
> 
> // ClientCtx returns cosmos client context with values present for the Wallet.
> func (w Wallet) ClientCtx(parentCtx cosmosclient.Context) cosmosclient.Context {
> ```

the idea is to move Wallet to integration-test folder with some usage changes

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`integration-tests/bank/transfer.go` line 72 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBN_A5O5xFzWXBg8Nes:-NBN_A5P8r_Umznhbuxr:bky11yp) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/integration-tests/bank/transfer.go#L72)):*
> ```Go
> 		ctx,
> 		sender.ClientCtx(chain.ClientContext),
> 		chain.TxFactory().WithGas(chain.NetworkConfig.Fee.DeterministicGas.BankSend),
> ```

I also don't like this long chain of calls to get DeterministicGas
`chain.NetworkConfig.Fee.DeterministicGas.BankSend`

if smn has idea of how to improve it LMK pls



_Code quote:_
```Go
chain.NetworkConfig.Fee.DeterministicGas.BankSend
```
___
*[`pkg/types/types.go` line 40 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBNZqLr9Cmo9lkXpQKn:-NBNZqLr9Cmo9lkXpQKo:b-gd4otd) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/pkg/types/types.go#L40)):*
> ```Go
> 
> // ClientCtx returns cosmos client context with values present for the Wallet.
> func (w Wallet) ClientCtx(parentCtx cosmosclient.Context) cosmosclient.Context {
> ```

Another thought on this method location was Chain
```golang
func (c Chain) ClientCtxForWallet(wallet types.Wallet) {...}
```

but I prefer this version

wojtek-coreum (changes requested):
I like this approach

> Reviewed 7 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162#-:-NBQebzUDOFZQ6frSPF8:bous4il)* status: all files reviewed, 5 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 72 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBN_A5O5xFzWXBg8Nes:-NBQbWhQ2c4Mf0I_rUSw:bh30om8) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/integration-tests/bank/transfer.go#L72)):*

I also don't like this long chain of calls to get DeterministicGas
`chain.NetworkConfig.Fee.DeterministicGas.BankSend`

if smn has idea of how to improve it LMK pls
</blockquote></details>

I could imagine a function `chain.GasLimit(&banktypes.MsgSend{})` having this signture:

`func GasLimit(msg sdk.Msg)`

and returning appropriate gas limit based on reflection
___
*[`integration-tests/bank/transfer.go` line 66 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQc9J8BznehE-Kl5Y2:-NBQc9J8BznehE-Kl5Y3:b-p3z8g6) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/bank/transfer.go#L66)):*
> ```Go
> 		FromAddress: sender.Address().String(),
> 		ToAddress:   receiver.Address().String(),
> 		Amount: []sdk.Coin{
> ```

I had a plan to add function like:

`chain.Coin(amount sdk.Int) sdk.Coin` which creates a coin with appropriate denom to make all those calls shorter. Maybe you could add it in this PR as well?
___
*[`integration-tests/bank/transfer.go` line 67 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQcPo14LNTKaSwIYYo:-NBQcPo14LNTKaSwIYYp:b-ff9xqf) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/bank/transfer.go#L67)):*
> ```Go
> 		ToAddress:   receiver.Address().String(),
> 		Amount: []sdk.Coin{
> 			{Denom: chain.NetworkConfig.TokenSymbol, Amount: sdk.NewInt(10)},
> ```

I think `sdk.NewCoin` should be used instead of creating struct directly
___
*[`integration-tests/testing/chain.go` line 32 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQcwNJ6uLQzqxT7hpC:-NBQcwNJ6uLQzqxT7hpD:b-wa5mxx) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L32)):*
> ```Go
> // private key and stores mnemonic in Keyring.
> func (c Chain) RandomWallet() types.Wallet {
> 	name := uuid.New().String()
> ```

what about using address of the wallet instead of random string?
___
*[`integration-tests/testing/chain.go` line 54 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQdDPSEZhshEIlRJzQ:-NBQdDPSEZhshEIlRJzR:b-g5jjip) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L54)):*
> ```Go
> // TxFactory returns factory with present values for the Chain.
> func (c Chain) TxFactory() tx.Factory {
> 	return tx.Factory{}.
> ```

What about creating it once (with all the `With*` calls and then just return the same instance here?
___
*[`pkg/types/types.go` line 40 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBNZqLr9Cmo9lkXpQKn:-NBQe1LiB3BRCTvRPFxi:begmus4) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/pkg/types/types.go#L40)):*

Another thought on this method location was Chain
```golang
func (c Chain) ClientCtxForWallet(wallet types.Wallet) {...}
```

but I prefer this version
</blockquote></details>

I think this is not needed at all
___
*[`pkg/types/types.go` line 18 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQdZiY44_83sN0BHrl:-NBQdZiY44_83sN0BHrm:b-z7mpez) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/pkg/types/types.go#L18)):*
> ```Go
> 
> 	// Key is the private key of the wallet
> 	// TODO: Check if we really need to store private key here or public is enough since mnemonic is stored in keyring.
> ```

It is not needed. If you store keys by  address (instead of random string) in the keystore like I propose in another comment then this entire type is not needed at all. Address string might be used directly everywhere. Or (after looking at new `ClientCtx` method) I think that `chain.RandomWallet` could just return preconfigured `cosmosclient.Context` object (done now inside `ClientCtx`) to be passed later to `BroadcastTx` directly.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: all files reviewed, 8 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

___
*[`integration-tests/testing/chain.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBR1xJQ1j8V5x7jZERY:-NBR1xJQ1j8V5x7jZERZ:b-rm27y) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L26)):*
> ```Go
> 	Faucet        Faucet
> 
> 	Keyring keyring.UnsafeKeyring
> ```

Why do you use the "keyring.UnsafeKeyring" instead of the keyring.Keyring?
___
*[`integration-tests/testing/chain.go` line 31 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBR1d6c5SHWopj7iHHn:-NBR1d6c5SHWopj7iHHo:b-iuz199) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L31)):*
> ```Go
> // RandomWallet generates a wallet for the chain with random name and
> // private key and stores mnemonic in Keyring.
> func (c Chain) RandomWallet() types.Wallet {
> ```

I would remove that Wallet abstraction at all. If we work with the keyring just "name" of the key in the keyring is enough. Based on that name you can get any address type you need or key. So we can generate it, save to the keyring and return that name.
___
*[`pkg/types/types.go` line 42 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBR3bFj2YbVS1nHsMwc:-NBR3bFj2YbVS1nHsMwd:brhvcp0) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/pkg/types/types.go#L42)):*
> ```Go
> func (w Wallet) ClientCtx(parentCtx cosmosclient.Context) cosmosclient.Context {
> 	return parentCtx.
> 		WithFromName(w.Name).
> ```

I'm still now sure that we need both the "FromName" and "FromAddress". Why you send the tx it should take the address from the keyring, using the "name", check it of possible. In that case we won't need the " ClientCtx" method.

miladz68 (changes requested):
> Reviewed 6 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: all files reviewed, 11 unresolved discussions (waiting on @begmaroman, @wojtek-coreum, and @ysv)

___
*[`integration-tests/testing/chain.go` line 24 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBRZguj2Z6xdJGsHkck:-NBRZguj2Z6xdJGsHkcl:bphpmjw) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L24)):*
> ```Go
> 
> 	NetworkConfig app.NetworkConfig
> 	Faucet        Faucet
> ```

I commented on it before. Maybe rename this faucet to OldFaucet and introduce your new faucet signature here that does not need a Wallet type, but accepts only address, so you can complete remove types.Wallet from new code.
___
*[`integration-tests/testing/chain.go` line 31 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBRWxRq9kM79eAMXtnc:-NBRWxRq9kM79eAMXtnd:b-cigwzb) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L31)):*
> ```Go
> // RandomWallet generates a wallet for the chain with random name and
> // private key and stores mnemonic in Keyring.
> func (c Chain) RandomWallet() types.Wallet {
> ```

we can just return sdk.Address and a modified clientCtx, and that should be enough.
I  understand that you need it for backward compatiblity with faucet signature. But at some point you should break this and introduce new faucet. So maybe you should introduce new faucet in this PR along side the old faucet, so your new functions can have the correct signature. 
___
*[`integration-tests/testing/chain.go` line 32 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQcwNJ6uLQzqxT7hpC:-NBRWISn7ntkghaye1OJ:b40kk0) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L32)):*

what about using address of the wallet instead of random string?
</blockquote></details>

this is a good idea, then we will not need to import uuid package and we will not need wallet type.
___
*[`pkg/types/types.go` line 40 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBNZqLr9Cmo9lkXpQKn:-NBRZLli4AhyXDzlDucQ:b-8j81e0) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/pkg/types/types.go#L40)):*

I think this is not needed at all
</blockquote></details>

I think it is a good idea and tests might need to have access to the clientCtx without generating a wallet.
___
*[`pkg/types/types.go` line 40 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBRXuCqDLvC_QCg9_PV:-NBRXuCqDLvC_QCg9_PW:b-buldtx) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/pkg/types/types.go#L40)):*
> ```Go
> 
> // ClientCtx returns cosmos client context with values present for the Wallet.
> func (w Wallet) ClientCtx(parentCtx cosmosclient.Context) cosmosclient.Context {
> ```

This should be provided on chain type so the tests will have access to it. Also RandomWallet type on chain should return a modified version of this

dzmitryhil (commented):
> Reviewed 7 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: all files reviewed, 11 unresolved discussions (waiting on @begmaroman, @wojtek-coreum, and @ysv)

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: 2 of 8 files reviewed, 11 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`integration-tests/bank/transfer.go` line 72 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBN_A5O5xFzWXBg8Nes:-NBSJ2N5CgjjUDE9fOwu:bqmjwcl) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/integration-tests/bank/transfer.go#L72)):*

I could imagine a function `chain.GasLimit(&banktypes.MsgSend{})` having this signture:

`func GasLimit(msg sdk.Msg)`

and returning appropriate gas limit based on reflection
</blockquote></details>

Added 
___
*[`integration-tests/bank/transfer.go` line 66 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQc9J8BznehE-Kl5Y2:-NBSGm0g23nSkmYpGbUv:b-dkerwy) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/bank/transfer.go#L66)):*

I had a plan to add function like:

`chain.Coin(amount sdk.Int) sdk.Coin` which creates a coin with appropriate denom to make all those calls shorter. Maybe you could add it in this PR as well?
</blockquote></details>

I like it
Added
___
*[`integration-tests/bank/transfer.go` line 67 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQcPo14LNTKaSwIYYo:-NBSGs6LBbAMfrBgEbGt:blivggg) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/bank/transfer.go#L67)):*

I think `sdk.NewCoin` should be used instead of creating struct directly
</blockquote></details>

changed to chain.NewCoin
___
*[`integration-tests/testing/chain.go` line 24 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBRZguj2Z6xdJGsHkck:-NBSMbev5FVQsuei3Gz1:b-d35st8) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L24)):*

I commented on it before. Maybe rename this faucet to OldFaucet and introduce your new faucet signature here that does not need a Wallet type, but accepts only address, so you can complete remove types.Wallet from new code.
</blockquote></details>

It will definitely be done but in the next PR
I prefer to make a few smaller PRs to get them merged one by one faster instead of one big

this one is about adding helpers for broadcast & then I will try to deal with types
___
*[`integration-tests/testing/chain.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBR1xJQ1j8V5x7jZERY:-NBSN9cv6FKmqFQZlkn7:b-c5d46z) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L26)):*

Why do you use the "keyring.UnsafeKeyring" instead of the keyring.Keyring?
</blockquote></details>

because of this line `	privKeyHex, err := c.Keyring.UnsafeExportPrivKeyHex(name)` in RandomWallet
It is mostly to keep compatibility.
I will add TODO for now and will be done in the one of next PRs
___
*[`integration-tests/testing/chain.go` line 31 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBR1d6c5SHWopj7iHHn:-NBSNmyJ5K9SpCaBUVqY:b-x8ay60) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L31)):*

I would remove that Wallet abstraction at all. If we work with the keyring just "name" of the key in the keyring is enough. Based on that name you can get any address type you need or key. So we can generate it, save to the keyring and return that name.
</blockquote></details>

will be done in the next PRs when removing types.Wallet

I prefer to make a few smaller PRs to get them merged one by one faster instead of one huge
___
*[`integration-tests/testing/chain.go` line 31 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBRWxRq9kM79eAMXtnc:-NBSO5pG12zPwT6ccxOx:b-fw7fp5) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L31)):*

we can just return sdk.Address and a modified clientCtx, and that should be enough.
I  understand that you need it for backward compatiblity with faucet signature. But at some point you should break this and introduce new faucet. So maybe you should introduce new faucet in this PR along side the old faucet, so your new functions can have the correct signature. 
</blockquote></details>

will be done in the next PRs when removing types.Wallet
___
*[`integration-tests/testing/chain.go` line 32 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQcwNJ6uLQzqxT7hpC:-NBTbtGcEorfZ-by01zQ:b-6zin5w) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L32)):*

this is a good idea, then we will not need to import uuid package and we will not need wallet type.
</blockquote></details>

yeah, but we don't know the address before we generate mnemonic 
Anyways I see benefits from this approach so I changed to it in the least painful way I found
___
*[`integration-tests/testing/chain.go` line 54 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQdDPSEZhshEIlRJzQ:-NBTnL7E2SHgh86HvB_2:bapbnc7) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L54)):*

What about creating it once (with all the `With*` calls and then just return the same instance here?
</blockquote></details>

because we have `WithKeybase(c.Keyring)` where keyring passed by value and we add new keys to it all of time so I doubt if new keys will be added to copy of keyring value. Maybe I'm wrong
I didn't actually check this but I feel that it should be method not attribute because it is constructed from other attributes only
___
*[`pkg/types/types.go` line 40 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBNZqLr9Cmo9lkXpQKn:-NBToUm07FVFrxeZ24xj:b-ydjkop) ([raw file](https://github.com/coreumfoundation/coreum/blob/2b20fabbf6d182f472203299a2d175ca5acfdcf2/pkg/types/types.go#L40)):*

I think it is a good idea and tests might need to have access to the clientCtx without generating a wallet.
</blockquote></details>

I agree this method is unnecessary at least for now
removed it
___
*[`pkg/types/types.go` line 18 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBQdZiY44_83sN0BHrl:-NBToL4Q8KzPIBUQAYqU:b70veqr) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/pkg/types/types.go#L18)):*

It is not needed. If you store keys by  address (instead of random string) in the keystore like I propose in another comment then this entire type is not needed at all. Address string might be used directly everywhere. Or (after looking at new `ClientCtx` method) I think that `chain.RandomWallet` could just return preconfigured `cosmosclient.Context` object (done now inside `ClientCtx`) to be passed later to `BroadcastTx` directly.
</blockquote></details>

changed todo to drop wallet type
storing by address is done
___
*[`pkg/types/types.go` line 40 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBRXuCqDLvC_QCg9_PV:-NBTodaNA6a_dUN8EQK_:b64guz0) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/pkg/types/types.go#L40)):*

This should be provided on chain type so the tests will have access to it. Also RandomWallet type on chain should return a modified version of this
</blockquote></details>

on chain type we have `ClientContext` as attribute I think it should be enough
as for method for now it is not necessary so remvoed it
___
*[`pkg/types/types.go` line 42 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBR3bFj2YbVS1nHsMwc:-NBTpSp6-MwvnfETBFi0:b600y5d) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/pkg/types/types.go#L42)):*

I'm still now sure that we need both the "FromName" and "FromAddress". Why you send the tx it should take the address from the keyring, using the "name", check it of possible. In that case we won't need the " ClientCtx" method. 
</blockquote></details>

I think both are required
if I remove address: `InvalidArgument desc = Address cannot be empty: invalid request`
if I remove name: `.info: key not found`

I think that name is used as sort of key so keyring is able to sign
and address is used to set address in tx and if it is empty it literally sets empty string there

ysv (commented):
sorry I forgot that it needs to be merged instead of rebasing

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162#-:-NBTsSk1-8V1w9VEywKo:b-i5jx4j)* status: 2 of 8 files reviewed, 11 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`integration-tests/testing/chain.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBR1xJQ1j8V5x7jZERY:-NBTtjQY-CT0JvFFNlEB:b-tf52r7) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2fae1cc9e41cca2d4ddfbee68c6ec1bd3236225/integration-tests/testing/chain.go#L26)):*

because of this line `	privKeyHex, err := c.Keyring.UnsafeExportPrivKeyHex(name)` in RandomWallet
It is mostly to keep compatibility.
I will add TODO for now and will be done in the one of next PRs
</blockquote></details>

Done after doing other changes

wojtek-coreum (changes requested):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: all files reviewed, 9 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/bank/transfer.go` line 73 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBVvbvA8nbbBWZ0RYH4:-NBVvbvA8nbbBWZ0RYH5:bgjni9q) ([raw file](https://github.com/coreumfoundation/coreum/blob/bd3f454a6e47470fadb9cc35894bd1fe6c36e2de/integration-tests/bank/transfer.go#L73)):*
> ```Go
> 	result, err := tx.BroadcastTx(
> 		ctx,
> 		chain.ClientContext.WithFromName(sender.Name).WithFromAddress(sdk.AccAddress(sender.Name)),
> ```

I would still prefer to write sth like this instead:

```go
func (ch Chain) WithRandomWallet(initialBalance []sdk.Coin) client.ClientContext {
    // ... do the key generation ...
    
    if len(initialBalance > 0) {
        // ... call faucet to fund account ...
    }
    return chain.ClientContext.WithFromName(keyInfo.GetAddress().String()).WithFromAddress(keyInfo.GetAddress())
}
```

- it can be used directly like here
- if address of the wallet needs to be retrieved directly, it might be taken from returned client context
___
*[`integration-tests/testing/chain.go` line 32 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBVudsz6Zhc268BSSGl:-NBVudsz6Zhc268BSSGm:bcaqlrt) ([raw file](https://github.com/coreumfoundation/coreum/blob/bd3f454a6e47470fadb9cc35894bd1fe6c36e2de/integration-tests/testing/chain.go#L32)):*
> ```Go
> // RandomWallet generates a wallet for the chain with random name and
> // private key and stores mnemonic in Keyring.
> func (c Chain) RandomWallet() types.Wallet {
> ```

I don't see a need for returning `types.Wallet` here. `sdk.AccAddress` seems to be a better option
___
*[`integration-tests/testing/chain.go` line 64 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBVuFixCcFM0c90Lj-R:-NBVuFixCcFM0c90Lj-S:b86eba4) ([raw file](https://github.com/coreumfoundation/coreum/blob/bd3f454a6e47470fadb9cc35894bd1fe6c36e2de/integration-tests/testing/chain.go#L64)):*
> ```Go
> }
> 
> func (c Chain) GasLimitByMsgs(msgs ...sdk.Msg) uint64 {
> ```

This logic has  been already implemented:
https://github.com/CoreumFoundation/coreum/blob/master/x/auth/ante/gas.go#L50
https://github.com/CoreumFoundation/coreum/blob/master/x/auth/ante/gas.go#L39

Could you modify the existing flow to export the appropriate function and use it in both places?
Otherwise we will forget to keep them in sync.

miladz68 (approved):
> Reviewed 1 of 8 files at r1, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: all files reviewed, 6 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @ysv)

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @wojtek-coreum)

___
*[`integration-tests/bank/transfer.go` line 73 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBVvbvA8nbbBWZ0RYH4:-NBWkkuJ8tBW6OrI4RmX:b-qt7pbc) ([raw file](https://github.com/coreumfoundation/coreum/blob/bd3f454a6e47470fadb9cc35894bd1fe6c36e2de/integration-tests/bank/transfer.go#L73)):*

I would still prefer to write sth like this instead:

```go
func (ch Chain) WithRandomWallet(initialBalance []sdk.Coin) client.ClientContext {
    // ... do the key generation ...
    
    if len(initialBalance > 0) {
        // ... call faucet to fund account ...
    }
    return chain.ClientContext.WithFromName(keyInfo.GetAddress().String()).WithFromAddress(keyInfo.GetAddress())
}
```

- it can be used directly like here
- if address of the wallet needs to be retrieved directly, it might be taken from returned client context
</blockquote></details>

I think that sdk.Address is better representation of account than ClientContext and I prefer to operate with it. sdk.Address is minimalistic and could be easily replaced while ClientContext is very specific and might be complicated to replace in the future

I also don't like idea of creating and fundig wallet in a single func.
___
*[`integration-tests/testing/chain.go` line 32 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBVudsz6Zhc268BSSGl:-NBWUlPN5j0TswrzSWZO:bb7f2mf) ([raw file](https://github.com/coreumfoundation/coreum/blob/bd3f454a6e47470fadb9cc35894bd1fe6c36e2de/integration-tests/testing/chain.go#L32)):*

I don't see a need for returning `types.Wallet` here. `sdk.AccAddress` seems to be a better option
</blockquote></details>

done
___
*[`integration-tests/testing/chain.go` line 64 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBVuFixCcFM0c90Lj-R:-NBWaHI3Bvu3MvaVog-I:b8is0yi) ([raw file](https://github.com/coreumfoundation/coreum/blob/bd3f454a6e47470fadb9cc35894bd1fe6c36e2de/integration-tests/testing/chain.go#L64)):*

This logic has  been already implemented:
https://github.com/CoreumFoundation/coreum/blob/master/x/auth/ante/gas.go#L50
https://github.com/CoreumFoundation/coreum/blob/master/x/auth/ante/gas.go#L39

Could you modify the existing flow to export the appropriate function and use it in both places?
Otherwise we will forget to keep them in sync.
</blockquote></details>

makes sense 
done

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: 6 of 9 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`integration-tests/bank/transfer.go` line 73 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/162#-NBVvbvA8nbbBWZ0RYH4:-NBWmManDWw-817D2XbM:b-4yt19c) ([raw file](https://github.com/coreumfoundation/coreum/blob/bd3f454a6e47470fadb9cc35894bd1fe6c36e2de/integration-tests/bank/transfer.go#L73)):*

I think that sdk.Address is better representation of account than ClientContext and I prefer to operate with it. sdk.Address is minimalistic and could be easily replaced while ClientContext is very specific and might be complicated to replace in the future

I also don't like idea of creating and fundig wallet in a single func.
</blockquote></details>

Because we had Wallet abstraction and now it is complicated to remove all references to it. While I prefer to have smth as minimal as possible

Anyways on the next PR I will also consider this as option

wojtek-coreum (approved):
very good PR, thx!

> Reviewed 1 of 3 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162#-:-NBX8GY44NkKSzrrx_bV:b7au1dd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @begmaroman, @dzmitryhil, and @miladz68)

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/162)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @begmaroman and @miladz68)

PR #170, opened 2022-09-12
Improve CI by caching go modules and linter

begmaroman:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)


wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBpNQS24VcA5CgutEng:-NBpNQS24VcA5CgutEnh:bk1snno) (no related file):*
Please add reviewers to the PR so they are notified

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBpQ3KWBE2Bpp-6VVfL:-NBpQ3KWBE2Bpp-6VVfM:b1eh6ea) (no related file):*
> crust lint > 6 mins vs native CLI which is ~3 min. 2x faster.

Did you compare it to current `crust` implementation or the one linting only current repo. Because now `crust` lints 3 repos: crust, coreum and faucet. So if it takes 6 mins to lint 3 repos versus 2 mins to lint single repo. Then it is not really faster.

Later I will prepare crust version which lints only one repo, so we can compare real results.

___
___
*[`.golangci.yaml` line 1 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBpNXMg3Uvw1qX9y5tS:-NBpNXMg3Uvw1qX9y5tT:b83yrw) ([raw file](https://github.com/coreumfoundation/coreum/blob/22b71436c349ee9e0a6990c892e76a5099020579/.golangci.yaml#L1)):*
> ```YAML
> run:
> ```

Let's not add separate CI config to each repo. It will be a nightmare to manage it and keep in consistent state everywhere. Otherwise we will have different coding standards in every repo which will be an even bigger nightmare. To answer the argument that it's a rare operation:
1. it's not, I updated config a minute ago: https://reviewable.io/reviews/CoreumFoundation/crust/85#- Usually it needs to be done after releasing every new version of golangci. So now you may wait until new change is in master and copy it here again. (I'm not really asking you to do it, I just try to convince you that you don't want to do it manually all the time)
2. even if it's rare I still prfer to make our life easier, not harder

Otherwise people will tend to ignore the need of changes rather than going through all repos to update rules everywhere. We all are lazy. And being lazy is the most important skill of software developers, because it leads us to do things once and automate everything. Please be lazy.
___
*[`.github/workflows/ci.yml` line 24 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBpOut_E6JeTaHaXbeH:-NBpOuta4B83BjIIBPGR:b-cayqma) ([raw file](https://github.com/coreumfoundation/coreum/blob/22b71436c349ee9e0a6990c892e76a5099020579/.github/workflows/ci.yml#L24)):*
> ```YAML
>           echo "$(pwd)/crust/bin" >> $GITHUB_PATH
>       - name: Lint
>         run: crust lint --log-format=yaml
> ```

There is a reason why building framework does all those operations together (in one line). Every operation there might have dependencies (look for `deps build.DepsFunc` in source code). If multiple operations require the same dependency (e.g. built binary) it is executed only once, on first request). By splitting operations between different commands you stop taking those benefits (deps are now executed many times) possibly leading to longer execution. Maybe the slower time is not visible here, because there are not so many deps, but it will be visible once we start building docker images in CI (in your solution binaries will be created twice, once in `build` command, second time in new `images` command. Take a look at this PR: https://reviewable.io/reviews/CoreumFoundation/crust/84#-)

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 5 unresolved discussions (waiting on @begmaroman)

___
*[`.github/workflows/golangci-lint.yml` line 1 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBpVqOcCNIUecxtSMCE:-NBpVqOcCNIUecxtSMCF:b-3ssit5) ([raw file](https://github.com/coreumfoundation/coreum/blob/22b71436c349ee9e0a6990c892e76a5099020579/.github/workflows/golangci-lint.yml#L1)):*
> ```YAML
> name: golangci-lint
> ```

Pasting here arguments I published on Slack on why crust lint was introduced:
- to store linter config in common place
- to do more things than only running golangci-lint (checking that files contain new lines, go mod tidy does not introduce any changes etc.)
- to run golangci-lint correctly - it's tricky if there are many repos, containing many modules
- to be able to reproduce the same CI behavior on local machine
- to be able to move quickly to another CI environment (github actions is a github-specific thing)
- to have one liner which may be used to lint everything before creating a PR

wojtek-coreum (changes requested):
> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

___
*[`.github/workflows/ci.yml` line 17 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqxMsu3352PthVgrL3:-NBqxMsvAGQcYaMmiCCT:b-20a3ui) ([raw file](https://github.com/coreumfoundation/coreum/blob/ae2f691990940a261a5177182a94124c211511b1/.github/workflows/ci.yml#L17)):*
> ```YAML
>       - name: Setup go modules cache
>         uses: actions/cache@v3
>         with:
> ```

Instead of creating many caches, create a common directory like `.cache/ci`, set up caching for this directories and point all the env variables (GOPATH, GOCACHE, GOLANGCI_LINT_CACHE) to subdirectories of `.cache/crust/go`.
___
*[`.github/workflows/ci.yml` line 30 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqx52S4gHdyaK19pql:-NBqx52S4gHdyaK19pqm:b5dhvpd) ([raw file](https://github.com/coreumfoundation/coreum/blob/ae2f691990940a261a5177182a94124c211511b1/.github/workflows/ci.yml#L30)):*
> ```YAML
>           path: ~/.cache/golangci-lint
>           key: ${{ runner.os }}-ci-cache
>       - name: Checkout coreum
> ```

You don't need to checkout other repositories. Crust does everything what is needed

begmaroman (commented):
The PR is still in WIP 

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170#-:-NBvM3SS3C6OuQoX9v5Y:b9kahtl)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

___
*[`.github/workflows/ci.yml` line 24 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBpOut_E6JeTaHaXbeH:-NBvcWGT2AjGkfV2F-7s:bza2tht) ([raw file](https://github.com/coreumfoundation/coreum/blob/22b71436c349ee9e0a6990c892e76a5099020579/.github/workflows/ci.yml#L24)):*

There is a reason why building framework does all those operations together (in one line). Every operation there might have dependencies (look for `deps build.DepsFunc` in source code). If multiple operations require the same dependency (e.g. built binary) it is executed only once, on first request). By splitting operations between different commands you stop taking those benefits (deps are now executed many times) possibly leading to longer execution. Maybe the slower time is not visible here, because there are not so many deps, but it will be visible once we start building docker images in CI (in your solution binaries will be created twice, once in `build` command, second time in new `images` command. Take a look at this PR: https://reviewable.io/reviews/CoreumFoundation/crust/84#-)
</blockquote></details>

It builds and downloads deps if they don't exist. The first step downloads and builds everything, the next steps will reuse them. 
___
*[`.github/workflows/ci.yml` line 17 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqxMsu3352PthVgrL3:-NBvbZ5KETw3FfWsdOqq:b-zevy45) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L17)):*

Instead of creating many caches, create a common directory like `.cache/ci`, set up caching for this directories and point all the env variables (GOPATH, GOCACHE, GOLANGCI_LINT_CACHE) to subdirectories of `.cache/crust/go`.
</blockquote></details>

Go mod cache should be separated coz it depends on the go.sum hash. If we add or remove some modules, the cache should be updated. 
___
*[`.github/workflows/ci.yml` line 30 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqx52S4gHdyaK19pql:-NBvcySY1WSJzYY3ECzI:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L30)):*

You don't need to checkout other repositories. Crust does everything what is needed
</blockquote></details>

ok

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

___
*[`.github/workflows/ci.yml` line 24 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBpOut_E6JeTaHaXbeH:-NBvgzR0AQeT6-k_6LC7:befdi0r) ([raw file](https://github.com/coreumfoundation/coreum/blob/22b71436c349ee9e0a6990c892e76a5099020579/.github/workflows/ci.yml#L24)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

It builds and downloads deps if they don't exist. The first step downloads and builds everything, the next steps will reuse them. 
</blockquote></details>

I'm saying that there is a difference between running this:

```
$ crust lint
$ crust build
```

and this:

```
$ crust lint build
```

The second run may be faster depending on how many operations internally in the crust are shared between to commands.

You may take this PR as an example: https://reviewable.io/reviews/CoreumFoundation/crust/84#-

and compare:

```
crust build
crust images
```

to

```
crust build images
```

In second case binaries are built only once
___
*[`.github/workflows/ci.yml` line 17 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqxMsu3352PthVgrL3:-NBvgfft33lPPd3E3FB6:bb7r2f0) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L17)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

Go mod cache should be separated coz it depends on the go.sum hash. If we add or remove some modules, the cache should be updated. 
</blockquote></details>

why?

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

___
*[`.github/workflows/ci.yml` line 17 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqxMsu3352PthVgrL3:-NBvinQ-2ALXnu26r08g:b-h1gaxw) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L17)):*

why?
</blockquote></details>

Because we need to cache only the current dependencies and fetch only deps that we need based on go.sum file

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @begmaroman and @wojtek-coreum)

___
*[`.github/workflows/ci.yml` line 17 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqxMsu3352PthVgrL3:-NBvzavwDJGLZ-WYwliZ:b-s710db) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L17)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

Because we need to cache only the current dependencies and fetch only deps that we need based on go.sum file 
</blockquote></details>

But why do we care? On our laptop we don't clean the cache every time go.sum is updated. Is it about limited space? If yes, then I would prefer to clean the cache once per month maybe (by adding month and year to the key).

ysv (changes requested):
I like the improvement in terms of timing we already have :thumbsup: 
Just a few minor comments and should be ready to go

> Reviewed 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170#-:-NBx1ydcBN8QWXh6DmR-:b-e4csre)* status: 2 of 3 files reviewed, 7 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`.github/workflows/ci.yml` line 17 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqxMsu3352PthVgrL3:-NBwxciPDiKpE1Gu8Z36:b1la4wl) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L17)):*

But why do we care? On our laptop we don't clean the cache every time go.sum is updated. Is it about limited space? If yes, then I would prefer to clean the cache once per month maybe (by adding month and year to the key).
</blockquote></details>

I think Wojtek's point to store as much as possible in a single place makes sense except go.mod.

As far as I understand it uses `go.sum` file content hash for cache key and if it doesn't match cache will not be mounted.

Checked my local mod cache size and it is 7.5GB + 2.5GB of go build cache. I think it is too much and I prefer to have this key and reset it after go.sum changes. 

So Roman's version is ok for me. I think we can store all caches in a single dir but go modules is special case and needs to be stored separately
___
*[`.github/workflows/ci.yml` line 21 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBwxKqq43wMPFcJL6Le:-NBwxKqr3ouGGfhH3qKZ:bvrvnnm) ([raw file](https://github.com/coreumfoundation/coreum/blob/a73cfe9a64cd1ba4ad412b9cf26437ebfd75ba55/.github/workflows/ci.yml#L21)):*
> ```YAML
>             ~/.cache/go-build
>             ~/go/pkg/mod
>           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
> ```

does this line consider fact that we have go.sum in both crust & coreum ?

wojtek-coreum (commented):
> Reviewed 1 of 3 files at r2, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed (commit messages unreviewed), 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`.github/workflows/ci.yml` line 17 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqxMsu3352PthVgrL3:-NC-QxMx4H1XsY-G1W_A:bl0xfe8) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L17)):*

I think Wojtek's point to store as much as possible in a single place makes sense except go.mod.

As far as I understand it uses `go.sum` file content hash for cache key and if it doesn't match cache will not be mounted.

Checked my local mod cache size and it is 7.5GB + 2.5GB of go build cache. I think it is too much and I prefer to have this key and reset it after go.sum changes. 

So Roman's version is ok for me. I think we can store all caches in a single dir but go modules is special case and needs to be stored separately
</blockquote></details>

Ok, cool.
___
*[`.github/workflows/ci.yml` line 21 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBwxKqq43wMPFcJL6Le:-NC-RDlW-rZAYdOO6PkY:b-m7hlqt) ([raw file](https://github.com/coreumfoundation/coreum/blob/a73cfe9a64cd1ba4ad412b9cf26437ebfd75ba55/.github/workflows/ci.yml#L21)):*

does this line consider fact that we have go.sum in both crust & coreum ?
</blockquote></details>

I think it uses all go.sum from the coreum repo only. It might be a little complicated to hash go sum from crust as well, but seems like the coreum contains all needed deps.

miladz68 (approved):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @ysv)

___
*[`.github/workflows/ci.yml` line 30 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqx52S4gHdyaK19pql:-NC03PIn7cc8rHF4ccLT:b-vvj801) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L30)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

ok
</blockquote></details>

but it is still done
___
*[`.github/workflows/ci.yml` line 21 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBwxKqq43wMPFcJL6Le:-NC02EZm2VYNBE7YKQ45:b59mvpo) ([raw file](https://github.com/coreumfoundation/coreum/blob/a73cfe9a64cd1ba4ad412b9cf26437ebfd75ba55/.github/workflows/ci.yml#L21)):*
<details><summary><i>Previously, begmaroman (Roman Behma) wrote…</i></summary><blockquote>

I think it uses all go.sum from the coreum repo only. It might be a little complicated to hash go sum from crust as well, but seems like the coreum contains all needed deps. 
</blockquote></details>

changing `**/go.sum` to `../**/go.sum` might help
___
*[`.github/workflows/ci.yml` line 21 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NC02kT2BOo_FRvUrwh3:-NC02kT3CMgoNO844hIZ:binzt0q) ([raw file](https://github.com/coreumfoundation/coreum/blob/64a10a3fc03a283fd6bfc9b9c05f2cc5d435a959/.github/workflows/ci.yml#L21)):*
> ```YAML
>         run: |
>           cd ..
>           git clone https://github.com/CoreumFoundation/crust.git 
> ```

Could we use `git clone` instead of github action? It's a simple command which is portable and works everywhere, not only in github

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @ysv)

___
*[`.github/workflows/ci.yml` line 30 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBqx52S4gHdyaK19pql:-NC03Tq-55OAOI2oLgzl:b3wqng) ([raw file](https://github.com/coreumfoundation/coreum/blob/b7c286bfe733f60d7be698ade6aea0029f9c4949/.github/workflows/ci.yml#L30)):*

but it is still done
</blockquote></details>

ah, sorry, ignore my last comment

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`.github/workflows/ci.yml` line 21 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NC02kT2BOo_FRvUrwh3:-NC06Ydc4tl70hZ8HhDd:bibwtpe) ([raw file](https://github.com/coreumfoundation/coreum/blob/64a10a3fc03a283fd6bfc9b9c05f2cc5d435a959/.github/workflows/ci.yml#L21)):*

Could we use `git clone` instead of github action? It's a simple command which is portable and works everywhere, not only in github
</blockquote></details>

This is the github actions file and there is an existing common action to clone projects. I don't see any reason why we can't use the native action to clone repo.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, and @ysv)

___
*[`.github/workflows/ci.yml` line 21 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBwxKqq43wMPFcJL6Le:-NC07t-c5rMKF2ZsDKuM:ban4ppx) ([raw file](https://github.com/coreumfoundation/coreum/blob/a73cfe9a64cd1ba4ad412b9cf26437ebfd75ba55/.github/workflows/ci.yml#L21)):*

changing `**/go.sum` to `../**/go.sum` might help
</blockquote></details>

sorry, I see now that all repos are cloned to current dir, not the parent one so this change is not needed

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`.github/workflows/ci.yml` line 21 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NBwxKqq43wMPFcJL6Le:-NC07Eco9vM7KFnyfjzc:b5dgn42) ([raw file](https://github.com/coreumfoundation/coreum/blob/a73cfe9a64cd1ba4ad412b9cf26437ebfd75ba55/.github/workflows/ci.yml#L21)):*

sorry, I see now that all repos are cloned to current dir, not the parent one so this change is not needed
</blockquote></details>

It won't work

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

dzmitryhil (commented):
> Reviewed 2 of 3 files at r2, 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/integration_test.go` line 89 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NC3v9WY-5OvLMnqa1VW:-NC3v9WY-5OvLMnqa1VX:b73j7lu) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3ca86b5d4a60a52924652f6eeb75e74947172f2/integration-tests/integration_test.go#L89)):*
> ```Go
> 
> func newChain(ctx context.Context, cfg config) (coreumtesting.Chain, error) {
> 	//nolint:contextcheck
> ```

please change both cases to `//nolint:contextcheck // Comment saying why it complains`

dzmitryhil (dismissed):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, and @ysv)

begmaroman (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/integration_test.go` line 89 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/170#-NC3v9WY-5OvLMnqa1VW:-NC550qb93zIO0nIkgDE:bb6un1j) ([raw file](https://github.com/coreumfoundation/coreum/blob/f3ca86b5d4a60a52924652f6eeb75e74947172f2/integration-tests/integration_test.go#L89)):*

please change both cases to `//nolint:contextcheck // Comment saying why it complains`
</blockquote></details>

Done

wojtek-coreum (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @miladz68, and @ysv)

wojtek-coreum (approved):
> Reviewed 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @miladz68, and @ysv)

dzmitryhil (approved):

ysv (approved):
> Reviewed 1 of 3 files at r9, 2 of 2 files at r10, 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68)

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/170)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68)

PR #179, opened 2022-09-14
PoC of new deterministic gas

wojtek-coreum:
Please don't waste your time on leaving comments in this PR for now. Just briefly understand the code and then we will discuss because there are some concerns


wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`app/gas.go` line 80 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NBxnV912IjNlVEXGwnl:-NBxnV924thuwxkqrOoi:b9fbqll) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L80)):*
> ```Go
> 
> func (s *deterministicMsgServer) RegisterService(sd *googlegrpc.ServiceDesc, handler interface{}) {
> 	// this is magic
> ```

Inspecting `RegisterService` of wrapped implementation might help to understand this piece:
https://github.com/cosmos/cosmos-sdk/blob/ff416ee63d32da5d520a8b2d16b00da762416146/baseapp/msg_service_router.go#L109

dzmitryhil (commented):
I like that concept. It seems more elegant, and precise than previous.

> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179#-:-NC0K8vCAgqvscYGbHDn:b-i4qw3p)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`app/gas.go` line 50 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NC0JhfC-4o4lv3KMYqP:-NC0JhfC-4o4lv3KMYqQ:b-gho15h) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L50)):*
> ```Go
> 
> func (r *deterministicGasRouter) AddRoute(route sdk.Route) sdk.Router {
> 	fmt.Println(route.Path())
> ```

Don't forget to remove
___
*[`app/gas.go` line 101 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NC0JqXu04ADXGxoJJJD:-NC0JqXu04ADXGxoJJJE:b-nvgyza) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L101)):*
> ```Go
> 	if gasRequired > 0 {
> 		ctx.GasMeter().ConsumeGas(gasRequired, "DeterministicGas")
> 		ctx = ctx.WithGasMeter(sdk.NewGasMeter(5 * gasRequired))
> ```

What is the purpose of the 5*, to have more gas in case we made a mistake in our computation?

dzmitryhil (commented):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`app/gas.go` line 101 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NC0JqXu04ADXGxoJJJD:-NC0LjV03cci4cN6_Qdp:bbf3ojk) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L101)):*

What is the purpose of the 5*, to have more gas in case we made a mistake in our computation?
</blockquote></details>

We don't want to charge for the worst case, rather an average one. So some transfers might take more gas, so we need a margin

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`app/gas.go` line 50 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NC0JhfC-4o4lv3KMYqP:-NC0MdH68ZcIJaWrIbKV:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L50)):*

Don't forget to remove
</blockquote></details>

Done.

dzmitryhil (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 1 unresolved discussion (waiting on @begmaroman, @miladz68, and @ysv)

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 1 of 21 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZfjWJCB8BpSFY7SJ3:-NCZfjWJCB8BpSFY7SJ4:b-w8v26e) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L11)):*
> ```Go
> func DefaultDeterministicGasRequirements() DeterministicGasRequirements {
> 	return DeterministicGasRequirements{
> 		FixedGas: 50000,
> ```

I don't like that those values are here. In my understanding the values should come from the either "network" where we keep chain specific values of from the genesis. 
And why the entire file is in the "ante" package? You can keep in the the gas.go file and update to the interface, which will be implemented by the Network type/subtype to provide the values. 
___
*[`x/auth/types/deterministic.go` line 18 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZiMHO3oANh5hnT8Vv:-NCZiMHO3oANh5hnT8Vw:by261gw) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L18)):*
> ```Go
> const (
> 	// FreeBytes defines how many tx bytes are stored for free (included in `FixedGas` price)
> 	FreeBytes = 2048
> ```

Why the FixedGas is param but the FreeBytes and FreeSignatures  are constant? Let's make them all parameters.
___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCZef6u7YsHprk02Q0S:b-ovhcr9) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*
> ```Go
> 	// - `FreeBytes` bytes of transaction
> 	FixedGas uint64
> 	BankSend uint64
> ```

Is is possible to keep the mappin in the map?

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 1 of 21 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZfjWJCB8BpSFY7SJ3:-NCZkQhIEeX-nzyd_f1F:bt07m6j) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L11)):*
> I don't like that those values are here. In my understanding the values should come from the either "network" where we keep chain specific values of from the genesis.

This function provides default values. They are then placed in NetworkConfig structure. So I believe it is exactly like you said.

> And why the entire file is in the "ante" package? You can keep in the the gas.go file and update to the interface, which will be implemented by the Network type/subtype to provide the values.

it's not in `ante` but in `x/auth/types`. It's because in cosmos everything related to validating fees and gas limits is inside `auth.
___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCZl1jf1gXRFDZ_OXtf:bsnse2r) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

Is is possible to keep the mappin in the map? 
</blockquote></details>

I was also thinking about creating a map like `map[reflect.Type]uint64` but let's do it in  separate PR maybe

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 1 of 21 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZfjWJCB8BpSFY7SJ3:-NC_NDpsAqrxswRQqtpb:b-63t876) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L11)):*

> I don't like that those values are here. In my understanding the values should come from the either "network" where we keep chain specific values of from the genesis.

This function provides default values. They are then placed in NetworkConfig structure. So I believe it is exactly like you said.

> And why the entire file is in the "ante" package? You can keep in the the gas.go file and update to the interface, which will be implemented by the Network type/subtype to provide the values.

it's not in `ante` but in `x/auth/types`. It's because in cosmos everything related to validating fees and gas limits is inside `auth.
</blockquote></details>

The same feeling - if the ante is abstract, let's keep it abstract without values, so the "DefaultDeterministicGasRequirements()" should live in the network config if you still need it.
___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NC_NbrZ8samxMkqL31i:b-gyk67p) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

I was also thinking about creating a map like `map[reflect.Type]uint64` but let's do it in  separate PR maybe
</blockquote></details>

map[reflect.Type]uint64 -> map[sdk.Msg]uint64, in that case you will be able to require that struct instead of the default values you return now in the efaultDeterministicGasRequirements(). And you will just fill it depending of the network.

dzmitryhil (commented):
> Reviewed 17 of 20 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCcsHbI53hlvO_F1nCH:bbhjj46) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

map[reflect.Type]uint64 -> map[sdk.Msg]uint64, in that case you will be able to require that struct instead of the default values you return now in the efaultDeterministicGasRequirements(). And you will just fill it depending of the network.
</blockquote></details>

`map[sdk.Msg]` won't work because messages are pointers, so I won't be able to just take the received messages and query the map. If we want to go with `map` then `reflect.Type` is the only option. And that's why I didn't decide to do it this way because reflection is slow.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCdCpq11Kw2a4tIrTYO:bfpg06x) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

`map[sdk.Msg]` won't work because messages are pointers, so I won't be able to just take the received messages and query the map. If we want to go with `map` then `reflect.Type` is the only option. And that's why I didn't decide to do it this way because reflection is slow.
</blockquote></details>

Then better to have a new strut { Message, Gas } rather then, map.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 20 of 21 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZfjWJCB8BpSFY7SJ3:-NCdL9IZAaEfys3Uel4K:b-m6ng5e) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L11)):*

The same feeling - if the ante is abstract, let's keep it abstract without values, so the "DefaultDeterministicGasRequirements()" should live in the network config if you still need it.
</blockquote></details>

Let's wait for this PR: https://reviewable.io/reviews/CoreumFoundation/coreum/191
Then I will move this code to `pkg/config`
___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCdLKkJ0Bf60h80aBSR:b-y79czh) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

Then better to have a new strut { Message, Gas } rather then, map. 
</blockquote></details>

At the end of the day you still need to use reflection, no matter if message is stored as key in the map or field in struct

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 16 of 21 files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`x/auth/types/deterministic.go` line 18 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZiMHO3oANh5hnT8Vv:-NCdQckMBfgb5UET-_Ol:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L18)):*

Why the FixedGas is param but the FreeBytes and FreeSignatures  are constant? Let's make them all parameters.
</blockquote></details>

Done.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 16 of 21 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZfjWJCB8BpSFY7SJ3:-NCe5IL632oQOpCDKq0y:b32w1t) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L11)):*

Let's wait for this PR: https://reviewable.io/reviews/CoreumFoundation/coreum/191
Then I will move this code to `pkg/config`
</blockquote></details>

Ok
___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCe5Ps-2AWPhceHMAFD:b-sv4hqy) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

At the end of the day you still need to use reflection, no matter if message is stored as key in the map or field in struct
</blockquote></details>

What is the use-case when you need the reflection?

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 16 of 21 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCe79ikD8pDxeUaBAxT:b-qcxqkp) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

What is the use-case when you need the reflection?
</blockquote></details>

1. I receive a message
2. Need to find correct deterministic gas
3. Need to compare received message with the dictionary, it must be done by type (not by value) so there are two options:
a: static, using `switch msg.(type)`
b: dynamic, using `reflect.TypeOf`

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 9 of 23 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`x/auth/types/deterministic.go` line 11 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZfjWJCB8BpSFY7SJ3:-NCeHcyB1kmiExethBeJ:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L11)):*

Ok
</blockquote></details>

Done.

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 9 of 23 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/auth/ante/gas.go` line 13 at r8](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCeVe-I8mJrnRpR0Yna:-NCeVe-JCQvemV4HHbjb:bqb1b0) ([raw file](https://github.com/coreumfoundation/coreum/blob/2551d514335f4aec236745158e44cb94a867c5a8/x/auth/ante/gas.go#L13)):*
> ```Go
> // CONTRACT: Must be the first decorator in the chain
> // CONTRACT: Tx must implement GasTx interface
> // FIXME (wojtek): THIS IS BAD, used only for testing
> ```

Didn't get that comment. Why we need it if it's bad ? And why it's bad?
___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCeVC8kE5hH_OXQ3Z1_:bhqvn1d) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*

1. I receive a message
2. Need to find correct deterministic gas
3. Need to compare received message with the dictionary, it must be done by type (not by value) so there are two options:
a: static, using `switch msg.(type)`
b: dynamic, using `reflect.TypeOf`
</blockquote></details>

Right, static is here :-) So we don't need the reflect.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 9 of 23 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`x/auth/ante/gas.go` line 13 at r8](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCeVe-I8mJrnRpR0Yna:-NCeYMyO0SEPf4o2b072:b-i4acxp) ([raw file](https://github.com/coreumfoundation/coreum/blob/2551d514335f4aec236745158e44cb94a867c5a8/x/auth/ante/gas.go#L13)):*

Didn't get that comment. Why we need it if it's bad ? And why it's bad?
</blockquote></details>

ah, it's a leftover
___
*[`x/auth/types/deterministic.go` line 31 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCZef6u7YsHprk02Q0R:-NCeYT8xCNQXDRMRHVQ3:b2jvmqt) ([raw file](https://github.com/coreumfoundation/coreum/blob/71b68bcb1a6e0bde0d08bfc7cc44d0e28ee6462d/x/auth/types/deterministic.go#L31)):*
But your original question was

> Is is possible to keep the mappin in the map?

So it might be done with reflection only

ysv (changes requested):
Finished review partially 
Going to continue on Monday

> Reviewed 1 of 20 files at r3, 1 of 13 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179#-:-NCf7fZr5rhlkSSeuz5B:b-nor1sy)* status: 10 of 23 files reviewed, 6 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`integration-tests/bank/gas.go` line 24 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf5OHlDG8TIS4YAxBN:-NCf5OHlDG8TIS4YAxBO:b-soonte) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L24)):*
> ```Go
> func TestTransferDeterministicGas(numOfTransactions int) testing.SingleChainSignature {
> 	return func(ctx context.Context, t testing.T, chain testing.Chain) {
> 		gasAssumed := chain.GasLimitByMsgs(&banktypes.MsgSend{})
> ```

`gasEstimated` ? here and some other places
___
*[`integration-tests/bank/gas.go` line 99 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf65rV8vJ163Qw0vI8:-NCf65rV8vJ163Qw0vI9:b-hg0xxn) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L99)):*
> ```Go
> 		1,
> 		amount,
> 	)))))
> ```

`)))))` a bit too much 
lets consider creating 1-2 local variables instead
___
*[`integration-tests/bank/gas.go` line 104 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf7Q6q5-aIhA7uJcF3:-NCf7Q6q5-aIhA7uJcF4:b-mdsqon) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L104)):*
> ```Go
> 	estimatedGas, err := chain.Client.EstimateGas(ctx, tx.BaseInput{
> 		Signer:   sender,
> 		GasLimit: 10 * gasAssumed, // set it to bigger value and check later that only `gasAssumed` is required
> ```

why do we actually pass GasLimit here ?

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 9 of 23 files reviewed, 5 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/bank/gas.go` line 24 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf5OHlDG8TIS4YAxBN:-NCsGbKx8WWiOwStt-ns:b-yd4w55) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L24)):*

`gasEstimated` ? here and some other places
</blockquote></details>

I don't think so. This is what I assume about the gas being used. But `gasExpected` is a better name I think
___
*[`integration-tests/bank/gas.go` line 99 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf65rV8vJ163Qw0vI8:-NCsKHw49XPrug1q7lAu:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L99)):*

`)))))` a bit too much 
lets consider creating 1-2 local variables instead
</blockquote></details>

Done.
___
*[`integration-tests/bank/gas.go` line 104 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf7Q6q5-aIhA7uJcF3:-NCsG62I7nv-0wyS6N1a:b-nh77bu) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L104)):*

why do we actually pass GasLimit here ?
</blockquote></details>

it's not needed

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 9 of 23 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

ysv (changes requested):
> Reviewed 1 of 8 files at r1, 4 of 20 files at r3, 2 of 2 files at r5, 1 of 1 files at r6, 10 of 13 files at r8, 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 19 of 23 files reviewed, 13 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`app/gas.go` line 80 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NBxnV912IjNlVEXGwnl:-NCtvM7JE8feJ5mub5Vn:b-b6sf3g) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L80)):*

Inspecting `RegisterService` of wrapped implementation might help to understand this piece:
https://github.com/cosmos/cosmos-sdk/blob/ff416ee63d32da5d520a8b2d16b00da762416146/baseapp/msg_service_router.go#L109
</blockquote></details>

can you leave add a reference to the file you mentioned above in comment
You can replace "this is magic" with referencing this comment
___
*[`app/gas.go` line 60 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtw7vZ4raNk0aijc1V:-NCtw7vZ4raNk0aijc1W:b-u2w7t7) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L60)):*
> ```Go
> 	for i, method := range sd.Methods {
> 		method := method
> 		sd.Methods[i].Handler = func(srv interface{}, ctx context.Context, dec func(interface{}) error, interceptor googlegrpc.UnaryServerInterceptor) (interface{}, error) {
> ```

is it possible to add some comments here to understand better. Because this part is very complicated to read
___
*[`app/gas.go` line 77 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtwV5sDeLvM2uIwLpb:-NCtwV5sDeLvM2uIwLpc:b-629kev) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L77)):*
> ```Go
> func ctxForDeterministicGas(ctx sdk.Context, msg sdk.Msg, deterministicGasRequirements config.DeterministicGasRequirements) sdk.Context {
> 	gasRequired := deterministicGasRequirements.GasRequiredByMessage(msg)
> 	if gasRequired > 0 {
> ```

do you think just having 0 here is safe enough ?
Maybe we should return `int, bool` and use it like this ? 
```
gasRequired, ok := deterministicGasRequirements.GasRequiredByMessage(msg)
if !ok {
  // ...
}
```

For me this seems to be more safe & explicit way. Even logically this method returns 0 which means that method required 0 gas which is not true
___
*[`app/gas.go` line 79 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtxCzZ4luz-nxDzMB5:-NCtxCzZ4luz-nxDzMB6:bw76lbi) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L79)):*
> ```Go
> 	if gasRequired > 0 {
> 		ctx.GasMeter().ConsumeGas(gasRequired, fmt.Sprintf("DeterministicGas (gas required: %d, message type: %T)", gasRequired, msg))
> 		ctx = ctx.WithGasMeter(sdk.NewGasMeter(5 * gasRequired))
> ```

can you explicitly define this 5 as constant with a good name and add comment to this
___
*[`integration-tests/index.go` line 26 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCti3ti-t9YANaTRyzh:-NCti3ti-t9YANaTRyzi:b-tubvei) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/integration-tests/index.go#L26)):*
> ```Go
> 			bank.TestCoreTransfer,
> 			bank.TestTransferFailsIfNotEnoughGasIsProvided,
> 			bank.TestTransferDeterministicGas(20),
> ```

does it still makes sense to run it 20 times here ?
because the value is the same al of the time, isn't nit ?
___
*[`integration-tests/bank/gas.go` line 24 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf5OHlDG8TIS4YAxBN:-NCtjnh_B-m7DAmmJeDi:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L24)):*

I don't think so. This is what I assume about the gas being used. But `gasExpected` is a better name I think
</blockquote></details>

ok
___
*[`integration-tests/bank/gas.go` line 99 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf65rV8vJ163Qw0vI8:-NCtjqan4ZpKj163W-nJ:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L99)):*

Done.
</blockquote></details>

ok
___
*[`integration-tests/bank/gas.go` line 104 at r9](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCf7Q6q5-aIhA7uJcF3:-NCtjrroASL2yhUt0jpw:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/bbefa474ec5afd227e905c8a23faa91466250762/integration-tests/bank/gas.go#L104)):*

it's not needed
</blockquote></details>

ok
___
*[`integration-tests/bank/gas.go` line 23 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtk2XZ-GElAQF45IRf:-NCtk2XZ-GElAQF45IRg:b-k5z5o3) ([raw file](https://github.com/coreumfoundation/coreum/blob/3a741c33ac6b8e5fe5c16b7c1afa907afe444b6c/integration-tests/bank/gas.go#L23)):*
> ```Go
> 
> // TestTransferMaximumGas checks that transfer does not take more gas than assumed
> func TestTransferMaximumGas(numOfTransactions int) testing.SingleChainSignature {
> ```

I'm just wondering if there is a way to see "real" non-deterministic gas usage with standard cosmos sdk gas meter.

Because during different msgs integration we will probably have to find this number from somewhere so we can define deterministic gas value.

Do you have smth in mind ? Maybe we can have some magical decorator/param ? Smth similar to way it is done in WASM ?
___
*[`x/auth/ante/ante.go` line 64 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCuIrAT6dX6IJLH13AE:-NCuIrAT6dX6IJLH13AF:b-ihgia1) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/ante/ante.go#L64)):*
> ```Go
> 	anteDecorators := []sdk.AnteDecorator{
> 		authante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
> 		NewSetupGasMeterDecorator(),
> ```

describe the way all these decorators work together in comment
___
*[`x/auth/ante/ante.go` line 64 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCuJnbz993HGg0hjRpV:-NCuJnbz993HGg0hjRpW:b-ja9vrr) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/ante/ante.go#L64)):*
> ```Go
> 	anteDecorators := []sdk.AnteDecorator{
> 		authante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
> 		NewSetupGasMeterDecorator(),
> ```

add check that tx has at least BaseGas



_Code quote:_
```Go
	NewFreeGasDecorator(infiniteAccountKeeper, options.DeterministicGasRequirements),
```
___
*[`x/auth/keeper/account.go` line 9 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtt6me6L8WWho1DXrU:-NCtt6me6L8WWho1DXrV:b-z4j2p) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/keeper/account.go#L9)):*
> ```Go
> )
> 
> func NewInfiniteAccountKeeper(ak ante.AccountKeeper) InfiniteAccountKeeper {
> ```

Add description for this keeper and why we need it
___
*[`x/auth/keeper/account.go` line 9 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCuK-jg76hitPNnK0st:-NCuK-jg76hitPNnK0su:b-rehdnb) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/keeper/account.go#L9)):*
> ```Go
> )
> 
> func NewInfiniteAccountKeeper(ak ante.AccountKeeper) InfiniteAccountKeeper {
> ```

quickly check why linter didn't report issue that the func is without comment
___
*[`pkg/config/deterministic.go` line 1 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtlIFN47gekpx1omNU:-NCtlIFN47gekpx1omNV:b-ss2cxh) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/pkg/config/deterministic.go#L1)):*
> ```Go
> package config
> ```

deterministic.go -> deterministig-gas.go ?

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 19 of 23 files reviewed, 11 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/bank/gas.go` line 23 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtk2XZ-GElAQF45IRf:-NCuP8TvA8vUkdFyNuoq:bmdirh8) ([raw file](https://github.com/coreumfoundation/coreum/blob/3a741c33ac6b8e5fe5c16b7c1afa907afe444b6c/integration-tests/bank/gas.go#L23)):*

I'm just wondering if there is a way to see "real" non-deterministic gas usage with standard cosmos sdk gas meter.

Because during different msgs integration we will probably have to find this number from somewhere so we can define deterministic gas value.

Do you have smth in mind ? Maybe we can have some magical decorator/param ? Smth similar to way it is done in WASM ?
</blockquote></details>

ok discussed on the call. There is no obvious way to do this
I think it makes sense to put a comment with the flow on how to define deterministic gas value: putting 0 for tx type as non-deterministic and then run integration test. Because it might seem non trivial task. I think this comments makes sense in DeterministicRequirements struct in BankSend part

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 12 of 23 files reviewed, 11 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`app/gas.go` line 80 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NBxnV912IjNlVEXGwnl:-NCx_Ca6BrLoyK4nmA1m:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L80)):*

can you leave add a reference to the file you mentioned above in comment
You can replace "this is magic" with referencing this comment
</blockquote></details>

Done.
___
*[`app/gas.go` line 60 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtw7vZ4raNk0aijc1V:-NCx_D7mBV8bYRAGUmma:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L60)):*

is it possible to add some comments here to understand better. Because this part is very complicated to read
</blockquote></details>

Done.
___
*[`app/gas.go` line 77 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtwV5sDeLvM2uIwLpb:-NCx_FGw9Lna9MY5c3Ku:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L77)):*

do you think just having 0 here is safe enough ?
Maybe we should return `int, bool` and use it like this ? 
```
gasRequired, ok := deterministicGasRequirements.GasRequiredByMessage(msg)
if !ok {
  // ...
}
```

For me this seems to be more safe & explicit way. Even logically this method returns 0 which means that method required 0 gas which is not true
</blockquote></details>

Done.
___
*[`app/gas.go` line 79 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtxCzZ4luz-nxDzMB5:-NCx_GJR95wFEjsw0rTW:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L79)):*

can you explicitly define this 5 as constant with a good name and add comment to this
</blockquote></details>

Done.
___
*[`integration-tests/index.go` line 26 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCti3ti-t9YANaTRyzh:-NCx_tuV6qTQB8FJFXvU:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/integration-tests/index.go#L26)):*

does it still makes sense to run it 20 times here ?
because the value is the same al of the time, isn't nit ?
</blockquote></details>

Done.
___
*[`integration-tests/bank/gas.go` line 23 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtk2XZ-GElAQF45IRf:-NCxaXeY-M5t3ZI3PNDR:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/3a741c33ac6b8e5fe5c16b7c1afa907afe444b6c/integration-tests/bank/gas.go#L23)):*

ok discussed on the call. There is no obvious way to do this
I think it makes sense to put a comment with the flow on how to define deterministic gas value: putting 0 for tx type as non-deterministic and then run integration test. Because it might seem non trivial task. I think this comments makes sense in DeterministicRequirements struct in BankSend part
</blockquote></details>

Done.
___
*[`x/auth/ante/ante.go` line 64 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCuJnbz993HGg0hjRpV:-NCxcdGS9NLXRvHRV3ms:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/ante/ante.go#L64)):*

add check that tx has at least BaseGas
</blockquote></details>

Done.
___
*[`x/auth/ante/ante.go` line 64 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCuIrAT6dX6IJLH13AE:-NCxgcl6Ea9otUPit_vC:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/ante/ante.go#L64)):*

describe the way all these decorators work together in comment
</blockquote></details>

Done.
___
*[`x/auth/keeper/account.go` line 9 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtt6me6L8WWho1DXrU:-NCxdN2U8AEld7gGI1tn:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/keeper/account.go#L9)):*

Add description for this keeper and why we need it
</blockquote></details>

Done.
___
*[`pkg/config/deterministic.go` line 1 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtlIFN47gekpx1omNU:-NCxagC90_LuSi8u7PfT:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/pkg/config/deterministic.go#L1)):*

deterministic.go -> deterministig-gas.go ?
</blockquote></details>

Done.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 12 of 23 files reviewed, 11 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`x/auth/keeper/account.go` line 9 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCuK-jg76hitPNnK0st:-NCxkQtr3uUdc_qMMQpX:b-u1tkpe) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/keeper/account.go#L9)):*

quickly check why linter didn't report issue that the func is without comment
</blockquote></details>

weird, it seems that this piece of config:

```
  revive:
    rules:
      - name: package-comments
        disabled: true
```

disables not only complains about missing package level  comments but about all the comments

ysv (changes requested):
> Reviewed 2 of 2 files at r11, 1 of 1 files at r12, 8 of 8 files at r13, 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`app/gas.go` line 80 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NBxnV912IjNlVEXGwnl:-NCzSrNd9ocmehd9GJQ0:b4c9lyr) ([raw file](https://github.com/coreumfoundation/coreum/blob/6ae3a47377fb7003212c583db472016091bee6b5/app/gas.go#L80)):*

Done.
</blockquote></details>

:thumbsup: 
___
*[`app/gas.go` line 79 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtxCzZ4luz-nxDzMB5:-NCzT2Rk43bjWdFCL1MI:b-51u6ks) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L79)):*

Done.
</blockquote></details>

worth changing to:
`We pass much higher amount of gas to handler to be sure that it succeeds. We want to avoid passing infinite gas meter to alway have limit in case of mistake.`
___
*[`app/gas.go` line 17 at r14](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCzVzIW3hcnj46mOukm:-NCzVzIW3hcnj46mOukn:b-13c68z) ([raw file](https://github.com/coreumfoundation/coreum/blob/6d9e173f720dfee8e6c5134ecea22756274dca4e/app/gas.go#L17)):*
> ```Go
> 
> // NewDeterministicGasRouter returns wrapped router charging deterministic amount of gas for defined message types
> func NewDeterministicGasRouter(baseRouter sdk.Router, deterministicGasRequirements config.DeterministicGasRequirements) sdk.Router {
> ```

I'm just thinking if we should group all these functionality related to deterministic fees into a single pkg ? I'm not sure what is the right place for it ./app/deterministic-fees ?

and have in this pkg these parts related to our fees:  
- deterministicGasRouter
- GasMeterDecorator
- GasDecorator
- FinalGasDecorator
___
*[`x/auth/ante/gas.go` line 18 at r14](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCzXLruCRih7uMmQhdv:-NCzXLruCRih7uMmQhdw:bfqun1v) ([raw file](https://github.com/coreumfoundation/coreum/blob/6d9e173f720dfee8e6c5134ecea22756274dca4e/x/auth/ante/gas.go#L18)):*
> ```Go
> 
> // NewSetupGasMeterDecorator creates new SetupGasMeterDecorator
> func NewSetupGasMeterDecorator(deterministicGasRequirements config.DeterministicGasRequirements) SetupGasMeterDecorator {
> ```

what do you think about putting a bit better names to these decorators:

1. SetupGasMeterDecorator -> SetInfinityGasMeterDecorator
2. FreeGasDecorator -> AddBaseGasDecorator & also rename `DeterministicGasRequirements.TxBonusGas -> TxBaseGas`
3. FinalGasDecorator -> ChargeFixedGasDecorator

we can discuss these names
___
*[`x/auth/keeper/account.go` line 9 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCuK-jg76hitPNnK0st:-NCzUsQs0beFvcwGjSIT:b-h3eknw) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/x/auth/keeper/account.go#L9)):*

weird, it seems that this piece of config:

```
  revive:
    rules:
      - name: package-comments
        disabled: true
```

disables not only complains about missing package level  comments but about all the comments
</blockquote></details>

strange
lets fix in other PR I will create a task
___
*[`pkg/config/deterministic_gas.go` line 52 at r14](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCziZdFC722HKoKX9-M:-NCziZdFC722HKoKX9-N:bf6op8o) ([raw file](https://github.com/coreumfoundation/coreum/blob/6d9e173f720dfee8e6c5134ecea22756274dca4e/pkg/config/deterministic_gas.go#L52)):*
> ```Go
> 
> // TxBonusGas is the free gas we give to every transaction to cover costs of tx size and signature verification
> func (dgr DeterministicGasRequirements) TxBonusGas(params authtypes.Params) uint64 {
> ```

I'm a bit confused by TxBonusGas vs FixedGas
So as far as I understand FixedGas is amount we charge & TxBonusGas is amount we give for executing base handlers ?

I suggest renaming TxBonusGas to TxBaseGas/BaseGas

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 4 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`pkg/config/deterministic_gas.go` line 52 at r14](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCziZdFC722HKoKX9-M:-ND1vW7mCUs48VT94kMO:b3uijrw) ([raw file](https://github.com/coreumfoundation/coreum/blob/6d9e173f720dfee8e6c5134ecea22756274dca4e/pkg/config/deterministic_gas.go#L52)):*
> So as far as I understand FixedGas is amount we charge & TxBonusGas is amount we give for executing base handlers ?

yes

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: all files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`app/gas.go` line 79 at r10](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCtxCzZ4luz-nxDzMB5:-ND1xvs53b4J2LExXOJk:bb6un1j) ([raw file](https://github.com/coreumfoundation/coreum/blob/13088a9f1018ee934839423eb205abde434daa10/app/gas.go#L79)):*

worth changing to:
`We pass much higher amount of gas to handler to be sure that it succeeds. We want to avoid passing infinite gas meter to alway have limit in case of mistake.`
</blockquote></details>

Done
___
*[`x/auth/ante/gas.go` line 18 at r14](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCzXLruCRih7uMmQhdv:-ND2-aW05BAnDl5S_u7l:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/6d9e173f720dfee8e6c5134ecea22756274dca4e/x/auth/ante/gas.go#L18)):*

what do you think about putting a bit better names to these decorators:

1. SetupGasMeterDecorator -> SetInfinityGasMeterDecorator
2. FreeGasDecorator -> AddBaseGasDecorator & also rename `DeterministicGasRequirements.TxBonusGas -> TxBaseGas`
3. FinalGasDecorator -> ChargeFixedGasDecorator

we can discuss these names
</blockquote></details>

Done.
___
*[`pkg/config/deterministic_gas.go` line 52 at r14](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCziZdFC722HKoKX9-M:-ND1yU1c6JhxJSGoFVtB:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/6d9e173f720dfee8e6c5134ecea22756274dca4e/pkg/config/deterministic_gas.go#L52)):*

> So as far as I understand FixedGas is amount we charge & TxBonusGas is amount we give for executing base handlers ?

yes
</blockquote></details>

Done.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: 18 of 23 files reviewed, 3 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`app/gas.go` line 17 at r14](https://reviewable.io/reviews/coreumfoundation/coreum/179#-NCzVzIW3hcnj46mOukm:-ND22LhX44IU8jdDTLU8:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/6d9e173f720dfee8e6c5134ecea22756274dca4e/app/gas.go#L17)):*

I'm just thinking if we should group all these functionality related to deterministic fees into a single pkg ? I'm not sure what is the right place for it ./app/deterministic-fees ?

and have in this pkg these parts related to our fees:  
- deterministicGasRouter
- GasMeterDecorator
- GasDecorator
- FinalGasDecorator

</blockquote></details>

Done.

ysv (approved):
> Reviewed 5 of 5 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/179)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @begmaroman, @dzmitryhil, and @miladz68)

PR #191, opened 2022-09-21
Create separate package for config so it is not required to compile other binaries with WASM support

wojtek-coreum:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/191)


wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/191)* status: 0 of 23 files reviewed, 1 unresolved discussion (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/191#-NCUUjDqDvIXADnYC_LF:-NCUUjDqDvIXADnYC_LG:bq27z5b) (no related file):*
First commit is about moving files to new directory
Second commit is about updating stuff in new `config` package
Third commit is about switching to `config` package everywhere

___

dzmitryhil (dismissed):
> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/191)* status: all files reviewed, 1 unresolved discussion (waiting on @begmaroman, @miladz68, and @ysv)

ysv (changes requested):
> Reviewed 22 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/191)* status: all files reviewed, 2 unresolved discussions (waiting on @begmaroman, @miladz68, and @wojtek-coreum)

___
*[`pkg/config/network_test.go` line 72 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/191#-NCdy--IBun2X9I-KY4h:-NCdy--IBun2X9I-KY4i:b-113ujo) ([raw file](https://github.com/coreumfoundation/coreum/blob/ac340c926c2726a3c40982ab8e4422b58199e548/pkg/config/network_test.go#L72)):*
> ```Go
> }
> 
> func knownNetworks(t *testing.T) []config.Network {
> ```

First of all I didn't like the way this func looks. So I decided to dig dipper on its puprpose and I suggest to rewise approach taken here.

The issue is that we have only devnet hardcoded here. As result if we add another network in config we should change this func also. I suggest instead to do the following:

in config pkg add `func EnabledNetworks() []config.Network` (or any other name) and use it instead of knownNetworks

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/191)* status: 21 of 23 files reviewed, 2 unresolved discussions (waiting on @begmaroman, @dzmitryhil, @miladz68, and @ysv)

___
*[`pkg/config/network_test.go` line 72 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/191#-NCdy--IBun2X9I-KY4h:-NCe3Qm-3bpmc453z59c:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/ac340c926c2726a3c40982ab8e4422b58199e548/pkg/config/network_test.go#L72)):*

First of all I didn't like the way this func looks. So I decided to dig dipper on its puprpose and I suggest to rewise approach taken here.

The issue is that we have only devnet hardcoded here. As result if we add another network in config we should change this func also. I suggest instead to do the following:

in config pkg add `func EnabledNetworks() []config.Network` (or any other name) and use it instead of knownNetworks
</blockquote></details>

Done.

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/191)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @begmaroman and @miladz68)

PR #226, opened 2022-10-14
Add bank keeper wrapper with the integration with the asset keeper

dzmitryhil:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)


miladz68 (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 9 unresolved discussions (waiting on @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`app/app.go` line 252 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKuhyQEPYovCXl-Gp9:-NEKuhyQEPYovCXl-GpA:b-ibosii) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/app/app.go#L252)):*
> ```Go
> 	AssetKeeper       assetkeeper.Keeper
> 	FeeModelKeeper    feemodelkeeper.Keeper
> 	WrappedBankKeeper wbankkeeper.BaseKeeperWrapper
> ```

let's keep it as `BankeKeeper` since `bank.Keeper` is just an interface, and `WrapedBankBaseKeeper` interface correctly implements that interface. 
___
*[`app/app.go` line 341 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKvTUjECEfU1O63aHq:-NEKvTUjECEfU1O63aHr:bla96nn) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/app/app.go#L341)):*
> ```Go
> 		appCodec,
> 		keys[assettypes.StoreKey],
> 		&app.WrappedBankKeeper,
> ```

I see that you have passed a pointer here and the struct will be initialized later, to avoid circular dependency between asset keeper and wrapped bank keeper. 
I am thinking for a way around this but cannot find one. we should put more thought to get around this. 
___
*[`x/bank/module.go` line 15 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NELAwxtAy0no3VZmpLQ:-NELAwxuE-5FYYZVkf7k:b-pd7pu8) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/module.go#L15)):*
> ```Go
> 
> // AppModuleBasic defines the basic application module used by the wrapped bank module.
> type AppModuleBasic struct {
> ```

I think we can avoid this wrapper by just passing a wrapped keeper as an interface to the original bank AppModule.
___
*[`x/bank/keeper/keeper.go` line 31 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKrY_I-JZU6vz2BR4V:-NEKrY_I-JZU6vz2BR4W:b-em3x1f) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L31)):*
> ```Go
> ) BaseKeeperWrapper {
> 	return BaseKeeperWrapper{
> 		BaseKeeper: bankkeeper.NewBaseKeeper(cdc, storeKey, ak, paramSpace, blockedAddrs),
> ```

I see that the you have embeded the BaseKeepr into BaseKeeperWrapper, so it means that all the methods are provided by default unless we choose to overwrite them. That is a great choice :thumbsup: 
___
*[`x/bank/keeper/keeper.go` line 57 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKredX68IDeli44-Ge:-NEKredX68IDeli44-Gf:bv8m6tf) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L57)):*
> ```Go
> // It will panic if either module account does not exist.
> // !!! The code is the copy of the corresponding func of the bank module !!!
> func (k BaseKeeperWrapper) SendCoinsFromModuleToModule(ctx sdk.Context, senderModule, recipientModule string, amt sdk.Coins) error {
> ```

sending from module to module does not concern FT, I think this should be removed. 
___
*[`x/bank/keeper/keeper.go` line 74 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKsj528RTKNqJuDXx4:-NEKsj528RTKNqJuDXx5:b-n4ppob) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L74)):*
> ```Go
> // It will panic if the module account does not exist.
> // !!! The code is the copy of the corresponding func of the bank module !!!
> func (k BaseKeeperWrapper) SendCoinsFromAccountToModule(ctx sdk.Context, senderAddr sdk.AccAddress, recipientModule string, amt sdk.Coins) error {
> ```

again, an account cannot be whitelisted or black listed, so we need to remove this. 
___
*[`x/bank/keeper/keeper.go` line 89 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKt_rx3rHxXl6xA1tJ:-NEKt_rx3rHxXl6xA1tK:b-3f8va1) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L89)):*
> ```Go
> 	}
> 
> 	balances := k.GetAllBalances(ctx, fromAddr)
> ```

get all balances is a bad idea because of performance. 
___
*[`x/bank/keeper/keeper.go` line 90 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKu7tm2QFQDBOxXq6n:-NEKu7tn9O23KK1YZQaD:b-hmkpz3) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L90)):*
> ```Go
> 
> 	balances := k.GetAllBalances(ctx, fromAddr)
> 	locked := k.ftProvider.GetLockedCoins(ctx, fromAddr, balances)
> ```

I think this `GetLockedCoins` needs to be improved for performance as this part of the app is very performance critical. A better idea is to check whether the request amount to transfer is spendable or not. 
___
*[`x/bank/keeper/benchmark_test.go` line 18 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKj_V8D7y7d2hqNxsW:-NEKj_V8D7y7d2hqNxsX:b-58zbds) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/benchmark_test.go#L18)):*
> ```Go
> 
> func Benchmark100KDenomBankSend(b *testing.B) {
> 	simApp := simapp.New()
> ```

Did you run these benchmarks to see the effect ?
___
*[`x/bank/keeper/benchmark_test.go` line 22 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NELBAlfBCXxKmCMQUM8:-NELBAlfBCXxKmCMQUM9:b-b1jmon) ([raw file](https://github.com/coreumfoundation/coreum/blob/0f21633b7142db20a4b3299404f5adc2b58a2145/x/bank/keeper/benchmark_test.go#L22)):*
> ```Go
> func createSimApp(b *testing.B) *app.App {
> 	simapp.FlagEnabledValue = true
> 	_, db, dir, logger, _, err := simapp.SetupSimulation("coreum-app-sim", "Simulation")
> ```

The original Idea behind using this function `createSimApp` and not the sim app in the app package, is that this one is using temporary file to run the database. Did you check that the simapp in testutil uses temp files ? otherwise these benchamrks would be off.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: 5 of 8 files reviewed, 9 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`app/app.go` line 252 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKuhyQEPYovCXl-Gp9:-NELfoI7-NOF-8im7WSB:b8m3gg3) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/app/app.go#L252)):*

let's keep it as `BankeKeeper` since `bank.Keeper` is just an interface, and `WrapedBankBaseKeeper` interface correctly implements that interface. 
</blockquote></details>

Ok, restored.



_Code quote:_
```Go
WrappedBankKeeper
```
___
*[`app/app.go` line 341 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKvTUjECEfU1O63aHq:-NELfW5159TCEAs6rmG6:b-9z7omf) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/app/app.go#L341)):*

I see that you have passed a pointer here and the struct will be initialized later, to avoid circular dependency between asset keeper and wrapped bank keeper. 
I am thinking for a way around this but cannot find one. we should put more thought to get around this. 
</blockquote></details>

I won't find it without the full refactoring of the BankKeeper :-)
___
*[`x/bank/module.go` line 15 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NELAwxtAy0no3VZmpLQ:-NELfuiB5Hk5-dZKY-kD:bs6p1pp) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/module.go#L15)):*

I think we can avoid this wrapper by just passing a wrapped keeper as an interface to the original bank AppModule.
</blockquote></details>

We can't since it the bank module uses the direct cust to type in the "RegisterServices"
___
*[`x/bank/keeper/keeper.go` line 57 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKredX68IDeli44-Ge:-NELgeXf1olwSlw6LE7c:b-gt443o) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L57)):*

sending from module to module does not concern FT, I think this should be removed. 
</blockquote></details>

All of them uses the "Send" function if we don't do it that way, the "SendCoinsFromModuleToModule" will use the "Send" on the original keeper.
___
*[`x/bank/keeper/keeper.go` line 74 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKsj528RTKNqJuDXx4:-NELgwOn4Fsr7WN7lQOd:b-1w89i1) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L74)):*

again, an account cannot be whitelisted or black listed, so we need to remove this. 
</blockquote></details>

See the comment above.+ If the token is frozen, you can't send it anywhere. 
___
*[`x/bank/keeper/keeper.go` line 89 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKt_rx3rHxXl6xA1tJ:-NELh7V05bI5YCDaj8lS:buvlc5x) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L89)):*

get all balances is a bad idea because of performance. 
</blockquote></details>

1This is just an example of the integration, it can be remove in the real imp
___
*[`x/bank/keeper/keeper.go` line 90 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKu7tm2QFQDBOxXq6n:-NELhHWiBPDr_cZxF689:b75b5i7) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L90)):*

I think this `GetLockedCoins` needs to be improved for performance as this part of the app is very performance critical. A better idea is to check whether the request amount to transfer is spendable or not. 
</blockquote></details>

This is just an example. For sure the real impl will be different.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: 5 of 8 files reviewed, 9 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/bank/keeper/benchmark_test.go` line 18 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKj_V8D7y7d2hqNxsW:-NELmYyIBkBDjjUId1Gi:bdl26qe) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/benchmark_test.go#L18)):*

Did you run these benchmarks to see the effect ?
</blockquote></details>

restored prev.
___
*[`x/bank/keeper/benchmark_test.go` line 22 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NELBAlfBCXxKmCMQUM8:-NELmKdY1GP1Q2hlCBdD:b-l7qkzy) ([raw file](https://github.com/coreumfoundation/coreum/blob/0f21633b7142db20a4b3299404f5adc2b58a2145/x/bank/keeper/benchmark_test.go#L22)):*

The original Idea behind using this function `createSimApp` and not the sim app in the app package, is that this one is using temporary file to run the database. Did you check that the simapp in testutil uses temp files ? otherwise these benchamrks would be off. 
</blockquote></details>

Restored the previous version. But I see now that we use the CLI simulation to run it. Not in the scope of that PR,  but in general we should move such functions for the simulation to our simapp.

miladz68 (dismissed):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum and @ysv)

wojtek-coreum (changes requested):
> Reviewed 4 of 8 files at r1, 1 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 6 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`app/app.go` line 108 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdh26a0-LXFB9ca8yF:-NEdh26a0-LXFB9ca8yG:bt767co) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/app/app.go#L108)):*
> ```Go
> 	assettypes "github.com/CoreumFoundation/coreum/x/asset/types"
> 	"github.com/CoreumFoundation/coreum/x/auth/ante"
> 	wbank "github.com/CoreumFoundation/coreum/x/bank"
> ```

This wrapper does not make sense without `asset` module so I would move it to `x/asset/bank` and rename import to `assetbank`
___
*[`app/app.go` line 341 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdgljc7ZkwHWtKCH8t:-NEdgljd5U_DT2gAEtKn:br0ickn) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/app/app.go#L341)):*
> ```Go
> 		appCodec,
> 		keys[assettypes.StoreKey],
> 		&app.BankKeeper,
> ```

If you need to take pointer here it means that maybe keeper should be stored by pointer?

```
BankKeeper     *wbankkeeper.BaseKeeperWrapper
```
___
*[`x/bank/module.go` line 41 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdj4bq7Up6p6Q_3ZCk:-NEdj4br8G6ch3p1rJUk:b-lanq09) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/x/bank/module.go#L41)):*
> ```Go
> 
> 	m := bankkeeper.NewMigrator(am.keeper.BaseKeeper)
> 	if err := cfg.RegisterMigration(banktypes.ModuleName, 1, m.Migrate1to2); err != nil {
> ```

How do we know when new migration is added to the original module?
___
*[`x/bank/keeper/keeper.go` line 89 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKt_rx3rHxXl6xA1tJ:-NEdk8C80AIAMX7K4UFE:bbk1vfs) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L89)):*

1This is just an example of the integration, it can be remove in the real imp
</blockquote></details>

Isn't this the real implementation?
___
*[`x/bank/keeper/keeper.go` line 90 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKu7tm2QFQDBOxXq6n:-NEdkLc2BgvnMxJ0mLC7:bxwjmev) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L90)):*

This is just an example. For sure the real impl will be different. 
</blockquote></details>

same thing here
___
*[`x/bank/types/expected_keepers.go` line 7 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdhp5KBQ2cbmsesj7A:-NEdhp5KBQ2cbmsesj7B:b-8dx9i4) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/x/bank/types/expected_keepers.go#L7)):*
> ```Go
> // FungibleTokenProvider defines an interface to interact with the fungible token functionality.
> type FungibleTokenProvider interface {
> 	GetLockedCoins(ctx sdk.Context, address sdk.AccAddress, currentBalances sdk.Coins) sdk.Coins
> ```

why do you provide `currentBalances` here? If it's needed, provider might query for it.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 6 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`app/app.go` line 108 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdh26a0-LXFB9ca8yF:-NEdysPe0l6jM00a_YMG:b-3xggki) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/app/app.go#L108)):*

This wrapper does not make sense without `asset` module so I would move it to `x/asset/bank` and rename import to `assetbank`
</blockquote></details>

I would go with that approach since it's more visible. You see that the bank will be replace once you open the repo.
___
*[`app/app.go` line 341 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdgljc7ZkwHWtKCH8t:-NEdz4Nv599ISdTKrkjH:b-799w9e) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/app/app.go#L341)):*

If you need to take pointer here it means that maybe keeper should be stored by pointer?

```
BankKeeper     *wbankkeeper.BaseKeeperWrapper
```
</blockquote></details>

This is the strategy of the each module to work with the interfaces, so I would keep the interface here.
___
*[`x/bank/module.go` line 41 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdj4bq7Up6p6Q_3ZCk:-NEdzG4LBQmk4Azz10W-:bltacxd) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/x/bank/module.go#L41)):*

How do we know when new migration is added to the original module?
</blockquote></details>

I didn't get yout question. 
___
*[`x/bank/keeper/keeper.go` line 89 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKt_rx3rHxXl6xA1tJ:-NEe-UKB6Fr18zM8JE6o:b-kdu9kc) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L89)):*

Isn't this the real implementation?
</blockquote></details>

Nope, this is sample to show 2 way interaction. 
___
*[`x/bank/keeper/keeper.go` line 90 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKu7tm2QFQDBOxXq6n:-NEe-ZQJ-YtYgihuRKrp:bbwjvc3) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L90)):*
> Nope, this is sample to show 2 way interaction. 
Nope, this is sample to show 2 way interaction. 
___
*[`x/bank/types/expected_keepers.go` line 7 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdhp5KBQ2cbmsesj7A:-NEe-a2c5vNDIOVtLcQY:b-i2fbhh) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/x/bank/types/expected_keepers.go#L7)):*

why do you provide `currentBalances` here? If it's needed, provider might query for it.
</blockquote></details>

This is sample to show 2 way interaction. The real impl will be different.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`app/app.go` line 341 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdgljc7ZkwHWtKCH8t:-NEfLenrBNAdWw-eKzGk:b-7aj07k) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/app/app.go#L341)):*

This is the strategy of the each module to work with the interfaces, so I would keep the interface here.
</blockquote></details>

Now I see that you don't need to pass pointer here
___
*[`x/bank/module.go` line 41 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdj4bq7Up6p6Q_3ZCk:-NEfLtnVB5JmuESyJHpc:b-aut448) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/x/bank/module.go#L41)):*

I didn't get yout question. 
</blockquote></details>

At some point new cosmos sdk version is released containing some new migration in bank module (let's say). How will we find that our implementation is broken? It might not fail immediately in tests.
___
*[`x/bank/keeper/keeper.go` line 89 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKt_rx3rHxXl6xA1tJ:-NEfMHSY3AGLefsUGF5m:b-n07gwa) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L89)):*

Nope, this is sample to show 2 way interaction. 
</blockquote></details>

I don't get it. Why don't we implement the real one instead?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`app/app.go` line 341 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdgljc7ZkwHWtKCH8t:-NEfObolD0RW76vn8mwq:b-o2x32l) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/app/app.go#L341)):*

Now I see that you don't need to pass pointer here
</blockquote></details>

Why? I need it to be with the assetKeeper which will be set later
___
*[`x/bank/keeper/keeper.go` line 89 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKt_rx3rHxXl6xA1tJ:-NEfORdH7EZsoGAb6eq-:b-p2wsg) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L89)):*

I don't get it. Why don't we implement the real one instead?
</blockquote></details>

It will be implemented under another ticket.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`x/bank/module.go` line 41 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdj4bq7Up6p6Q_3ZCk:-NEfOuR_0LnXH0sOUoFU:bt9tzb5) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/x/bank/module.go#L41)):*

At some point new cosmos sdk version is released containing some new migration in bank module (let's say). How will we find that our implementation is broken? It might not fail immediately in tests.
</blockquote></details>

Just review. Do you have a better approach?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`app/app.go` line 341 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdgljc7ZkwHWtKCH8t:-NEitWzTCRFV_8yiTTAy:b-x98v9d) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/app/app.go#L341)):*

I just removed `&` and everything is fine so pointer is not needed and original bank keeper also was used by value, no?
</blockquote></details>

Updated to direct usage of the keeper as we agreed. 
___
*[`x/bank/module.go` line 41 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEdj4bq7Up6p6Q_3ZCk:-NEj0VYB539jwT7A1rjY:bis0m37) ([raw file](https://github.com/coreumfoundation/coreum/blob/92f29f969da431e1f3506e2ad5ff5dd79bf27fa7/x/bank/module.go#L41)):*

At least a unit test maybe verifying that version in original bank was not incremented.
</blockquote></details>

Agree. Done.
___
*[`x/bank/keeper/keeper.go` line 89 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEKt_rx3rHxXl6xA1tJ:-NEj0QtSATcuPs0nCZCe:b-3ssake) ([raw file](https://github.com/coreumfoundation/coreum/blob/67772fbc14233f2c2d3057ae42b6dba2e1c10501/x/bank/keeper/keeper.go#L89)):*

So do we must to merge this PR? What's the reason for having temporary not used code in master?
</blockquote></details>

removed that code, and left TODO to implement the right version

miladz68 (dismissed):
> Reviewed 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum and @ysv)

miladz68 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum and @ysv)

wojtek-coreum (changes requested):
> Reviewed 5 of 5 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 4 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`app/app.go` line 342 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkHyLY80NYN-K-JTC7:-NEkHyLY80NYN-K-JTC8:b-wooxc8) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/app/app.go#L342)):*
> ```Go
> 		keys[assettypes.StoreKey],
> 		// for the asset we use the clear bank keeper without the assets integration to prevent cycling calls.
> 		bankkeeper.NewBaseKeeper(appCodec, keys[banktypes.StoreKey], app.AccountKeeper, app.GetSubspace(banktypes.ModuleName), app.ModuleAccountAddrs()),
> ```

This will create troubles immediately. In snapshot PoC for the purpose of testing I'm adding simple airdrop functionality, which needs to transfer coins. By using bank keeper passed to asset module it will use the wrong keeper meaning that all the conditions (freezing) won't  be checked. I told it yesterday and I'm repeating it now, let's not use two different bank keeper implementations. Sooner then later we will screw up the blockchain by doing this.
___
*[`pkg/store/key_test.go` line 23 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkIq4b7KdoCtK_YKNO:-NEkIq4b7KdoCtK_YKNP:bl4d1yl) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/pkg/store/key_test.go#L23)):*
> ```Go
> 	exp := make([]byte, 0)
> 	exp = append(exp, prefix...)
> 	exp = append(exp, []byte(strconv.Itoa(len(denom)))...)
> ```

You didn't update to use proto here
___
*[`x/asset/keeper/keeper.go` line 95 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkIyqE-lkRX7c9k4P0:-NEkIyqE-lkRX7c9k4P1:b-auh0rk) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/x/asset/keeper/keeper.go#L95)):*
> ```Go
> }
> 
> // GetLockedCoins consumes returns the locked coins for the account.
> ```

remove "consumes"
___
*[`x/bank/module_test.go` line 1 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkJTK09ucwPiFjD1rK:-NEkJTK1Cx4399Y_jvqd:b-rnkvnw) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/x/bank/module_test.go#L1)):*
> ```Go
> package bank
> ```

I didn't review this file fully because I expected very simple thing:

```go
assert.Equal(t, 2, bank.AppModule{}.ConsensusVersion())
```

Can't it be done this way?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`app/app.go` line 342 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkHyLY80NYN-K-JTC7:-NEkPPyP1bEdbANmd1n1:blv3ncu) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/app/app.go#L342)):*

This will create troubles immediately. In snapshot PoC for the purpose of testing I'm adding simple airdrop functionality, which needs to transfer coins. By using bank keeper passed to asset module it will use the wrong keeper meaning that all the conditions (freezing) won't  be checked. I told it yesterday and I'm repeating it now, let's not use two different bank keeper implementations. Sooner then later we will screw up the blockchain by doing this.
</blockquote></details>

Then you missed the final solution we agreed on yesterday. It was to use different. So we will need one more round today.
___
*[`pkg/store/key_test.go` line 23 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkIq4b7KdoCtK_YKNO:-NEkPeBN-4vAWAw0U15Y:b-bu91xm) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/pkg/store/key_test.go#L23)):*

You didn't update to use proto here
</blockquote></details>

Updated
___
*[`x/asset/keeper/keeper.go` line 95 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkIyqE-lkRX7c9k4P0:-NEkPp9UBHkOM3RRq8gK:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/x/asset/keeper/keeper.go#L95)):*

remove "consumes"
</blockquote></details>

Done.
___
*[`x/bank/module_test.go` line 1 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkJTK09ucwPiFjD1rK:-NEkPwlO-Dg3QaYBBpjT:b-kemt1u) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/x/bank/module_test.go#L1)):*

I didn't review this file fully because I expected very simple thing:

```go
assert.Equal(t, 2, bank.AppModule{}.ConsensusVersion())
```

Can't it be done this way?
</blockquote></details>

Didn't get you. We don't override the ConsensusVersion() why we should test it?

miladz68 (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 4 unresolved discussions (waiting on @wojtek-coreum and @ysv)

miladz68 (approved):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 4 unresolved discussions (waiting on @wojtek-coreum and @ysv)

wojtek-coreum (changes requested):
> Reviewed 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`x/bank/module_test.go` line 1 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkJTK09ucwPiFjD1rK:-NEkYg6iEyBe_mTglHQD:b-re8gmz) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/x/bank/module_test.go#L1)):*

Didn't get you. We don't override the ConsensusVersion() why we should test it?
</blockquote></details>

To verify that new version of cosmos sdk doesn't make any changes in state of bank, by providing new migration we don't include in our wrapper. Whenever there is new migration value returned from `ConsensusVersion()` should be incremented, right?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 2 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`x/bank/module_test.go` line 1 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkJTK09ucwPiFjD1rK:-NEkb3Qx9wlRt_8Mbt3h:b-hnhgh0) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/x/bank/module_test.go#L1)):*

To verify that new version of cosmos sdk doesn't make any changes in state of bank, by providing new migration we don't include in our wrapper. Whenever there is new migration value returned from `ConsensusVersion()` should be incremented, right?
</blockquote></details>

Not sure :-) That's why the test is a bit different. It will fail in case any new migration is registrated

wojtek-coreum (changes requested):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil and @ysv)

___
*[`x/bank/module_test.go` line 1 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NEkJTK09ucwPiFjD1rK:-NEkpgRq1d3Sda4PGow7:b-5ywduv) ([raw file](https://github.com/coreumfoundation/coreum/blob/1983b5c9aae9f2fc2b4df05179f0e2dce7882a90/x/bank/module_test.go#L1)):*

Not sure :-) That's why the test is a bit different. It will fail in case any new migration is registrated
</blockquote></details>

Please add that one assertion to the test and I will accept it. I will try to investigate the relationship between migrations and consensus version tomorrow.

wojtek-coreum (approved):
> Reviewed 1 of 1 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ysv)

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/226)* status: all files reviewed, 1 unresolved discussion

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NFEuLBU9BsHPeSeXnOd:-NFT6jjQ3BhJuKb0_Gju:b-896fix) (no related file):*

Would be nice if you could make a quick guide through the code showing how things work in this patch
</blockquote></details>

Done.

___
___
*[`pkg/store/key.go` line 15 at r12](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NFEpems4_EC8YK9nMVT:-NFT6mXr0rOd8Jv4K-jM:bvo83y5) ([raw file](https://github.com/coreumfoundation/coreum/blob/56de3bc77f796606a09fbb20a63eb606bf466957/pkg/store/key.go#L15)):*

I still don't get why we need this trick can you elaborate pls ?
I mean I understand the issue it solves but see the only place where it is used (`func GetFungibleTokenKey(denom string) []byte`) and not sure why it is needed there
</blockquote></details>

Explained.
___
*[`testutil/simapp/simapp.go` line 45 at r12](https://reviewable.io/reviews/coreumfoundation/coreum/226#-NFEo53a8WBOxEjo39OM:-NFT6pcuD8CvAu1XqL0U:b-lwk6x3) ([raw file](https://github.com/coreumfoundation/coreum/blob/56de3bc77f796606a09fbb20a63eb606bf466957/testutil/simapp/simapp.go#L45)):*

do we need to call `network.SetSDKConfig` here ?
</blockquote></details>

It's optional. but without it the "cosmos" prefixes will be used

PR #230, opened 2022-10-17
PoC of snapshotting

wojtek-coreum:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/230)


dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/230)* status: 0 of 30 files reviewed, 9 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/snapshot/v1/query.proto` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeFwli4_zipPWB8ZmH:-NEeFwli4_zipPWB8ZmI:b-pj5wec) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/proto/coreum/snapshot/v1/query.proto#L12)):*
> ```protobuf
> 
> service Query {
>   rpc PendingFreezeRequests(QueryPendingFreezeRequestsRequest) returns (QueryPendingFreezeRequestsResponse) {
> ```

Let's find another name not "Freeze" since it's already allocated by the Asset so we will be mixing it up.
___
*[`proto/coreum/snapshot/v1/types.proto` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeHctx-KH_Nvj5hgSZ:-NEeHctx-KH_Nvj5hgS_:b1fbyga) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/proto/coreum/snapshot/v1/types.proto#L12)):*
> ```protobuf
>   option (gogoproto.equal) = false;
>   string StoreName = 1;
>   bytes Name = 2;
> ```

I would go with the seq ID here. Since you shouldn't be create about the names, names can be the same.
___
*[`x/asset/keeper/keeper.go` line 107 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeLgvE7CdgGLhlrnZL:-NEeLgvE7CdgGLhlrnZM:b-y530p3) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/asset/keeper/keeper.go#L107)):*
> ```Go
> }
> 
> func (k Keeper) SnapshotFungibleToken(ctx sdk.Context, request types.FreezeRequestFungibleToken) error {
> ```

We need and example how to get the data from the stored key. Without it the full POC isn't clear for me. 
___
*[`x/asset/keeper/keeper.go` line 110 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeHvb09AeFMILG0of6:-NEeHvb09AeFMILG0of7:b-5faf99) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/asset/keeper/keeper.go#L110)):*
> ```Go
> 	// FIXME (wojtek): verify that denom exists
> 
> 	return k.snapshotKeeper.RequestFreeze(ctx, snapshottypes.FreezeRequest{
> ```

Does it mean that you will snapshot the entire bank?
___
*[`x/snapshot/ante/ante.go` line 30 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeLLJw8bzk0se0T8FO:-NEeLLJw8bzk0se0T8FP:brn7qb5) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/ante/ante.go#L30)):*
> ```Go
> // AnteHandle resets the gas limit inside GasMeter
> func (ald AddListenersDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
> 	return next(ctx.WithMultiStore(store.New(ctx.MultiStore(), ald.snapshotKey, ald.transformations)), tx, simulate)
> ```

Why do we do it in the ante?
___
*[`x/snapshot/keeper/keeper.go` line 42 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeIp6B3IOKGL71blEJ:-NEeIp6C9l7XNdRU3yhF:b-pi3us7) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/keeper/keeper.go#L42)):*
> ```Go
> 		return errors.WithStack(err)
> 	}
> 	store.Set(types.Join(types.FreezeRequestsKey, address.MustLengthPrefix(ownerAddress), indexMarshaled), must.Bytes(request.Marshal()))
> ```

Thing like this should live in the key.go file. To have just one place where we build the keys. Same for other keys.
___
*[`x/snapshot/keeper/keys.go` line 1 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeK8UlC7jSbCFc6Eqw:-NEeK8Um3qKqrKP_khlS:bowb0na) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/keeper/keys.go#L1)):*
> ```Go
> package keeper
> ```

Empty file. 
___
*[`x/snapshot/store/store.go` line 168 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeJw5GDIPOyXFB6rVX:-NEeJw5GDIPOyXFB6rVY:by73ujt) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/store/store.go#L168)):*
> ```Go
> }
> 
> func (s SnapshotIDStore) Claim(index sdk.Int, key []byte) error {
> ```

Some functions are not used so it's had to understand how they will work, we need a call to discuss it.
___
*[`x/snapshot/types/helpers.go` line 3 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeKH5X7Bmd8BAEvigE:-NEeKH5X7Bmd8BAEvigF:b-7a0mbm) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/types/helpers.go#L3)):*
> ```Go
> package types
> 
> func Join(keyComponents ...[]byte) []byte {
> ```

This join shoul be in the key.go file.

dzmitryhil (commented):
> Reviewed 30 of 30 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/230)* status: all files reviewed, 9 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

miladz68 (changes requested):
> Reviewed 28 of 30 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/230)* status: all files reviewed, 10 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`x/snapshot/keeper/keeper.go` line 46 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeY10a1nQmEnPYdcUv:-NEeY10a1nQmEnPYdcUw:blm8dsa) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/keeper/keeper.go#L46)):*
> ```Go
> }
> 
> func (k Keeper) Freeze(ctx sdk.Context) {
> ```

I think the critical part of implementing the snapshot can be expressed like this:

1. switch to new keys (you have done this in `Freeze` function)
2. resolve the correct balance at the current state and also at each snapshot when there are multiple snapshots from different accounts. 
3. remove entry from old snapshots and ensure that it correctly propogates to the next snapshots and ensure correct state is preserved. I think we can assume that the claim functionality will also remove entries and will be the only way to remove entries. We may even decide to limit the usage of snapshots to paying dividens with a fixed formula, but that requires a discussion.
4. think and find the edge cases where the implementation might misbehave and write tests for those edge cases. 

to be fair points 4 is too much for a POC, but part 2 and 3 should be implemented (at least partially) to give us a better sense about the complexity of this algorithm.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/230)* status: 9 of 32 files reviewed, 10 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`proto/coreum/snapshot/v1/query.proto` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeFwli4_zipPWB8ZmH:-NEixjIJ1LaGZfEzuEHX:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/proto/coreum/snapshot/v1/query.proto#L12)):*

Let's find another name not "Freeze" since it's already allocated by the Asset so we will be mixing it up.
</blockquote></details>

Done.
___
*[`proto/coreum/snapshot/v1/types.proto` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeHctx-KH_Nvj5hgSZ:-NEj8VT-9XwBPOaWNGlI:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/proto/coreum/snapshot/v1/types.proto#L12)):*

I would go with the seq ID here. Since you shouldn't be create about the names, names can be the same.
</blockquote></details>

Done.
___
*[`x/asset/keeper/keeper.go` line 110 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeHvb09AeFMILG0of6:-NEiy1946WeQlPuBvPzT:b-xa5r9g) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/asset/keeper/keeper.go#L110)):*

Does it mean that you will snapshot the entire bank?
</blockquote></details>

What to snapshot is decided by the "transformation". It's kind of driver which tells the `snapshot` module what and how to snapshot.
Here is an example of such "transformation" for fungible tokens: https://github.com/CoreumFoundation/coreum/blob/wojtek/snapshots-poc/x/asset/types/snapshot.go#L39
___
*[`x/snapshot/ante/ante.go` line 30 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeLLJw8bzk0se0T8FO:-NEiyJsp2Vd6F2LyMBuI:bqhi9wz) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/ante/ante.go#L30)):*

Why do we do it in the ante?
</blockquote></details>

Each transaction receives it's own cache multi store (to be able to ignore changes if transaction fails). It means that we need to create our layer on top of that multi store. Ante handler is the only place where we can do it globally for every tx type.
___
*[`x/snapshot/keeper/keeper.go` line 42 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeIp6B3IOKGL71blEJ:-NEj48ct7ToXRN4AUtnA:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/keeper/keeper.go#L42)):*

Thing like this should live in the key.go file. To have just one place where we build the keys. Same for other keys.
</blockquote></details>

Done.
___
*[`x/snapshot/keeper/keeper.go` line 46 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeY10a1nQmEnPYdcUv:-NEj50EfBlSDZI2QLCPE:b4fei7) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/keeper/keeper.go#L46)):*

I think the critical part of implementing the snapshot can be expressed like this:

1. switch to new keys (you have done this in `Freeze` function)
2. resolve the correct balance at the current state and also at each snapshot when there are multiple snapshots from different accounts. 
3. remove entry from old snapshots and ensure that it correctly propogates to the next snapshots and ensure correct state is preserved. I think we can assume that the claim functionality will also remove entries and will be the only way to remove entries. We may even decide to limit the usage of snapshots to paying dividens with a fixed formula, but that requires a discussion.
4. think and find the edge cases where the implementation might misbehave and write tests for those edge cases. 

to be fair points 4 is too much for a POC, but part 2 and 3 should be implemented (at least partially) to give us a better sense about the complexity of this algorithm.
</blockquote></details>

Yes, you pretty accurately summarized the algorithm
___
*[`x/snapshot/keeper/keys.go` line 1 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeK8UlC7jSbCFc6Eqw:-NEj49pb9j__Fi0o5AJv:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/keeper/keys.go#L1)):*

Empty file. 
</blockquote></details>

Done.
___
*[`x/snapshot/types/helpers.go` line 3 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/230#-NEeKH5X7Bmd8BAEvigE:-NEj4KQpCfop9VTiNmT0:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/9e07f693f15ff3d451b69f5f250bfee0e5c283e4/x/snapshot/types/helpers.go#L3)):*

This join shoul be in the key.go file.
</blockquote></details>

Done.

miladz68 (commented):
> Reviewed 1 of 30 files at r1, 1 of 1 files at r2, 21 of 22 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/230)* status: all files reviewed, 10 unresolved discussions (waiting on @dzmitryhil, @wojtek-coreum, and @ysv)

miladz68 (commented):
> Reviewed 2 of 2 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/230)* status: all files reviewed, 10 unresolved discussions (waiting on @dzmitryhil, @wojtek-coreum, and @ysv)

PR #236, opened 2022-10-20
Remove ComputeNeededBalance helper

dzmitryhil:
* Squash test TestTransferFailsIfNotEnoughGasIsProvided into TestTransferDeterministicGas
* Optimize tests to use auto gas where possible
* Replace the usage of the chain InitialGasPrice price with favor to tx.GetGasPrice

- - -
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/236)


wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/236)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/auth/gas.go` line 22 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp-3Kx6pPGVz3rpTgH:-NEp-3Kx6pPGVz3rpTgI:b-5kts7z) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/gas.go#L22)):*
> ```Go
> 
> 	maxBlockGas := chain.NetworkConfig.Fee.FeeModel.Params().MaxBlockGas
> 	require.NoError(t, chain.Faucet.FundAccounts(ctx, testing.NewFundedAccount(sender, chain.NewCoin(testing.ComputeNeededBalance(
> ```

Balance was added here intentionally despite the fact it should not be consumed. Otherwise we might be in condition where tx fails as expected due to lack of balance, not because of the state we are really checking.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/236)* status: 0 of 19 files reviewed, 2 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp0_8k4l1FWp4fWLYN:-NEp0_8lCGmoPtX0jpD1:b-u7gdiu) (no related file):*
Depending on `tx,GetGasPirce` is not a good idea because that price might go up and down at any moment, meaning that we might start receiving random test failures caused by too low fee offered.

___

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/236)* status: 0 of 19 files reviewed, 3 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/asset/issue.go` line 25 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp0xliE674EOEMg_6T:-NEp0xlj1mkjqQLr-cF1:b-b7uzhu) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/asset/issue.go#L25)):*
> ```Go
> 	issuer := chain.GenAccount()
> 	recipient := chain.GenAccount()
> 	requireT.NoError(chain.Faucet.FundAccounts(ctx,
> ```

How can tx succeed if issuer does not hold balance to cover fees?

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/236)* status: 0 of 19 files reviewed, 5 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/auth/multisig.go` line 59 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1NL76e9Nr7pHmCIV:-NEp1NL82s0Mi0pqdMra:b-a7xy5w) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/multisig.go#L59)):*
> ```Go
> 		testing.NewFundedAccount(
> 			multisigAddress,
> 			chain.NewCoin(testing.ComputeNeededBalance(
> ```

Why? What if we change gas amount for bank-send in the future? Someone will have to scan the tests, find the cause and fix it. Now it adjusts automatically.
___
*[`integration-tests/auth/sequence.go` line 20 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1bpD9XC1_eT_ACsR:-NEp1bpD9XC1_eT_ACsS:bm6toq6) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/sequence.go#L20)):*
> ```Go
> 	sender := chain.GenAccount()
> 
> 	require.NoError(t, chain.Faucet.FundAccounts(ctx,
> ```

same thing here, if user does not hold balance we don't know if the tx failed for the exact reason we expect
___
*[`integration-tests/testing/chain.go` line 148 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1xHJ-nUic59nlAkp:-NEp1xHJ-nUic59nlAkq:b2zcklz) ([raw file](https://github.com/coreumfoundation/coreum/blob/9db09a2dc420bee23adcb195df5a9dbda4c22980/integration-tests/testing/chain.go#L148)):*
> ```Go
> 
> // GenFundedAccount generates a new account in the chain and fund it with the predefined balance
> func (c Chain) GenFundedAccount(ctx context.Context) (sdk.AccAddress, error) {
> ```

I disagree with this solution, I prefer to strictly assign balance required for tests as only in that case we really test the tx

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/236)* status: 0 of 19 files reviewed, 5 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp0_8k4l1FWp4fWLYN:-NEp2wtC13XINQW4KczO:btu9fk0) (no related file):*

Depending on `tx,GetGasPirce` is not a good idea because that price might go up and down at any moment, meaning that we might start receiving random test failures caused by too low fee offered.
</blockquote></details>

Do you prefer to use the InitialGasPrice? Can we overcome it in the tests?

___
___
*[`integration-tests/asset/issue.go` line 25 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp0xliE674EOEMg_6T:-NEp3EQ11LgF5EHn2b0n:b-obb6sv) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/asset/issue.go#L25)):*

How can tx succeed if issuer does not hold balance to cover fees?
</blockquote></details>

We send from the "issuer", the recipient should have the coins.
___
*[`integration-tests/auth/gas.go` line 22 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp-3Kx6pPGVz3rpTgH:-NEp3OCZ9prQ4k7YQmv5:bvn9sit) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/gas.go#L22)):*

Balance was added here intentionally despite the fact it should not be consumed. Otherwise we might be in condition where tx fails as expected due to lack of balance, not because of the state we are really checking.
</blockquote></details>

This condition is almost impossible. 
___
*[`integration-tests/auth/multisig.go` line 59 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1NL76e9Nr7pHmCIV:-NEp3RSc1rancwZ1PhC0:b-i3rg9x) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/multisig.go#L59)):*

Why? What if we change gas amount for bank-send in the future? Someone will have to scan the tests, find the cause and fix it. Now it adjusts automatically.
</blockquote></details>

The 1_000_000 is just for example here. We can use the same constant as in GenFundedAccount, if you prefere. 
___
*[`integration-tests/auth/sequence.go` line 20 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1bpD9XC1_eT_ACsR:-NEp3fbTBTzqPbkwvlEe:b-u5tv1b) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/sequence.go#L20)):*

same thing here, if user does not hold balance we don't know if the tx failed for the exact reason we expect
</blockquote></details>

I didn't get you.
___
*[`integration-tests/testing/chain.go` line 148 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1xHJ-nUic59nlAkp:-NEp3vb75lkek-7eHqW0:bds74q5) ([raw file](https://github.com/coreumfoundation/coreum/blob/9db09a2dc420bee23adcb195df5a9dbda4c22980/integration-tests/testing/chain.go#L148)):*

I disagree with this solution, I prefer to strictly assign balance required for tests as only in that case we really test the tx
</blockquote></details>

The PR is my understanding of the subtask:

"ComputeNeededBalance could possibly be simplified See: FIXME (wojtek): hardcode reasonable default values. But it could also use Estimage gas so calculations are simpler"

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/236)* status: 0 of 19 files reviewed, 5 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp0_8k4l1FWp4fWLYN:-NEp79KG57M-H3QB3z7o:b-6hd3a9) (no related file):*

Do you prefer to use the InitialGasPrice? Can we overcome it in the tests?
</blockquote></details>

Yes, that's why InitialGasPrice was used.

___
___
*[`integration-tests/auth/gas.go` line 22 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp-3Kx6pPGVz3rpTgH:-NEp6AEgE8winfSPHLC8:b-fm3hsj) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/gas.go#L22)):*

This condition is almost impossible. 
</blockquote></details>

"almost" is the keyword
___
*[`integration-tests/auth/multisig.go` line 59 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1NL76e9Nr7pHmCIV:-NEp6JS2Ca_86xqdq4dF:bc4hcu9) ([raw file](https://github.com/coreumfoundation/coreum/blob/7641bb8ae55d561df97c80c3678702cd10a1dd91/integration-tests/auth/multisig.go#L59)):*

The 1_000_000 is just for example here. We can use the same constant as in GenFundedAccount, if you prefere. 
</blockquote></details>

It doesn't matter. What matters is that you hardcode constants in places which might be affected by changing the parameter of the network. And then someone must iterate over all the tests and update constants.
___
*[`integration-tests/testing/chain.go` line 148 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/236#-NEp1xHJ-nUic59nlAkp:-NEp6aQD0t4-PizaRbk_:b-waz7hq) ([raw file](https://github.com/coreumfoundation/coreum/blob/9db09a2dc420bee23adcb195df5a9dbda4c22980/integration-tests/testing/chain.go#L148)):*

The PR is my understanding of the subtask:

"ComputeNeededBalance could possibly be simplified See: FIXME (wojtek): hardcode reasonable default values. But it could also use Estimage gas so calculations are simpler"


</blockquote></details>

Okay, so you misunderstood it :-). Let's discuss it after the daily. This task should go to me because only I discussed it with @ysv . But IIRC task description contained some code example.

PR #242, opened 2022-10-25
Post refactoring - asset basic issuance

dzmitryhil:
* Refactor FindTypedEvent to use generics
* Refactor the order of the FT issuance CLI args

- - -
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)


ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDOuljCHOIvA_NFw7b:-NFDOuljCHOIvA_NFw7c:b-9db7r6) (no related file):*
Continuing discussion here


___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDRtEqAyrN4k2LhpXV:-NFDRtEqAyrN4k2LhpXW:bheo3la) (no related file):*
continuing discussion here:


I see. So we are limited to this regexp for denom `^[a-zA-Z][a-zA-Z0-9/-]{2,127}$` which means that symbol should be subset of this. And we can say that `-` is sort of reserved by us for combining symbol + issuer + checksum

Ok we need to work on this but as a part of another task I think

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDUzJn0rnQoCXKCvl7:-NFDUzJn0rnQoCXKCvl8:b-m7c65u) (no related file):*


I see this in our code
```
if msg.InitialAmount.IsNil() || msg.InitialAmount.IsNegative() {
  return sdkerrors.Wrapf(ErrInvalidFungibleToken, "invalid initial amount %s, can't be negative", msg.InitialAmount.String())
}
```
It will probably fail only for negative and should be changed to `... || !msg.InitialAmount.IsPositive()`

It is also worth adding a test for this 

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDVXdACcUziwVucS61:-NFDVXdACcUziwVucS62:b-pk8kw5) (no related file):*


an example from cosmos: https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-024-coin-metadata.md
```
{
  "description": "The native staking token of the Cosmos Hub.",
  "denom_units": [
    {
      "denom": "uatom",
      "exponent": 0,
      "aliases": [
        "microatom"
      ],
    },
    {
      "denom": "matom",
      "exponent": 3,
      "aliases": [
        "milliatom"
      ]
    },
    {
      "denom": "atom",
      "exponent": 6,
    }
  ],
  "base": "uatom",
  "display": "atom",
}
```
we use `display: atom` not `uatom`

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDW-mvExqCRgTkFOj9:-NFDW-mvExqCRgTkFOjA:b-9mymfo) (no related file):*


To be discussed. I probably don't understand it fully.
It might also be inconvenient for users of this endpoint

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDW_zw383xtpRoqreG:-NFDW_zw383xtpRoqreH:b-6zf9h9) (no related file):*


___
___
*[`x/asset/client/cli/tx.go` line 78 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDR3d-4vPvzQXs1qtz:-NFDR3d-4vPvzQXs1qu-:b-qvui60) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/x/asset/client/cli/tx.go#L78)):*
> ```Go
> 			}
> 
> 			description := args[3]
> ```

Ideally I would order fields in all structures related to issuance according to this logic

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDOuljCHOIvA_NFw7b:-NFDbST27Y9y5fZDj6XY:b8gcye0) (no related file):*

Continuing discussion here


</blockquote></details>

Maybe FungibleTokenBase for the first struct 
anyways lets discuss on the call

___

miladz68 (commented):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil and @wojtek-coreum)

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 7 unresolved discussions (waiting on @wojtek-coreum and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDUzJn0rnQoCXKCvl7:-NFDnjhm8dxpxHvNUkSE:b-gbf3s4) (no related file):*



I see this in our code
```
if msg.InitialAmount.IsNil() || msg.InitialAmount.IsNegative() {
  return sdkerrors.Wrapf(ErrInvalidFungibleToken, "invalid initial amount %s, can't be negative", msg.InitialAmount.String())
}
```
It will probably fail only for negative and should be changed to `... || !msg.InitialAmount.IsPositive()`

It is also worth adding a test for this 
</blockquote></details>

We allow zero value. And we have test for it.  

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDVXdACcUziwVucS61:-NFDnu0F5Z4ACgruSQrD:b-hvvbn2) (no related file):*



an example from cosmos: https://github.com/cosmos/cosmos-sdk/blob/main/docs/architecture/adr-024-coin-metadata.md
```
{
  "description": "The native staking token of the Cosmos Hub.",
  "denom_units": [
    {
      "denom": "uatom",
      "exponent": 0,
      "aliases": [
        "microatom"
      ],
    },
    {
      "denom": "matom",
      "exponent": 3,
      "aliases": [
        "milliatom"
      ]
    },
    {
      "denom": "atom",
      "exponent": 6,
    }
  ],
  "base": "uatom",
  "display": "atom",
}
```
we use `display: atom` not `uatom`
</blockquote></details>

We have just one denom in the denom_unit now. So we can't set similar to that example. 

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDW-mvExqCRgTkFOj9:-NFDo32Q8vkV4ge4MzNa:b6ugaqp) (no related file):*



To be discussed. I probably don't understand it fully.
It might also be inconvenient for users of this endpoint
</blockquote></details>

Check the cosmos-module transactions. The queries are with the responses the transactions are with empty responses. 

___

wojtek-coreum (changes requested):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 9 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`integration-tests/asset/issue.go` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFIN7lu7bxpfQcyiY1f:-NFIN7lu7bxpfQcyiY1g:bfjs2qe) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/integration-tests/asset/issue.go#L12)):*
> ```Go
> 	"github.com/CoreumFoundation/coreum/integration-tests/testing"
> 	"github.com/CoreumFoundation/coreum/pkg/tx"
> 	"github.com/CoreumFoundation/coreum/testutil/event"
> ```

Could we keep it in integration-tests/testing/event?

I thnk we don't need another scope (testutil) for test helpers
___
*[`testutil/event/event.go` line 25 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFINnQL4zo3s414n4dP:-NFINnQMAhn0I6eufXMC:b-vqcx4n) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/testutil/event/event.go#L25)):*
> ```Go
> 		}
> 
> 		return msg.(T), nil
> ```

could we return an error if cast is not possible instead of panicking?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 9 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`integration-tests/asset/issue.go` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFIN7lu7bxpfQcyiY1f:-NFJJHIaEHi0N29TgYKH:b-crzxpo) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/integration-tests/asset/issue.go#L12)):*

Could we keep it in integration-tests/testing/event?

I thnk we don't need another scope (testutil) for test helpers
</blockquote></details>

We have shared func to get events and it's used in the integration tests and in simapp tests. Where would you prefer to keep it?
___
*[`x/asset/client/cli/tx.go` line 78 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDR3d-4vPvzQXs1qtz:-NFJKMkmBzM7zj7w9q07:bsy30wk) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/x/asset/client/cli/tx.go#L78)):*

Ideally I would order fields in all structures related to issuance according to this logic 
</blockquote></details>

Let's discuss the points above and I'll do it.



_Code quote:_
```Go
description := args[3]
```
___
*[`testutil/event/event.go` line 25 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFINnQL4zo3s414n4dP:-NFJKB8X9mFziHIBPGH0:b-iaftvq) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/testutil/event/event.go#L25)):*

could we return an error if cast is not possible instead of panicking?
</blockquote></details>

Agree, and updated it.



_Code quote:_
```Go
return msg.(T), nil
```

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 8 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`integration-tests/asset/issue.go` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFIN7lu7bxpfQcyiY1f:-NFN-eSr4QqeQJ8uaU6l:b-pmu2y2) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/integration-tests/asset/issue.go#L12)):*

We have shared func to get events and it's used in the integration tests and in simapp tests. Where would you prefer to keep it?
</blockquote></details>

I see, so maybe we should move `testing` package from `integration-tests`. But we might come back to this in the future if we see that more stuff is shared between integration tests and simapp tests. Resolving for now.
___
*[`testutil/event/event.go` line 27 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFMzrsH4eOIS9jCxoik:-NFMzrsH4eOIS9jCxoil:b-77x0re) ([raw file](https://github.com/coreumfoundation/coreum/blob/2a1f9a519dae2e9ff2920435277b865138be73fa/testutil/event/event.go#L27)):*
> ```Go
> 		typedMsg, ok := msg.(T)
> 		if !ok {
> 			return *new(T), errors.Errorf("can't cast found event to %+v", event) //nolint:gocritic // T(nil) doesn't work with the proto.Message
> ```

maybe %T instead of %+v?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: 6 of 30 files reviewed, 3 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDUzJn0rnQoCXKCvl7:-NFTqZ7GA34VCr1U4KPi:bax7kej) (no related file):*

lets just add comment why we allow minitng with 0 supply
</blockquote></details>

Added the comment

___
*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDW_zw383xtpRoqreG:-NFTqbRH5XlqC15mCLA-:bx7qtzj) (no related file):*
> Added the comment
Added the comment

___
___
*[`integration-tests/asset/issue.go` line 12 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFIN7lu7bxpfQcyiY1f:-NFTqmdI30ekLWkMEmsm:bz08ru4) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/integration-tests/asset/issue.go#L12)):*

I see, so maybe we should move `testing` package from `integration-tests`. But we might come back to this in the future if we see that more stuff is shared between integration tests and simapp tests. Resolving for now.
</blockquote></details>

I would keep it like now. The "testing" in the "integration-tests" is for the integration tests only, and reusable code can be moved to testutils a I did.
___
*[`x/asset/client/cli/tx.go` line 78 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDR3d-4vPvzQXs1qtz:-NFTr9041GAp8gx2aYhC:btl1r2k) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/x/asset/client/cli/tx.go#L78)):*

So ideally I would sort the fields in structs according to their importance
But if this requires a lot of time we can postpone it
</blockquote></details>

Sorted.
___
*[`testutil/event/event.go` line 27 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFMzrsH4eOIS9jCxoik:-NFTr5n61fZ8YrtIxuSM:b-bprc3) ([raw file](https://github.com/coreumfoundation/coreum/blob/2a1f9a519dae2e9ff2920435277b865138be73fa/testutil/event/event.go#L27)):*

maybe %T instead of %+v?
</blockquote></details>

Agree, done



_Code quote:_
```Go
errors.Errorf("can't cast found event to
```

ysv (changes requested):
> Reviewed 2 of 7 files at r1, 25 of 25 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 2 unresolved discussions (waiting on @dzmitryhil and @wojtek-coreum)

___
*[`x/asset/client/cli/tx.go` line 78 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFDR3d-4vPvzQXs1qtz:-NFjD3jfDJBQY2m9pevl:b2niddr) ([raw file](https://github.com/coreumfoundation/coreum/blob/a97f1a520c7a4e545e9d64f366d25b8c70cfb5c2/x/asset/client/cli/tx.go#L78)):*

Sorted.
</blockquote></details>

txh
___
*[`testutil/event/event.go` line 13 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFjBPTB1WOD8Yit6WS1:-NFjBPTE0Mv5H-lM8nHK:b-8tq59u) ([raw file](https://github.com/coreumfoundation/coreum/blob/fb43b79c806fba7d9c88254e2286b6e760561dc7/testutil/event/event.go#L13)):*
> ```Go
> 
> // FindTypedEvent finds the event in the list of events, and set marshals it the to the event.
> func FindTypedEvent[T proto.Message](event T, events []tmtypes.Event) (T, error) {
> ```

What I don't like in your solution is that we have this argument `event T` which is not needed and added just to determine type

I've tried myself to solve similar problem using generics and found this solution to be more sophisticated and natural

```
package main

import (
	"errors"
	"fmt"
)

func FindSmthByType[T any](slice []any) (T, error) {
	for _, el := range slice {
		typedEl, ok := el.(T)
		if !ok {
			continue
		}

		return typedEl, nil
	}

	return *new(T), errors.New("not found")
}

func main() {
	a1 := []interface{}{1, "string1", []byte("bytes"), 1.2}

	el, err := FindSmthByType[int](a1)
	if err != nil {
		fmt.Printf("err: %v", err)
	} else {
		fmt.Printf("%T: %v", el, el)
	}
}
```
I propose to proceed with similar approach.
I think it is better because we don't really need to pass an argument when we need only type

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: 27 of 30 files reviewed, 2 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`testutil/event/event.go` line 13 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFjBPTB1WOD8Yit6WS1:-NFmRSja1jy1t2_YCw0T:b-duhpis) ([raw file](https://github.com/coreumfoundation/coreum/blob/fb43b79c806fba7d9c88254e2286b6e760561dc7/testutil/event/event.go#L13)):*

What I don't like in your solution is that we have this argument `event T` which is not needed and added just to determine type

I've tried myself to solve similar problem using generics and found this solution to be more sophisticated and natural

```
package main

import (
	"errors"
	"fmt"
)

func FindSmthByType[T any](slice []any) (T, error) {
	for _, el := range slice {
		typedEl, ok := el.(T)
		if !ok {
			continue
		}

		return typedEl, nil
	}

	return *new(T), errors.New("not found")
}

func main() {
	a1 := []interface{}{1, "string1", []byte("bytes"), 1.2}

	el, err := FindSmthByType[int](a1)
	if err != nil {
		fmt.Printf("err: %v", err)
	} else {
		fmt.Printf("%T: %v", el, el)
	}
}
```
I propose to proceed with similar approach.
I think it is better because we don't really need to pass an argument when we need only type
</blockquote></details>

Agree. Updated.

ysv (dismissed):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 1 unresolved discussion (waiting on @wojtek-coreum)

wojtek-coreum (changes requested):
> Reviewed 24 of 25 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil)

___
*[`pkg/config/genesis/genesis.tmpl.json` line 67 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFrmJ8Y5gfvNJ2eagKj:-NFrmJ8Y5gfvNJ2eagKk:bkq6f5v) ([raw file](https://github.com/coreumfoundation/coreum/blob/5354ad88117c5ece0c27ccf2d136a00657d2248b/pkg/config/genesis/genesis.tmpl.json#L67)):*
> ```JSON
>           "display": "{{ .MetadataDisplayDenom }}",
>           "name": "{{ .Denom }}",
>           "symbol": "{{ .Denom | ToUpper }}"
> ```

Why `ToUpper`?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: all files reviewed, 1 unresolved discussion (waiting on @wojtek-coreum)

___
*[`pkg/config/genesis/genesis.tmpl.json` line 67 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/242#-NFrmJ8Y5gfvNJ2eagKj:-NFrwBPJA68Mv6iJvwKG:b-h2nkb9) ([raw file](https://github.com/coreumfoundation/coreum/blob/5354ad88117c5ece0c27ccf2d136a00657d2248b/pkg/config/genesis/genesis.tmpl.json#L67)):*

Why `ToUpper`?
</blockquote></details>

The symbol is usually used in upper case.

wojtek-coreum (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dzmitryhil)

wojtek-coreum (approved):
> Reviewed 12 of 12 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/242)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68 and @ysv)

PR #268, opened 2022-11-22
Implement asset precision

miladz68:

PR #271, opened 2022-11-24
Global freeze implementation

ysv:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)


ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: 0 of 14 files reviewed, 2 unresolved discussions

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHf25VN7Dp_Hv_A8Zq4:-NHf25VN7Dp_Hv_A8Zq5:b5c2qbg) (no related file):*
I have a few topics to discuss:

1.
How to we store global freeze:

Option 1: in FungibleTokenDefinition
pros:
- Best case 1 read of FungibleTokn. Worst case 2 reads: FT & frozen balance. 
In most cases we can just check if freeze feature enabled inside ft and if it is not we don't need additional reads to check frozen balance for user.
- no additional key needed to just store a single true/false
- As for me it is logical to return if global freeze enabled in `cored q asset ft info` query

cons:
- on each tx including bank send we should read full FungibleTokenDefinition

Option 2: under a separate key
oposite to FungibleToken

___
___
*[`x/asset/keeper/global_freeze.go` line 16 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHf4xV321Mma3eDYupY:-NHf4xV321Mma3eDYupZ:b-p47u1x) ([raw file](https://github.com/coreumfoundation/coreum/blob/083b7e51e90a74ed97fc60044bb5fe7619eaae0d/x/asset/keeper/global_freeze.go#L16)):*
> ```Go
> 	}
> 
> 	// FIXME: Fungible token is read from store twice.
> ```

IMO `checkFeatureAllowed` should be a method of `FungibleTokenDefinition ` and a typical flow should look like this:

```
func BurnFungibleToken() {
 // read from store
 ft, err := k.GetFungibleTokenDefinition(ctx, denom)
 ...
 // check in memore
 ft.checkFeatureAllowed()
}
```

if everyone is ok I will refactor to this version everywhere

miladz68 (commented):
> Reviewed 11 of 14 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 6 unresolved discussions (waiting on @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHf25VN7Dp_Hv_A8Zq4:-NHi6VJtEbk8b6EDUxLQ:b-krmsk9) (no related file):*

I have a few topics to discuss:

1.
How to we store global freeze:

Option 1: in FungibleTokenDefinition
pros:
- Best case 1 read of FungibleTokn. Worst case 2 reads: FT & frozen balance. 
In most cases we can just check if freeze feature enabled inside ft and if it is not we don't need additional reads to check frozen balance for user.
- no additional key needed to just store a single true/false
- As for me it is logical to return if global freeze enabled in `cored q asset ft info` query

cons:
- on each tx including bank send we should read full FungibleTokenDefinition

Option 2: under a separate key
oposite to FungibleToken
</blockquote></details>

to me definition is an immutable structure and modifying it makes me uncomfortable. we can have the mutable configuration stored separately and we can store acl under the same key. 

___
___
*[`proto/coreum/asset/v1/tx.proto` line 24 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhz0EMDHXHcPYYZYAf:-NHhz0EMDHXHcPYYZYAg:b-6rs60k) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L24)):*
> ```protobuf
>   // BurnFungibleToken burns the specified fungible tokens from senders balance if the sender has enough balance
>   rpc BurnFungibleToken(MsgBurnFungibleToken) returns (EmptyResponse);
> ```

I see you are removing the comments, I like this approach. It will be better if you remove previous comments and also remove the empty lines between methods. 
___
*[`proto/coreum/asset/v1/tx.proto` line 32 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhuK1839psJcb-Qhnb:-NHhuK1839psJcb-Qhnc:b-tksof4) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L32)):*
> ```protobuf
> // MsgIssueFungibleToken defines message to issue new fungible token.
> message MsgIssueFungibleToken {
>   option (gogoproto.equal) = false; // Why do we need this ?
> ```

this tag adds Equal method to the type. I think the default value for it is false, and removing it is the same as setting it to false. 
___
*[`x/asset/handler.go` line 14 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhzn6t514GV4uhGPOo:-NHhzn6t514GV4uhGPOp:b-yy3tou) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/handler.go#L14)):*
> ```Go
> 
> // NewHandler return tx handler of the asset module.
> func NewHandler(ms keeper.MsgServer) sdk.Handler {
> ```

as dicsussed in the slack, this handler is legacy stuff and should be completely removed. 
___
*[`x/asset/client/cli/tx.go` line 323 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHi0FML6jDY7XpzwdAs:-NHi0FML6jDY7XpzwdAt:b52sia7) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/client/cli/tx.go#L323)):*
> ```Go
> func CmdTxGlobalFreezeFungibleToken() *cobra.Command {
> 	cmd := &cobra.Command{
> 		Use:   "global-freeze [denom] --from [sender]",
> ```

just an immature idea, we can combine the freeze and global freeze, you can do `freeze [address]` of `freeze global`
___
*[`x/asset/keeper/global_freeze.go` line 16 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHf4xV321Mma3eDYupY:-NHi7T0B2pkPH8nrVCF1:b757pjs) ([raw file](https://github.com/coreumfoundation/coreum/blob/083b7e51e90a74ed97fc60044bb5fe7619eaae0d/x/asset/keeper/global_freeze.go#L16)):*

IMO `checkFeatureAllowed` should be a method of `FungibleTokenDefinition ` and a typical flow should look like this:

```
func BurnFungibleToken() {
 // read from store
 ft, err := k.GetFungibleTokenDefinition(ctx, denom)
 ...
 // check in memore
 ft.checkFeatureAllowed()
}
```

if everyone is ok I will refactor to this version everywhere
</blockquote></details>

I like it

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 17 unresolved discussions (waiting on @ysv)

___
*[`integration-tests/asset/global_freeze.go` line 19 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiY_Gu2ckUoDZWg9dD:-NHiY_Gu2ckUoDZWg9dE:b-r7nj5w) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/integration-tests/asset/global_freeze.go#L19)):*
> ```Go
> func TestGlobalFreezeFungibleToken(ctx context.Context, t testing.T, chain testing.Chain) {
> 	requireT := require.New(t)
> 	//assertT := assert.New(t)
> ```

Is it draft why commented?
___
*[`integration-tests/asset/global_freeze.go` line 75 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiYfz-6t3MvytWnDPT:-NHiYfz-6t3MvytWnDPU:b-uoz9b0) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/integration-tests/asset/global_freeze.go#L75)):*
> ```Go
> 
> 	requireT.NoError(err)
> 	fungibleTokenIssuedEvt, err := event.FindTypedEvent[*assettypes.EventFungibleTokenIssued](res.Events)
> ```

I guess you don't need to check it here, since we do it in other tests. 
___
*[`proto/coreum/asset/v1/asset.proto` line 15 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHid3xUCBW9o8qSFadg:-NHid3xUCBW9o8qSFadh:b8o7yif) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/asset.proto#L15)):*
> ```protobuf
> 
> // FungibleTokenDefinition defines the fungible token settings to store.
> message FungibleTokenDefinition {
> ```

Don't forget to add `global_freeze_enabled` to the `FungibleToken` and fill it in the query.
___
*[`proto/coreum/asset/v1/asset.proto` line 21 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiYtBl8R1nvgPqzbLM:-NHiYtBl8R1nvgPqzbLN:bz2f7uz) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/asset.proto#L21)):*
> ```protobuf
>   string issuer = 2;
>   repeated FungibleTokenFeature features = 3;
>   bool global_freeze_enabled = 4;
> ```

I would keep it separately,  to keep the design with the features alongside the definition. So you can add it to the query for the FungibleToken but don't add to the definition.  
___
*[`proto/coreum/asset/v1/tx.proto` line 27 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiZY9s3C-GRNsxNpul:-NHiZY9s3C-GRNsxNpum:b-3kp0q3) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L27)):*
> ```protobuf
>   rpc GlobalFreezeFungibleToken(MsgGlobalFreezeFungibleToken) returns (EmptyResponse);
> 
>   rpc GlobalUnfreezeFungibleToken(MsgGlobalUnfreezeFungibleToken) returns (EmptyResponse);
> ```

Don't forget to add coments. 
___
*[`x/asset/client/cli/tx.go` line 325 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiaNyn--9n5z0KBdE7:-NHiaNyn--9n5z0KBdE8:b9sw7fv) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/client/cli/tx.go#L325)):*
> ```Go
> 		Use:   "global-freeze [denom] --from [sender]",
> 		Args:  cobra.ExactArgs(1),
> 		Short: "Globally freeze fungible token on all accounts",
> ```

Not sure about the part `on all accounts`. It just globally freezes, without touching the accounts. (some for unfreeze)
___
*[`x/asset/keeper/global_freeze.go` line 10 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHictpa6v4k2r9xg8ek:-NHictpa6v4k2r9xg8el:b-odsgm0) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L10)):*
> ```Go
> )
> 
> func (k Keeper) SetGlobalFreezeEnabled(ctx sdk.Context, sender sdk.AccAddress, denom string, enabled bool) error {
> ```

Where is the logic with the checking of the globally frozen token?
___
*[`x/asset/keeper/global_freeze.go` line 16 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHib9uD4rM58fBN3U3H:-NHib9uD4rM58fBN3U3I:b-6t7e4b) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L16)):*
> ```Go
> 	}
> 
> 	// FIXME: Fungible token is read from store twice.
> ```

As I mention in the proto IMHO we should add that flag to the definition.  
___
*[`x/asset/keeper/global_freeze.go` line 23 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHibP7U5ENDql4cRPOa:-NHibP7U5ENDql4cRPOb:b9upe66) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L23)):*
> ```Go
> 	}
> 
> 	if ft.GlobalFreezeEnabled == enabled {
> ```

Why we should return the error? It's an idempotent operation, so it is safer to allow to set the same twice here. 
___
*[`x/asset/keeper/global_freeze.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHibHeLEvtDNCjjGfyN:-NHibHeLEvtDNCjjGfyO:b62s1bv) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L26)):*
> ```Go
> 		if enabled {
> 			return sdkerr.Wrap(types.ErrGlobalFreezeEnabled, "already enabled")
> 		} else {
> ```

No need to `else` here you return in `if`
___
*[`x/asset/types/msgs.go` line 141 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHicOmx5jLCM8K-jO7R:-NHicOmx5jLCM8K-jO7S:b9r429m) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/types/msgs.go#L141)):*
> ```Go
> 	}
> 
> 	if _, _, err := DeconstructFungibleTokenDenom(msg.Denom); err != nil {
> ```

Don't understand why you need it (same for next message). Will you `Deconstruct` later in the code?

dzmitryhil (commented):
> Reviewed 11 of 14 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 17 unresolved discussions (waiting on @ysv)

miladz68 (changes requested):
> Reviewed 12 of 12 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 17 unresolved discussions (waiting on @ysv)

___
*[`x/asset/types/keys.go` line 45 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI2CCzHCx_tlDx-qLtK:-NI2CCzHCx_tlDx-qLtL:b-5ahmgp) ([raw file](https://github.com/coreumfoundation/coreum/blob/fd8decfe0d49553983269c37e1be3c49e0f9210a/x/asset/types/keys.go#L45)):*
> ```Go
> 
> func CreateGlobalFreezePrefix(denom string) []byte {
> 	return store.JoinKeysWithLength(GlobalFreezeKeyPrefix, []byte(denom))
> ```

we don't need a prefix store for each denom. just a single prefix store where denom is the key and presence of that key indicates that denom is frozen.

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: 8 of 21 files reviewed, 14 unresolved discussions (waiting on @dzmitryhil and @miladz68)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHf25VN7Dp_Hv_A8Zq4:-NHySzyxBAHjx1QHFOJ5:b-2gq4x3) (no related file):*

to me definition is an immutable structure and modifying it makes me uncomfortable. we can have the mutable configuration stored separately and we can store acl under the same key. 
</blockquote></details>

agree refactored to store it separately 

___
___
*[`integration-tests/asset/global_freeze.go` line 19 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiY_Gu2ckUoDZWg9dD:-NHyT1jG98J8YseO-E4X:b-3v3y9n) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/integration-tests/asset/global_freeze.go#L19)):*

Is it draft why commented?
</blockquote></details>

draft )
___
*[`integration-tests/asset/global_freeze.go` line 75 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiYfz-6t3MvytWnDPT:-NHyT3k9CJ61iAjE9nJ7:b-21nrw) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/integration-tests/asset/global_freeze.go#L75)):*

I guess you don't need to check it here, since we do it in other tests. 
</blockquote></details>

it is still a draft
___
*[`proto/coreum/asset/v1/asset.proto` line 15 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHid3xUCBW9o8qSFadg:-NHyUiOQ753LOyWQgO9X:b7h7a75) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/asset.proto#L15)):*

Don't forget to add `global_freeze_enabled` to the `FungibleToken` and fill it in the query.
</blockquote></details>

exactly thx
done
___
*[`proto/coreum/asset/v1/asset.proto` line 21 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiYtBl8R1nvgPqzbLM:-NHyUkRm95wmUd0eHAUO:b-lo6lz9) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/asset.proto#L21)):*

I would keep it separately,  to keep the design with the features alongside the definition. So you can add it to the query for the FungibleToken but don't add to the definition.  
</blockquote></details>

agree done
___
*[`proto/coreum/asset/v1/tx.proto` line 24 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhz0EMDHXHcPYYZYAf:-NHzI1RM41MNF4CdzFvc:b-vcf1lr) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L24)):*

I see you are removing the comments, I like this approach. It will be better if you remove previous comments and also remove the empty lines between methods. 
</blockquote></details>

it was a draft to discuss some points so they were not added yet
___
*[`proto/coreum/asset/v1/tx.proto` line 27 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiZY9s3C-GRNsxNpul:-NHzI5NjEQ5ZH4_zUyF6:bb7f2mf) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L27)):*

Don't forget to add coments. 
</blockquote></details>

done
___
*[`proto/coreum/asset/v1/tx.proto` line 32 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhuK1839psJcb-Qhnb:-NHzK-H422WTq02p8npQ:b-cw0j3g) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L32)):*

this tag adds Equal method to the type. I think the default value for it is false, and removing it is the same as setting it to false. 
</blockquote></details>

it actually affects generated code a bit but we probably don't need this option

Am I right ? @dzmitryhil  As far as I recall you added this option


___
*[`x/asset/handler.go` line 14 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhzn6t514GV4uhGPOo:-NHzKQhP8ISzGOd7yNq7:b8sb2g7) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/handler.go#L14)):*

as dicsussed in the slack, this handler is legacy stuff and should be completely removed. 
</blockquote></details>

but we use it in module.go
And I'm not sure what to put there instead of it

```
// Route returns the asset module's message routing key.
func (am AppModule) Route() sdk.Route {
	return sdk.NewRoute(types.RouterKey, NewHandler(keeper.NewMsgServer(am.keeper)))
}
```
___
*[`x/asset/client/cli/tx.go` line 323 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHi0FML6jDY7XpzwdAs:-NHzLArP9D4WUTyHEH8v:b-scoyfb) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/client/cli/tx.go#L323)):*

just an immature idea, we can combine the freeze and global freeze, you can do `freeze [address]` of `freeze global`
</blockquote></details>

Liked the idea initially but then explored that in freeze we pass amount e.g `10satoshi-...` while in global freeze we pass denom name. So as for me these are 2 completely separate commands and it is better to not mix them at all
___
*[`x/asset/client/cli/tx.go` line 325 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiaNyn--9n5z0KBdE7:-NHyXT2qCTHYQr38fF0i:b-uhx051) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/client/cli/tx.go#L325)):*

Not sure about the part `on all accounts`. It just globally freezes, without touching the accounts. (some for unfreeze)
</blockquote></details>

agree, renamed
___
*[`x/asset/keeper/global_freeze.go` line 16 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHf4xV321Mma3eDYupY:-NHzPliq5ILHuabskfjG:bp30iex) ([raw file](https://github.com/coreumfoundation/coreum/blob/083b7e51e90a74ed97fc60044bb5fe7619eaae0d/x/asset/keeper/global_freeze.go#L16)):*
> checkFeatureAllowed
done
___
*[`x/asset/keeper/global_freeze.go` line 10 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHictpa6v4k2r9xg8ek:-NHyXW0lETHCMbXLpYMe:b-z1u88q) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L10)):*

Where is the logic with the checking of the globally frozen token?
</blockquote></details>

it was a draft added it now
___
*[`x/asset/keeper/global_freeze.go` line 16 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHib9uD4rM58fBN3U3H:-NHzQ1ZLChDO3tRyXw7J:bloeypc) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L16)):*

As I mention in the proto IMHO we should add that flag to the definition.  
</blockquote></details>

I didn't get
Can you elaborate pls ?
___
*[`x/asset/keeper/global_freeze.go` line 23 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHibP7U5ENDql4cRPOa:-NHzQ6Vn1AYJEDGdTETq:b-bbsmbl) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L23)):*

Why we should return the error? It's an idempotent operation, so it is safer to allow to set the same twice here. 
</blockquote></details>

I was 50/50 if it should be idempotent
If you think so then I tend to keep idempotent version here without error

done
___
*[`x/asset/keeper/global_freeze.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHibHeLEvtDNCjjGfyN:-NHzQJvZ7kz6rJhXtKpu:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L26)):*

No need to `else` here you return in `if`
</blockquote></details>

Done.
___
*[`x/asset/types/keys.go` line 45 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI2CCzHCx_tlDx-qLtK:-NI4kby25qmP-9mMDlBo:b-uy1oqn) ([raw file](https://github.com/coreumfoundation/coreum/blob/fd8decfe0d49553983269c37e1be3c49e0f9210a/x/asset/types/keys.go#L45)):*

we don't need a prefix store for each denom. just a single prefix store where denom is the key and presence of that key indicates that denom is frozen. 
</blockquote></details>

I actually had this question to be asked:

```
why do we use store.JoinKeysWithLength GetFungibleTokenKey but don't use it in  freezing ?
And which one I should use in `CreateGlobalFreezePrefix` ?

I probably don't understand fully the logic of `JoinKeysWithLength`
```

Do you mean that I should use just JoinKeys here ?
___
*[`x/asset/types/msgs.go` line 141 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHicOmx5jLCM8K-jO7R:-NHzR02B0warKHKo8XI5:b-h23p89) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/types/msgs.go#L141)):*

Don't understand why you need it (same for next message). Will you `Deconstruct` later in the code?
</blockquote></details>

no I don't deconstruct it but that is the only func I found for validating denom so I used it
The same one is used in `MsgFreezeFungibleToken` and other

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: 8 of 21 files reviewed, 9 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`proto/coreum/asset/v1/tx.proto` line 32 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhuK1839psJcb-Qhnb:-NI7L51z7LSNF_XrCq94:b8jpbbk) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L32)):*

it actually affects generated code a bit but we probably don't need this option

Am I right ? @dzmitryhil  As far as I recall you added this option



</blockquote></details>

IMHO we can always go with default and add tags only if we know the reason for it. 
___
*[`proto/coreum/asset/v1/tx.proto` line 29 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7KtDX0THDflI8xIcb:-NI7KtDX0THDflI8xIcc:bpbmnjl) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/proto/coreum/asset/v1/tx.proto#L29)):*
> ```protobuf
>   // GlobalFreezeFungibleToken freezes fungible token so no operations are allowed with it before unfrozen.
>   // This operation is idempotent so global freeze of already frozen token does nothing.
>   rpc GlobalFreezeFungibleToken(MsgGlobalFreezeFungibleToken) returns (EmptyResponse);
> ```

In terms of action "GloballyFreezeFungible"  IMHO is a more correct name.



_Code quote:_
```protobuf
GlobalFreezeFungible
```
___
*[`x/asset/genesis.go` line 11 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7LH92EHaPIjJY-lOi:-NI7LH933S0Dmd4JP_cI:bgs1hhm) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/genesis.go#L11)):*
> ```Go
> )
> 
> // TODO(yaroslav): Add global freezing logic to genesis once coreum #268 is merged.
> ```

Blocking until initi/export is done
___
*[`x/asset/handler.go` line 14 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhzn6t514GV4uhGPOo:-NI7LMGW28SJtroA0rxc:b-kaqzc9) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/handler.go#L14)):*

but we use it in module.go
And I'm not sure what to put there instead of it

```
// Route returns the asset module's message routing key.
func (am AppModule) Route() sdk.Route {
	return sdk.NewRoute(types.RouterKey, NewHandler(keeper.NewMsgServer(am.keeper)))
}
```
</blockquote></details>

You can keep it, I've already removed it in my pending commit.
___
*[`x/asset/keeper/global_freeze.go` line 16 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHib9uD4rM58fBN3U3H:-NI7LxHp5_4OBzfykuMb:b-oil2qd) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L16)):*

I didn't get
Can you elaborate pls ?
</blockquote></details>

Looks like the line error :-) Resolved
___
*[`x/asset/keeper/global_freeze.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHibHeLEvtDNCjjGfyN:-NI7MK2MBVZhH-fcq0Nr:b-9x8s3s) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L26)):*

Done.
</blockquote></details>

none blocking: I would prefer to keep return and remove else :-)
___
*[`x/asset/keeper/global_freeze.go` line 11 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7OjGhAoYw-Hh1AdXT:-NI7OjGhAoYw-Hh1AdXU:b-g22d1w) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L11)):*
> ```Go
> 
> // SetGlobalFreezeEnabled enables or disables global freeze on a fungible token depending on enabled arg.
> func (k Keeper) SetGlobalFreezeEnabled(ctx sdk.Context, sender sdk.AccAddress, denom string, enabled bool) error {
> ```

I tried, but can't get used to it :-) 
How about having 2 separate functions instead? 
___
*[`x/asset/keeper/global_freeze.go` line 25 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7MF3jCmGey785JWZR:-NI7MF3jCmGey785JWZS:b-1o7rdv) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L25)):*
> ```Go
> 	// Global freeze implemented in an idempotent way, so it is allowed to freeze/unfreeze multiple times without effect.
> 	if enabled {
> 		store.Set(types.CreateGlobalFreezePrefix(denom), []byte{0x01})
> ```

What is the "magic" bytes value? :-) 
Can we call it an empty value and add it to the constants? And what if we save the empty slice will the `Has` function work?

dzmitryhil (commented):
> Reviewed 6 of 12 files at r3, 13 of 13 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 9 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`proto/coreum/asset/v1/tx.proto` line 32 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhuK1839psJcb-Qhnb:-NI7b7k72WW6BkhtM7h7:bfm9vf2) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/proto/coreum/asset/v1/tx.proto#L32)):*

IMHO we can always go with default and add tags only if we know the reason for it. 
</blockquote></details>

ok
but getters are not needed so disabled them
___
*[`proto/coreum/asset/v1/tx.proto` line 29 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7KtDX0THDflI8xIcb:-NI7jda28hTg8mELzkiZ:bzajw1) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/proto/coreum/asset/v1/tx.proto#L29)):*

In terms of action "GloballyFreezeFungible"  IMHO is a more correct name.
</blockquote></details>

agree renamed everywhere including cli, msgs & usage
___
*[`x/asset/handler.go` line 14 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHhzn6t514GV4uhGPOo:-NI7cwDi7t9BQqRi8HaG:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/handler.go#L14)):*

You can keep it, I've already removed it in my pending commit.
</blockquote></details>

ok
___
*[`x/asset/keeper/global_freeze.go` line 26 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHibHeLEvtDNCjjGfyN:-NI7jy3Q8zWeEGGFF2jk:b-pz83e9) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/x/asset/keeper/global_freeze.go#L26)):*

none blocking: I would prefer to keep return and remove else :-)
</blockquote></details>

ok, I also prefer the way it looks without else
___
*[`x/asset/keeper/global_freeze.go` line 25 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7MF3jCmGey785JWZR:-NI7oDiDD374B4paMvvP:bj2rtlu) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L25)):*

What is the "magic" bytes value? :-) 
Can we call it an empty value and add it to the constants? And what if we save the empty slice will the `Has` function work?
</blockquote></details>

agree 
refactored to use `globalFreezeEnabledStoreVal`
also use bytes.Equal to compare values

wojtek-coreum (changes requested):
> Reviewed 3 of 12 files at r3, 5 of 13 files at r4, 14 of 14 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 9 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`proto/coreum/asset/v1/asset.proto` line 32 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NIBpMMWAWROmViSC262:-NIBpMMX4lF5s9fLjTX2:bqgcaur) ([raw file](https://github.com/coreumfoundation/coreum/blob/53c9a0ecce288898c3ec882fe7304ad91d57791b/proto/coreum/asset/v1/asset.proto#L32)):*
> ```protobuf
>   string description = 4;
>   repeated FungibleTokenFeature features = 5;
>   bool globally_frozen = 6;
> ```

please put it above features
___
*[`x/asset/handler.go` line 14 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NIBpqbg4DvzpTxH_zt7:-NIBpqbg4DvzpTxH_zt8:bw8h7hc) ([raw file](https://github.com/coreumfoundation/coreum/blob/53c9a0ecce288898c3ec882fe7304ad91d57791b/x/asset/handler.go#L14)):*
> ```Go
> 
> // NewHandler return tx handler of the asset module.
> func NewHandler(ms keeper.MsgServer) sdk.Handler {
> ```

According to our findings, this function might be removed completely
___
*[`x/asset/keeper/global_freeze.go` line 25 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7MF3jCmGey785JWZR:-NIBtecx8KlYJmCIVr5x:b9wf3u) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L25)):*

agree 
refactored to use `globalFreezeEnabledStoreVal`
also use bytes.Equal to compare values
</blockquote></details>

maybe it would be possible to store empty `[]byte{}`,? Never tried that but we use only `Has` to test the presence so it should work
___
*[`x/asset/types/keys.go` line 45 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI2CCzHCx_tlDx-qLtK:-NIBqOwKATUlW0UmeN20:bnhpizu) ([raw file](https://github.com/coreumfoundation/coreum/blob/fd8decfe0d49553983269c37e1be3c49e0f9210a/x/asset/types/keys.go#L45)):*

I actually had this question to be asked:

```
why do we use store.JoinKeysWithLength GetFungibleTokenKey but don't use it in  freezing ?
And which one I should use in `CreateGlobalFreezePrefix` ?

I probably don't understand fully the logic of `JoinKeysWithLength`
```

Do you mean that I should use just JoinKeys here ?
</blockquote></details>

If you don't create subkeys then `JoinKeys` is enough

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/asset/keeper/global_freeze.go` line 25 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7MF3jCmGey785JWZR:-NICOIpC4mDnz2p0OnOg:b-54jn8l) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L25)):*

maybe it would be possible to store empty `[]byte{}`,? Never tried that but we use only `Has` to test the presence so it should work
</blockquote></details>

Right @ysv can you try with empty?

dzmitryhil (commented):
> Reviewed 11 of 14 files at r5, 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`proto/coreum/asset/v1/asset.proto` line 32 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NIBpMMWAWROmViSC262:-NICR2na8cH1pHhr9ImA:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/53c9a0ecce288898c3ec882fe7304ad91d57791b/proto/coreum/asset/v1/asset.proto#L32)):*

please put it above features
</blockquote></details>

Done.
___
*[`x/asset/genesis.go` line 11 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7LH92EHaPIjJY-lOi:-NICR6gI593UwtThtBui:bmiz1f8) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/genesis.go#L11)):*

Blocking until initi/export is done
</blockquote></details>

IDK what is the state of 268 but if this one is ready before we can merge it first and then I will do another PR with genesis stuff
___
*[`x/asset/handler.go` line 14 at r5](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NIBpqbg4DvzpTxH_zt7:-NICRKt37ng-aVHWzMb-:b-wzxu52) ([raw file](https://github.com/coreumfoundation/coreum/blob/53c9a0ecce288898c3ec882fe7304ad91d57791b/x/asset/handler.go#L14)):*

According to our findings, this function might be removed completely
</blockquote></details>

It is done in Dmitry's PR so I just keep to not cause merge conflicts
___
*[`x/asset/keeper/global_freeze.go` line 11 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7OjGhAoYw-Hh1AdXT:-NICTnO6CMF0rXdJW13x:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L11)):*

I tried, but can't get used to it :-) 
How about having 2 separate functions instead? 
</blockquote></details>

Done.
___
*[`x/asset/keeper/global_freeze.go` line 25 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7MF3jCmGey785JWZR:-NICVR_pBEU3XXafj1Dg:bwyyyij) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L25)):*

Right @ysv can you try with empty?
</blockquote></details>

it will not work
The thing is that []byte{} is equal to nil
and nil is returned from store.Get if key doesn't exist. I checked it by running unit tests and they failed.
Actually I saw this approach is @miladz68 's PR initially. He told that he saw it to be done like this in some std cosmos module


Screenshot 2022-12-01 at 12.44.56
___
*[`x/asset/types/keys.go` line 45 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI2CCzHCx_tlDx-qLtK:-NICQx1w7OxXgfxY1Nu_:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/fd8decfe0d49553983269c37e1be3c49e0f9210a/x/asset/types/keys.go#L45)):*

If you don't create subkeys then `JoinKeys` is enough
</blockquote></details>

Done.



_Code quote:_
```Go
CreateGlobalFreezePrefix
```

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/asset/global_freeze.go` line 75 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiYfz-6t3MvytWnDPT:-NICbrov9gbY6dSE7sjW:b7750em) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/integration-tests/asset/global_freeze.go#L75)):*

it is still a draft
</blockquote></details>

This commen is still actual
___
*[`x/asset/genesis.go` line 11 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7LH92EHaPIjJY-lOi:-NICaWlB7lRGgo4BhBNV:b-ekiux1) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/genesis.go#L11)):*

IDK what is the state of 268 but if this one is ready before we can merge it first and then I will do another PR with genesis stuff
</blockquote></details>

Ok, let's do that way. 
___
*[`x/asset/keeper/global_freeze.go` line 25 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7MF3jCmGey785JWZR:-NICb7mk3LjA4zx-vtNI:b-p52d7g) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L25)):*

it will not work
The thing is that []byte{} is equal to nil
and nil is returned from store.Get if key doesn't exist. I checked it by running unit tests and they failed.
Actually I saw this approach is @miladz68 's PR initially. He told that he saw it to be done like this in some std cosmos module


Screenshot 2022-12-01 at 12.44.56
</blockquote></details>

Ok, how about "0x0"
___
*[`x/asset/keeper/issue.go` line 80 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NICbO2z6dhzqFpYTH70:-NICbO2z6dhzqFpYTH71:btr6er1) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcae4e64c72738f9a9986848a3ae33704dfe154a/x/asset/keeper/issue.go#L80)):*
> ```Go
> 		Description:    metadata.Description,
> 		Features:       definition.Features,
> 		GloballyFrozen: bytes.Equal(globFreezeVal, globalFreezeEnabledStoreVal),
> ```

Let's move `that part + globalFreezeEnabledStoreVal + globFreezeVal := ctx.KVStore(k.storeKey).Get(types.CreateGlobalFreezePrefix(denom)) ` to one method `isGlobalyFrozen(ctx, denom) bool` and to the `global_freeze.go` file.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/asset/keeper/issue.go` line 80 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NICbO2z6dhzqFpYTH70:-NICc7p5CjGDU0XWesn3:bw5rhge) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcae4e64c72738f9a9986848a3ae33704dfe154a/x/asset/keeper/issue.go#L80)):*

Let's move `that part + globalFreezeEnabledStoreVal + globFreezeVal := ctx.KVStore(k.storeKey).Get(types.CreateGlobalFreezePrefix(denom)) ` to one method `isGlobalyFrozen(ctx, denom) bool` and to the `global_freeze.go` file. 
</blockquote></details>

And checl that part: `ctx.KVStore(k.storeKey).Has(types.CreateGlobalFreezePrefix(coin.Denom))`, you do that same here, but in in different way. Reuse the new func here as well.

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`integration-tests/asset/global_freeze.go` line 75 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NHiYfz-6t3MvytWnDPT:-NICvwr6AQgg5wddIEXW:bg99t5) ([raw file](https://github.com/coreumfoundation/coreum/blob/562f41fccc0b9ac1a5c3635d7de9d09d834cea9b/integration-tests/asset/global_freeze.go#L75)):*

This commen is still actual
</blockquote></details>

The thing is that I do this to not test events but to get denom of an FT
Of course I can build it from symbol + issuer but since it is an integration test I try to relay only on what system provides to outside world so if one day the way we build denom i changed tests are still same

But if you insist I can "hardcode" denom here
___
*[`x/asset/genesis.go` line 11 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7LH92EHaPIjJY-lOi:-NICwXxo8xaNe-z1OSSG:bkmxdy1) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/genesis.go#L11)):*

Ok, let's do that way. 
</blockquote></details>

if you are ok with a separate PR pls resolve 
I've added this to task checklist also to not forget
___
*[`x/asset/keeper/global_freeze.go` line 25 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NI7MF3jCmGey785JWZR:-NICxp8w1iBDb9XlONzc:bb7f2mf) ([raw file](https://github.com/coreumfoundation/coreum/blob/0be6644389fd0a98bf9359d8d32f86693968681e/x/asset/keeper/global_freeze.go#L25)):*

Ok, how about "0x0"
</blockquote></details>

done
___
*[`x/asset/keeper/issue.go` line 80 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NICbO2z6dhzqFpYTH70:-NICvI2pDcFT8u8iZE8s:b-5fdftb) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcae4e64c72738f9a9986848a3ae33704dfe154a/x/asset/keeper/issue.go#L80)):*

And checl that part: `ctx.KVStore(k.storeKey).Has(types.CreateGlobalFreezePrefix(coin.Denom))`, you do that same here, but in in different way. Reuse the new func here as well.
</blockquote></details>

very good point thx !

@miladz68 pls note the way `globalFreezeEnabledStoreVal` is done in my PR you should probably do the same in your asset exponent:
1. move to constant
2. use 0x00

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 6 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

___
*[`x/asset/keeper/issue.go` line 80 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NICbO2z6dhzqFpYTH70:-NID2YziDaIBH2rlIq9L:b-ojzbox) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcae4e64c72738f9a9986848a3ae33704dfe154a/x/asset/keeper/issue.go#L80)):*

very good point thx !

@miladz68 pls note the way `globalFreezeEnabledStoreVal` is done in my PR you should probably do the same in your asset exponent:
1. move to constant
2. use 0x00
</blockquote></details>

Don't see the changes!

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: 18 of 22 files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`x/asset/keeper/issue.go` line 80 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/271#-NICbO2z6dhzqFpYTH70:-NIDeoHW61ofd5N9mgOz:bb4xt84) ([raw file](https://github.com/coreumfoundation/coreum/blob/bcae4e64c72738f9a9986848a3ae33704dfe154a/x/asset/keeper/issue.go#L80)):*

Don't see the changes!
</blockquote></details>

pushed now

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: 18 of 22 files reviewed, 5 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

dzmitryhil (approved):
> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 5 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

wojtek-coreum (approved):
> Reviewed 5 of 6 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/271)* status: all files reviewed, 1 unresolved discussion (waiting on @miladz68)

PR #275, opened 2022-11-28
Deterministic gas for Send and MultiSend depending on number of coins transferred

wojtek-coreum:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)

### DIFF
--- a/pkg/config/network_test.go
+++ b/pkg/config/network_test.go
@@ -34,7 +34,7 @@ var feeConfig = config.FeeConfig{
 		LongEmaBlockLength:      5,
 	}),
 	DeterministicGas: config.DeterministicGasRequirements{
-		BankSend: 10,
+		BankSendPerEntry: 10,
 	},
 }
 
@@ -195,7 +195,7 @@ func TestAddFundsToGenesis(t *testing.T) {
 
 func TestDeterministicGas(t *testing.T) {
 	assert.Equal(t, config.DeterministicGasRequirements{
-		BankSend: 10,
+		BankSendPerEntry: 10,
 	}, testNetwork().DeterministicGas())
 }
 
@@ -308,6 +308,8 @@ func TestNetworkConfigConditions(t *testing.T) {
 	assertT := assert.New(t)
 	for _, n := range config.EnabledNetworks() {
 		assert.NoError(t, n.FeeModel().Params().ValidateBasic())
-		assertT.Greater(n.DeterministicGas().BankSend, uint64(0))
+
+		// FIXME (wojtek): add all the deterministic gas fields here
+		assertT.Greater(n.DeterministicGas().BankSendPerEntry, uint64(0))
 	}
 }


miladz68 (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @wojtek-coreum, and @ysv)

___
*[`integration-tests/index.go` line 47 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy1kXTCBqssdksFlHt:-NHy1kXTCBqssdksFlHu:b-umezen) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/integration-tests/index.go#L47)):*
> ```Go
> 				bank.TestMultiSendDeterministicGasManyCoins,
> 				bank.TestMultiSendDeterministicGasMoreCoins,
> 				bank.TestMultiSend,
> ```

non-blocking
maybe it is time to group package tests at package level and import a list here.

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 4 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`integration-tests/index.go` line 46 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy20h7A07mIPpXioQG:-NHy20h7A07mIPpXioQH:bsuh5o7) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/integration-tests/index.go#L46)):*
> ```Go
> 				bank.TestSendGasEstimation,
> 				bank.TestMultiSendDeterministicGasManyCoins,
> 				bank.TestMultiSendDeterministicGasMoreCoins,
> ```

Duplicated the line. 
___
*[`pkg/config/deterministic_gas.go` line 72 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy4gXuDIJpinA0c7Jz:-NHy4gXuDIJpinA0c7K-:b-a3zcz1) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/pkg/config/deterministic_gas.go#L72)):*
> ```Go
> 
> 	// x/bank
> 	BankSendTokenNumberLimit   int
> ```

I would choose another name for that var since it's not `Limit` it's a `BankSendTokenNumberThreshold` or  `IankSendTokenIncludedNumber` or any other name :-)
___
*[`pkg/config/deterministic_gas.go` line 137 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy3nTs7XTK61Sm6OU8:-NHy3nTs7XTK61Sm6OU9:bu9fie6) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/pkg/config/deterministic_gas.go#L137)):*
> ```Go
> 
> 		if numOfCoins > dgr.BankSendTokenNumberLimit {
> 			gas += uint64(numOfCoins-dgr.BankSendTokenNumberLimit) * dgr.BankSendAdditionalTransfer
> ```

Good idea. In that case, the gas is fully deterministic.  We actually can do the same for any TX type (apart from wasm).

dzmitryhil (commented):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 4 unresolved discussions (waiting on @wojtek-coreum and @ysv)

miladz68 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 5 unresolved discussions (waiting on @dzmitryhil and @wojtek-coreum)

___
*[`integration-tests/index.go` line 46 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy20h7A07mIPpXioQG:-NI0rs3i5njO6zwmB7MG:bki7sal) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/integration-tests/index.go#L46)):*

Duplicated the line. 
</blockquote></details>

not really duplicates, one is `ManyCoins` the other i `MoreCoins`

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 5 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

___
*[`pkg/config/deterministic_gas.go` line 27 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NI7usPy20OWCk5M1eAA:-NI7usPy20OWCk5M1eAB:b-y9sw43) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/pkg/config/deterministic_gas.go#L27)):*
> ```Go
> 		AssetBurnFungibleToken:     35000,
> 
> 		BankSendTokenNumberLimit:   5,
> ```

can you check roughly how much of gas each additional Bank send adds ?
After this lets calculate how much entities could fit into a single BankSend/BankMultisend. Is it 1k, 10k ... ?
Also pls check how what is the size of it in bytes maybe we will reach max block size faster than gas limit ?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: 10 of 12 files reviewed, 5 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/index.go` line 47 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy1kXTCBqssdksFlHt:-NIC_X_CCjxOykNEfnq0:b-afi0at) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/integration-tests/index.go#L47)):*

100%
</blockquote></details>

Can we make it in separate PR?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 5 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: 6 of 17 files reviewed, 5 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/index.go` line 47 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy1kXTCBqssdksFlHt:-NICdGS32dADMjA-zKtQ:b2nigmo) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/integration-tests/index.go#L47)):*

Can we make it in separate PR?
</blockquote></details>

yea
___
*[`pkg/config/deterministic_gas.go` line 72 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy4gXuDIJpinA0c7Jz:-NICSkGdDtICxWd7odnn:b-l7u96y) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/pkg/config/deterministic_gas.go#L72)):*

I would choose another name for that var since it's not `Limit` it's a `BankSendTokenNumberThreshold` or  `IankSendTokenIncludedNumber` or any other name :-)
</blockquote></details>

removed it completely
___
*[`pkg/config/deterministic_gas.go` line 137 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NHy3nTs7XTK61Sm6OU8:-NICSpknERORVLu1TA--:b-ecxd5t) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/pkg/config/deterministic_gas.go#L137)):*

Good idea. In that case, the gas is fully deterministic.  We actually can do the same for any TX type (apart from wasm). 
</blockquote></details>

yea, please don't block on this comment :-)

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: 6 of 17 files reviewed, 4 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/asset/freeze.go` line 96 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIChzwVA7Fo6Ij44yoK:-NIChzwVA7Fo6Ij44yoL:bchkfml) ([raw file](https://github.com/coreumfoundation/coreum/blob/adfab58db8e18e1032fce6aff1438ee408aaf4fd/integration-tests/asset/freeze.go#L96)):*
> ```Go
> 		chain.Faucet.FundAccountsWithOptions(ctx, recipient, testing.BalancesOptions{
> 			Messages: []sdk.Msg{
> 				&banktypes.MsgSend{Amount: make(sdk.Coins, 1)},
> ```

IMHO we shuld include one coin to the default gas, so you won't have to that here. So the deterministic gas is for one coin for each next you pay extra. 
@ysv @miladz68  ?

dzmitryhil (commented):
> Reviewed 11 of 11 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 4 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 4 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/asset/freeze.go` line 96 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIChzwVA7Fo6Ij44yoK:-NIWGx_G0cISCN4Or7am:b-18cy4v) ([raw file](https://github.com/coreumfoundation/coreum/blob/adfab58db8e18e1032fce6aff1438ee408aaf4fd/integration-tests/asset/freeze.go#L96)):*

IMHO we shuld include one coin to the default gas, so you won't have to that here. So the deterministic gas is for one coin for each next you pay extra. 
@ysv @miladz68  ?
</blockquote></details>

I was also thinking about it, but then it's harder to explain to the users eventually. @ysv @miladz68

miladz68 (approved):
> Reviewed 8 of 11 files at r3, 1 of 2 files at r4, 1 of 1 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 3 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`integration-tests/asset/freeze.go` line 96 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIChzwVA7Fo6Ij44yoK:-NIXAJOA0qYhELsUu8oZ:bv3s076) ([raw file](https://github.com/coreumfoundation/coreum/blob/adfab58db8e18e1032fce6aff1438ee408aaf4fd/integration-tests/asset/freeze.go#L96)):*

I was also thinking about it, but then it's harder to explain to the users eventually. @ysv @miladz68 
</blockquote></details>

I agree with Dima here, first coin on the house :smile: , then you start paying

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: 9 of 18 files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`integration-tests/asset/freeze.go` line 96 at r3](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIChzwVA7Fo6Ij44yoK:-NIaN7Hm6iGdHtcYIulJ:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/adfab58db8e18e1032fce6aff1438ee408aaf4fd/integration-tests/asset/freeze.go#L96)):*

Agree with Dima
</blockquote></details>

Done.
___
*[`integration-tests/bank/gas.go` line 29 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYSqqR0xf8LGkWAJ6_:-NIaNznyEGrjr6v9Ok89:bx7z2xo) ([raw file](https://github.com/coreumfoundation/coreum/blob/5f21708081d4df2630d80df6d5bb04253342df91/integration-tests/bank/gas.go#L29)):*

I don't like these magic in all tests
WDYT about considering 1 to be default length of `Amount` if it is 0 in integration tests only in `ComputeNeededBalanceFromOptions` ?
Maybe it is too magical but it is still test helper 

Or maybe it even makes sense to do like this in deterministicGas because if Amount length is 0 it is an invalid msg

Or maybe you have any other idea to avoid this `Amount: make(sdk.Coins, 1)` in each test ?
</blockquote></details>

We decided to include 1 free coin in the deterministic gas, so now it is simpler
___
*[`integration-tests/bank/gas.go` line 146 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYAaCl1Sn8rCqMjFNN:-NIaOU-81aLJlEkGZrA0:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/5f21708081d4df2630d80df6d5bb04253342df91/integration-tests/bank/gas.go#L146)):*

you can actually move this check to be done right after you do `bankSendGas := chain.GasLimitByMsgs(msg)` because actual request is not required to make this assertion and we can fail fast
</blockquote></details>

Done.
___
*[`pkg/config/deterministic_gas.go` line 27 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NI7usPy20OWCk5M1eAA:-NIaPqvBCPn55b42Kp-Q:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/00069284e40c9029c3d6e33845fc7d81b91563e2/pkg/config/deterministic_gas.go#L27)):*

Hm they are slightly different actually
And I see that you introduced different variables for `BankSend & BankMultiSend` but set it to the same values 50000. Shouldn't multisend be higher.

Also don't forget to subtract FixedGas from these values.
I think that to be correct we should do the following:
1. use FixedGas as starting value for both Send & Multisend of 0 coins
2. define `BankSendPerTransfer: 22000` & `BankMultisendPerTransfer: 27000`

In this approach we have less variables and it is more clear:
e.g `fee = BaseFee + send_number*BankSenPerTransfer`
</blockquote></details>

Done.
___
*[`pkg/config/deterministic_gas.go` line 29 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYRzgeCFOUdBNRFEah:-NIaXCBn7gpD0qBisBI3:b-hr7i7h) ([raw file](https://github.com/coreumfoundation/coreum/blob/8a79eb1896b462f32ef51deb132391c88dc01984/pkg/config/deterministic_gas.go#L29)):*

> After this lets calculate how much entities could fit into a single BankSend/BankMultisend. Is it 1k, 10k ... ?
> Also pls check how what is the size of it in bytes maybe we will reach max block size faster than gas limit ?

these questions from previous comment is still relevant. Moved to a separate thread
Can you reference them pls ?
</blockquote></details>

It highly depends on the transferred amounts, because their length is the main factor here. IIRC, simple bank send takes ~550 bytes. Then you need to add the length of each additional amount, and in case of multi send also the address, which is 46-byte long in case of devnet. So if you use 14-byte amounts&symbol pairs , then you have 60 bytes bet each sender/receiver.
___
*[`pkg/config/deterministic_gas.go` line 124 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYWKI_11mXSl_6VitZ:-NIaQ-aH-vdwwl1Hx3-Z:b-oe7c8) ([raw file](https://github.com/coreumfoundation/coreum/blob/5f21708081d4df2630d80df6d5bb04253342df91/pkg/config/deterministic_gas.go#L124)):*

minor;
1. why don't use name it numOfInputCoins ?
2. I prefer to explicitly set 0 in such cases: `numOfInputCoins := 0` instead of `var`
</blockquote></details>

Ad. 1. because later we eventually assign the value of `numOfOutputCoins` to it

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: 9 of 18 files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`pkg/config/deterministic_gas.go` line 29 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYRzgeCFOUdBNRFEah:-NIabGjgDQQ7jo-EE7ja:bdtn6xr) ([raw file](https://github.com/coreumfoundation/coreum/blob/8a79eb1896b462f32ef51deb132391c88dc01984/pkg/config/deterministic_gas.go#L29)):*

It highly depends on the transferred amounts, because their length is the main factor here. IIRC, simple bank send takes ~550 bytes. Then you need to add the length of each additional amount, and in case of multi send also the address, which is 46-byte long in case of devnet. So if you use 14-byte amounts&symbol pairs , then you have 60 bytes bet each sender/receiver.
</blockquote></details>

And in case of multiple-senders in multi send you have additional signature for each additional sender

dzmitryhil (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: 9 of 18 files reviewed, 5 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

miladz68 (dismissed):
> Reviewed 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 5 unresolved discussions (waiting on @wojtek-coreum and @ysv)

ysv (changes requested):
> Reviewed 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 4 unresolved discussions (waiting on @wojtek-coreum)

___
*[`pkg/config/deterministic_gas.go` line 29 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYRzgeCFOUdBNRFEah:-NIbVOLVDtMzj3PXKuB5:bxz4thn) ([raw file](https://github.com/coreumfoundation/coreum/blob/8a79eb1896b462f32ef51deb132391c88dc01984/pkg/config/deterministic_gas.go#L29)):*

And in case of multiple-senders in multi send you have additional signature for each additional sender
</blockquote></details>

so roughly for current  config it should be 
- block size limit `22020096bytes`: `(22020096 - 550) / 60 = 367k` limited to 367k entries in terms of byte size
- gas`50000000`: `(50000000 - 50000) / 22000 = 2270`

I'm just not sure if for gas I should multiply it by gas price ? `(50000000 - 50000*0.0625) / (22000*0.0625)` ?

WDYT ?

I just need rough estimation for this value to give to Reza
___
*[`pkg/config/deterministic_gas.go` line 124 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYWKI_11mXSl_6VitZ:-NIbV7Hn0xB2nGLAT_HI:b32wtd) ([raw file](https://github.com/coreumfoundation/coreum/blob/5f21708081d4df2630d80df6d5bb04253342df91/pkg/config/deterministic_gas.go#L124)):*

Ad. 1. because later we eventually assign the value of `numOfOutputCoins` to it
</blockquote></details>

ok
___
*[`pkg/config/deterministic_gas.go` line 74 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIbTgMh5c0y9CMTCtsj:-NIbTgMh5c0y9CMTCtsk:b-v5rru3) ([raw file](https://github.com/coreumfoundation/coreum/blob/a724f97473cf7853038c97420366d0465e4b77af/pkg/config/deterministic_gas.go#L74)):*
> ```Go
> 
> 	// x/bank
> 	BankSendPerTransfer      uint64
> ```

I suggest to name if differently because send & transfer are quite synonymous words IMO
What about `BankSendPerEntry` or `BankSendPerCoin` 
Same applies to multisend but I'm not sure`BankMultiSendPerCoin` is a good name because we calculate max of in/out so I prefer entry 
___
*[`pkg/config/deterministic_gas.go` line 120 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIbUSIGBpuLCsfLlXc1:-NIbUSIGBpuLCsfLlXc2:bcseyw3) ([raw file](https://github.com/coreumfoundation/coreum/blob/a724f97473cf7853038c97420366d0465e4b77af/pkg/config/deterministic_gas.go#L120)):*
> ```Go
> 		return dgr.AssetBurnFungibleToken, true
> 	case *banktypes.MsgSend:
> 		if len(m.Amount) == 0 {
> ```

not-critical:
I would prefer smth like this for more explicit logic & avoid doing return twice


```
entriesNum := len(m.Amount)
if entriesNum == 0 {
  entriesNum = 1
}

return entriesNum*dgr.BankSendPerTransfer, true
```

same for multisend

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: all files reviewed, 3 unresolved discussions (waiting on @ysv)

___
*[`pkg/config/deterministic_gas.go` line 29 at r6](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIYRzgeCFOUdBNRFEah:-NIbYk9W6HEqe4_QNu4p:b-uvf1nk) ([raw file](https://github.com/coreumfoundation/coreum/blob/8a79eb1896b462f32ef51deb132391c88dc01984/pkg/config/deterministic_gas.go#L29)):*

so roughly for current  config it should be 
- block size limit `22020096bytes`: `(22020096 - 550) / 60 = 367k` limited to 367k entries in terms of byte size
- gas`50000000`: `(50000000 - 50000) / 22000 = 2270`

I'm just not sure if for gas I should multiply it by gas price ? `(50000000 - 50000*0.0625) / (22000*0.0625)` ?

WDYT ?

I just need rough estimation for this value to give to Reza



</blockquote></details>

In terms of gas, you don't multiply it by price - gas is gas
___
*[`pkg/config/deterministic_gas.go` line 74 at r7](https://reviewable.io/reviews/coreumfoundation/coreum/275#-NIbTgMh5c0y9CMTCtsj:-NIbZBRcDfy79Oqf-qKf:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/a724f97473cf7853038c97420366d0465e4b77af/pkg/config/deterministic_gas.go#L74)):*

I suggest to name if differently because send & transfer are quite synonymous words IMO
What about `BankSendPerEntry` or `BankSendPerCoin` 
Same applies to multisend but I'm not sure`BankMultiSendPerCoin` is a good name because we calculate max of in/out so I prefer entry 
</blockquote></details>

Done.

ysv (approved):
> Reviewed 2 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/275)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum)

PR #294, opened 2022-12-14
Collect metric describing difference between deterministic gas and real gas consumption

wojtek-coreum:
This change is [Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)


dzmitryhil (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68 and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 8 unresolved discussions (waiting on @miladz68 and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 7 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfVpX317CRs3pskXBK:-NJj2Jfr3jVPqW68xRqc:bsa7dsn) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L7)):*

I don't have exp working with metrics in cosmos
But I thought that everything is included in cosmos sdk and we don't need to add anything externally.
Just import & use this `"github.com/cosmos/cosmos-sdk/telemetry"`

Or how does it work ?

</blockquote></details>

I inspected their telemetry and I ddn't find functions to create histograms. That's why I used the package they use to create one.
___
*[`x/deterministicgas/types/gas.go` line 7 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfWxsQDEpZO1kh8RdO:-NJj2S_k30fhcWjpgSww:b-w1y47o) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L7)):*

do we need to somehow configure go-metrics here ?
</blockquote></details>

no, it works out of the box - cosmos sdk uses it the same way
___
*[`x/deterministicgas/types/gas.go` line 97 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfUqrcAeD-A9XglJQh:-NJj2rIZ9Lan7RhBaMxl:b-a8i1lj) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L97)):*

I see you use `method.MethodName`  but I sort of disagree with it because we should use msg type IMO not method name. 
We can do this in a multiple ways
1. use smth like: `proto.MessageName(msg)`
2. use smth like `fmt.Sprintf("%T")`
</blockquote></details>

Both proposed solutions use reflection under the hood
___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfZfl60AIyC_xbCf3f:-NJj33YT0Uaw9XLitJ0J:b-fs027l) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

I think it should be `(nondetermGas - deterministicGas) / deterministicGas`

IMO we want to know difference in % from value we estimated in deterministic gas but not from real consumption
</blockquote></details>

It depends on the question you want to ask. I assumed that we want to know how much deterministic gas is higher/lower from default behavior. But if you want to see it the other way, this is also a valid way.
___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJf_IHD57ydkIdmhmb1:-NJj3SYc7K21ivJkDRsk:b-2dofs8) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

is division by 0 a real case here ?
</blockquote></details>

good point, didn't think about it. Will add a check

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 8 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 121 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJf_PIUCZ3Z-bHomzE4:-NJj7eYN4ZuczAUZ4Ew-:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L121)):*

don't really like the name here
maybe smth like `reportDeterministicGasMetric`
</blockquote></details>

Done.
___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJf_IHD57ydkIdmhmb1:-NJj7rBV8NKqD42ctDMO:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

good point, didn't think about it. Will add a check
</blockquote></details>

Done.
___
*[`x/deterministicgas/types/gas.go` line 127 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJf_0Lg65mOLGxCrntm:-NJj7yIADE6AIqgXAkJB:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L127)):*

`msg_type` ?
</blockquote></details>

Done.

miladz68 (changes requested):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 9 unresolved discussions (waiting on @wojtek-coreum and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 7 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJjwQ_RC6G711Ci4xqJ:-NJjwQ_RC6G711Ci4xqK:bbduf68) ([raw file](https://github.com/coreumfoundation/coreum/blob/05d3d7804503a1663ca412d1c0b3dc40d87ac361/x/deterministicgas/types/gas.go#L7)):*
> ```Go
> 	"fmt"
> 
> 	"github.com/armon/go-metrics"
> ```

I looked at how armon/go-metrics generates those buckets, and it looks that they are created on the fly based on the historical ans statistical info, but I wonder how it will not cause problems when such buckets are synced into prometheus. first of all prometheus calculates the diff between each step to correctly understand rates and also service restarts. if thos buckets are constantly changing names then it will throw prometheus off, specially during service restarts. 
Maybe we should pre calculate those buckets and set them. I see that go-armon does not provide a way to do it. so we may need to have another tool ?

miladz68 (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 10 unresolved discussions (waiting on @wojtek-coreum and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJk4u2O9Bo4kdacPCxZ:-NJk4u2P8GRPh0ZzxOmu:b4e2ash) (no related file):*
I check the exported format and it looks like this
```
{"Name":"tx.msg.send","Value":1000000,"Labels":{"denom":"ducore"}}],"Points":[],"Counters":[{"Name":"new.account","Count":2,"Rate":0.2,"Sum":2,"Min":1,"Max":1,"Mean":1,"Stddev":0,"Labels":{}},
```

I don't think we will be able to extract meaningful information from this data. 

___

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`x/deterministicgas/types/gas.go` line 7 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfVpX317CRs3pskXBK:-NJkbDv6CSrOGoMFxIDF:b-hr69ny) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L7)):*

I inspected their telemetry and I ddn't find functions to create histograms. That's why I used the package they use to create one.
</blockquote></details>

hm, ok
lets talk on the call
___
*[`x/deterministicgas/types/gas.go` line 97 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfUqrcAeD-A9XglJQh:-NJkZiBqC8eZ5A6tbL3d:b-equ8av) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L97)):*

Both proposed solutions use reflection under the hood
</blockquote></details>

benchmarked 
```
BenchmarkSprintfReflect-12      	15879584	        73.12 ns/op
BenchmarkProtoMessageName-12    	21099068	        53.72 ns/op
```

I don't think that 50-75ns adds that much overhead comparing to some other slow parts so I would prefer to proceed with it
___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfZfl60AIyC_xbCf3f:-NJkaTCf0-Nk6EVFqded:bdur5ij) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

It depends on the question you want to ask. I assumed that we want to know how much deterministic gas is higher/lower from default behavior. But if you want to see it the other way, this is also a valid way.
</blockquote></details>

Hm so you want to compare difference in % from value we got for current tx
While I want difference in % from value we have put in our config

for me my metric makes a bit more sense because we always want to know deviation from value we consider to be average not specific value we get

lets say that we have msg gas for some type = 50k
and we get 2 specific cases of gas usage: 70k & 35k

according to you formula deviation is:
(35-50)/35 = 42%
(70-50)/70 = 28%

42% > 28% which means that 35 has bigger deviation than 70. 
50-35 < |50-75| it is vice versa if you calculate it is absolute value 

according to my formula:
(35-50)/50 = 30%
(70-50)/50 = 40%

30% < 40%
50-35 < |50-75|

@dzmitryhil & @miladz68 WDYT ?
___
*[`x/deterministicgas/types/gas.go` line 127 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJf_0Lg65mOLGxCrntm:-NJkb-9BEbMFx3UEX3nl:bc3b1ei) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L127)):*

Done.
</blockquote></details>

we should either do change I proposed in comment before to change this to be really msgType
or rename variable msgType to msgMethod
___
*[`x/deterministicgas/types/gas.go` line 7 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJjwQ_RC6G711Ci4xqJ:-NJkbApW8enbgQQQq1ly:b66x5jp) ([raw file](https://github.com/coreumfoundation/coreum/blob/05d3d7804503a1663ca412d1c0b3dc40d87ac361/x/deterministicgas/types/gas.go#L7)):*

I looked at how armon/go-metrics generates those buckets, and it looks that they are created on the fly based on the historical ans statistical info, but I wonder how it will not cause problems when such buckets are synced into prometheus. first of all prometheus calculates the diff between each step to correctly understand rates and also service restarts. if thos buckets are constantly changing names then it will throw prometheus off, specially during service restarts. 
Maybe we should pre calculate those buckets and set them. I see that go-armon does not provide a way to do it. so we may need to have another tool ?
</blockquote></details>

lets discuss on the call

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfZfl60AIyC_xbCf3f:-NJnf7PhCS5Xc0NBuz0Z:bo34v4z) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

Hm so you want to compare difference in % from value we got for current tx
While I want difference in % from value we have put in our config

for me my metric makes a bit more sense because we always want to know deviation from value we consider to be average not specific value we get

lets say that we have msg gas for some type = 50k
and we get 2 specific cases of gas usage: 70k & 35k

according to you formula deviation is:
(35-50)/35 = 42%
(70-50)/70 = 28%

42% > 28% which means that 35 has bigger deviation than 70. 
50-35 < |50-75| it is vice versa if you calculate it is absolute value 

according to my formula:
(35-50)/50 = 30%
(70-50)/50 = 40%

30% < 40%
50-35 < |50-75|

@dzmitryhil & @miladz68 WDYT ?

</blockquote></details>

Actually the both solutions gives the same result, since we know how to commute the gas we want based on the metric result. But I would implement it in the way @ysv did if I implemented that task :-)

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJk4u2O9Bo4kdacPCxZ:-NJncr-G90fVbEDq84JJ:bxosgjj) (no related file):*

I check the exported format and it looks like this
```
{"Name":"tx.msg.send","Value":1000000,"Labels":{"denom":"ducore"}}],"Points":[],"Counters":[{"Name":"new.account","Count":2,"Rate":0.2,"Sum":2,"Min":1,"Max":1,"Mean":1,"Stddev":0,"Labels":{}},
```

I don't think we will be able to extract meaningful information from this data. 
</blockquote></details>

This is what is logged by cored, right? But we don't care about this. We care only about what is available to prometheus:

```
al",quantile="0.9"} 1.5969055891036987
deterministic_gas_factor{tx_type="SubmitProposal",quantile="0.99"} 1.5969055891036987
deterministic_gas_factor_sum{tx_type="SubmitProposal"} 6.360880732536316
deterministic_gas_factor_count{tx_type="SubmitProposal"} 4
```

and everything we need is there

___
___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfZfl60AIyC_xbCf3f:-NJne-tY-SEa7-iub8GK:b-hywzpj) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

Hm so you want to compare difference in % from value we got for current tx
While I want difference in % from value we have put in our config

for me my metric makes a bit more sense because we always want to know deviation from value we consider to be average not specific value we get

lets say that we have msg gas for some type = 50k
and we get 2 specific cases of gas usage: 70k & 35k

according to you formula deviation is:
(35-50)/35 = 42%
(70-50)/70 = 28%

42% > 28% which means that 35 has bigger deviation than 70. 
50-35 < |50-75| it is vice versa if you calculate it is absolute value 

according to my formula:
(35-50)/50 = 30%
(70-50)/50 = 40%

30% < 40%
50-35 < |50-75|

@dzmitryhil & @miladz68 WDYT ?

</blockquote></details>

Yes, you are right. But it is a matter of question you want to ask and action you want to take based on the answer.
We can't affect the real gas taken by the message, we may only adjust deterministic gas value. That's why, IMO asking question
"how much deterministic gas deviates from real gas?" makes more sense in this case than "how much real gas deviates from deterministic gas". Because deterministic gas is the value we control and may adjust.

Maybe we shouldn't calculate percentages but only absolute values? Then we always get the difference in gas unit and we may take max, min, average and deviation which seems to be exactly what we need.
___
*[`x/deterministicgas/types/gas.go` line 7 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJjwQ_RC6G711Ci4xqJ:-NJndPcK1qUx43X8tdbc:bjpqwfm) ([raw file](https://github.com/coreumfoundation/coreum/blob/05d3d7804503a1663ca412d1c0b3dc40d87ac361/x/deterministicgas/types/gas.go#L7)):*

lets discuss on the call 
</blockquote></details>

but this is commonly used library so I believe that it doesn't explode prometheus

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68 and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 127 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJf_0Lg65mOLGxCrntm:-NJngGkF1_6-tcURcqpY:b-vjgxr) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L127)):*

we should either do change I proposed in comment before to change this to be really msgType
or rename variable msgType to msgMethod
</blockquote></details>

that's why it is "msg_method" for now

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 97 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfUqrcAeD-A9XglJQh:-NJniUBcDCKv1Gwwyrwq:b7z1i7n) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L97)):*

benchmarked 
```
BenchmarkSprintfReflect-12      	15879584	        73.12 ns/op
BenchmarkProtoMessageName-12    	21099068	        53.72 ns/op
```

I don't think that 50-75ns adds that much overhead comparing to some other slow parts so I would prefer to proceed with it


</blockquote></details>

Done.

Now metric looks like this:

```
deterministic_gas_factor{msg_type="*types.MsgSubmitProposal",quantile="0.5"} 1.587991714477539
deterministic_gas_factor{msg_type="*types.MsgSubmitProposal",quantile="0.9"} 1.587991714477539
deterministic_gas_factor{msg_type="*types.MsgSubmitProposal",quantile="0.99"} 1.587991714477539
deterministic_gas_factor_sum{msg_type="*types.MsgSubmitProposal"} 12.712847590446472
deterministic_gas_factor_count{msg_type="*types.MsgSubmitProposal"} 8
```
___
*[`x/deterministicgas/types/gas.go` line 127 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJf_0Lg65mOLGxCrntm:-NJniVTi1sV8uVgaJ6Ty:bb6un1j) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L127)):*

that's why it is "msg_method" for now
</blockquote></details>

Done

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfZfl60AIyC_xbCf3f:-NJodbj9CqU8dO0WbN52:b4s5z57) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

Yes, you are right. But it is a matter of question you want to ask and action you want to take based on the answer.
We can't affect the real gas taken by the message, we may only adjust deterministic gas value. That's why, IMO asking question
"how much deterministic gas deviates from real gas?" makes more sense in this case than "how much real gas deviates from deterministic gas". Because deterministic gas is the value we control and may adjust.

Maybe we shouldn't calculate percentages but only absolute values? Then we always get the difference in gas unit and we may take max, min, average and deviation which seems to be exactly what we need.

</blockquote></details>

lets discuss on the call

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @dzmitryhil and @miladz68)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJk4u2O9Bo4kdacPCxZ:-NJpEZkq1Tu08dnNcgw5:b8xhp7g) (no related file):*

This is what is logged by cored, right? But we don't care about this. We care only about what is available to prometheus:

```
al",quantile="0.9"} 1.5969055891036987
deterministic_gas_factor{tx_type="SubmitProposal",quantile="0.99"} 1.5969055891036987
deterministic_gas_factor_sum{tx_type="SubmitProposal"} 6.360880732536316
deterministic_gas_factor_count{tx_type="SubmitProposal"} 4
```

and everything we need is there
</blockquote></details>

resolved on call

___
___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfZfl60AIyC_xbCf3f:-NJpJYG6A-jy4n1zyfqc:b-5qo06v) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

lets discuss on the call
</blockquote></details>

decided on call:
lets use this formula
`(nondetermGas - deterministicGas) / deterministicGas`
___
*[`x/deterministicgas/types/gas.go` line 7 at r2](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJjwQ_RC6G711Ci4xqJ:-NJpJ0R7-NZqj-2jSeWb:byhk757) ([raw file](https://github.com/coreumfoundation/coreum/blob/df1e2a2e1125d3ef54e3c96d89d27e012cc01098/x/deterministicgas/types/gas.go#L7)):*

but this is commonly used library so I believe that it doesn't explode prometheus
</blockquote></details>

resolved on the call

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil and @wojtek-coreum)

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 1 unresolved discussion (waiting on @wojtek-coreum)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 1 unresolved discussion (waiting on @ysv)

___
*[`x/deterministicgas/types/gas.go` line 125 at r1](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJfZfl60AIyC_xbCf3f:-NJtkZOu6kXOptyMzF9I:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/5d42126bdec1c62b8c17b490fdb82f81071ece43/x/deterministicgas/types/gas.go#L125)):*

decided on call:
lets use this formula
`(nondetermGas - deterministicGas) / deterministicGas`
</blockquote></details>

Done.

ysv (changes requested):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @dzmitryhil, @miladz68, and @wojtek-coreum)

___
*[`x/deterministicgas/types/gas.go` line 124 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJu2HM2D4bZhBpGYkHd:-NJu2HM2D4bZhBpGYkHe:bt0f4o8) ([raw file](https://github.com/coreumfoundation/coreum/blob/1d8de556d4dffc692e7ab1c81a93f93ba8349213/x/deterministicgas/types/gas.go#L124)):*
> ```Go
> func reportDeterministicGasMetric(oldCtx, newCtx sdk.Context, gasBefore sdk.Gas, msg interface{}) {
> 	nondeterministicGas := newCtx.GasMeter().GasConsumed()
> 	if nondeterministicGas == 0 {
> ```

now we need this check for `deterministicGas`
on one hand I'm not sure if it could be 0 on the other it is still a nice fuse 
___
*[`x/deterministicgas/types/gas.go` line 132 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJu2hQ5E5CbnPbw3FXR:-NJu2hQ5E5CbnPbw3FXS:b-4c4ywm) ([raw file](https://github.com/coreumfoundation/coreum/blob/1d8de556d4dffc692e7ab1c81a93f93ba8349213/x/deterministicgas/types/gas.go#L132)):*
> ```Go
> 	gasFactor := (float32(nondeterministicGas) - float32(deterministicGas)) / float32(deterministicGas)
> 	metrics.AddSampleWithLabels([]string{"deterministic_gas_factor"}, gasFactor, []metrics.Label{
> 		{Name: "msg_type", Value: reflect.TypeOf(msg).String()},
> ```

minor:
I vote for changing arg `msg` in reportDeterministicGasMetric  to `msgType string` & using `proto.MessageName` instead of reflect because it contains full path to module with all details etc.

full view:
```
reportDeterministicGasMetric(sdkCtx, newSDKCtx, gasBefore, proto.MessageName(req))
```

dzmitryhil (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @miladz68 and @wojtek-coreum)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @miladz68, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 124 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJu2HM2D4bZhBpGYkHd:-NJubSRX4g5bKGSQW-fo:b-896fix) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2dce281d3da845fa0be64c5a014941c210110e2/x/deterministicgas/types/gas.go#L124)):*

now we need this check for `deterministicGas`
on one hand I'm not sure if it could be 0 on the other it is still a nice fuse 
</blockquote></details>

Done.
___
*[`x/deterministicgas/types/gas.go` line 132 at r4](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NJu2hQ5E5CbnPbw3FXR:-NJubr16CIQfIygDj5fl:bb6un1j) ([raw file](https://github.com/coreumfoundation/coreum/blob/a2dce281d3da845fa0be64c5a014941c210110e2/x/deterministicgas/types/gas.go#L132)):*

minor:
I vote for changing arg `msg` in reportDeterministicGasMetric  to `msgType string` & using `proto.MessageName` instead of reflect because it contains full path to module with all details etc.

full view:
```
reportDeterministicGasMetric(sdkCtx, newSDKCtx, gasBefore, proto.MessageName(req))
```
</blockquote></details>

Done

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @miladz68 and @ysv)

dzmitryhil (dismissed):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 1 unresolved discussion (waiting on @miladz68 and @ysv)

ysv (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68)

miladz68 (changes requested):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: all files reviewed, 1 unresolved discussion (waiting on @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NKIj7o1AWBJOrn_pQgQ:-NKIj7o1AWBJOrn_pQgR:beu9be) (no related file):*
All looks good, However I sometimes see `NaN` in metrics, message wojtek in slack. 

___

miladz68 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NKIj7o1AWBJOrn_pQgQ:-NKOFE9t9OpqhR9ysvSa:b-ud5fr4) (no related file):*

All looks good, However I sometimes see `NaN` in metrics, message wojtek in slack. 
</blockquote></details>

Approved on my side. 

___

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NKIj7o1AWBJOrn_pQgQ:-NKXKjHH4bVT8whxsNCd:bsbbj0m) (no related file):*

Approved on my side. 
</blockquote></details>

Changed the formula

___

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @silverspase)

miladz68 (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @silverspase)

*[a discussion](https://reviewable.io/reviews/coreumfoundation/coreum/294#-NKIj7o1AWBJOrn_pQgQ:-NKmRFZz3zQt0DCZ-5Wc:b-mnnl2i) (no related file):*

Changed the formula
</blockquote></details>

great, but I don't think the buckets will be as evenly distributed as the sample we saw in tendermint example.

___

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/coreumfoundation/coreum/294)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @silverspase)

PR #362, opened 2023-01-25
Finalized `pkg/client`, extracted config

wojtek-coreum:
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)


ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcoKKC8hDNlc_J3ob6:-NMcoKKD1yIBfHfaJnJj:bm86x01) (no related file):*
we discussed this a while ago and I think it is time to rename pkg/tx to pkg/client

Not sure if this is needed: Also we may consider restructuring it in a similar way as it is done in cosmos SDK
pkg/client/context.go & pkg/client/tx/tx.go

___
*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcwNNKD_8nZ2bbWjcW:-NMcwNNKD_8nZ2bbWjcX:bye5b4v) (no related file):*
I have an idea
Instead of just writing a doc for client usage we should write an example of usage and commit it to repo
And then in doc we will either embed it or copy. In this way we will be able to run in from repo & will know if we brake compatibility. Maybe we should even include it in unit tests

I mean smth like this:

```golang
func ExampleCoreumClientBankSend() {
 ...
}
```
and then we can use the same code in doc

___
___
*[`pkg/tx/context.go` line 3 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcsSTV0Bb_4OgiMt5e:-NMcsSTV0Bb_4OgiMt5f:b-dredgk) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L3)):*
> ```Go
> package tx
> 
> import (
> ```

I think we should add top level package comment to describe in general modifications we did in our client utils, why we need them etc

So devs know what to expect from our client
___
*[`pkg/tx/context.go` line 32 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcrSqkDbqkXZx0FULZ:-NMcrSqkDbqkXZx0FUL_:b-ezjuxt) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L32)):*
> ```Go
> 
> // Config stores context config
> type Config struct {
> ```

Config is too generic name here I propose to be more specific:
- RequestConfig
- TimeoutConfig
- HTTPConfig
- smth else ?
___
*[`pkg/tx/context.go` line 72 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMctdK732AqQMZAnYHw:-NMctdK732AqQMZAnYHx:b-16dnev) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L72)):*
> ```Go
> 
> // WithConfig returns a copy of the context with an updated config
> func (c ClientContext) WithConfig(config Config) ClientContext {
> ```

@wojtek-coreum @dzmitryhil @miladz68 should I think we should make gasPriceMultiplier configurable on ClientContext level ? Or somewhere else ? What is the current value ?
___
*[`pkg/tx/context.go` line 83 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcqB0bDcSrjvTskGh5:-NMcqB0bDcSrjvTskGh6:b1fgucg) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L83)):*
> ```Go
> 
> // WithChainID returns a copy of the context with an updated chain ID
> func (c ClientContext) WithChainID(chainID string) ClientContext {
> ```

can you please go through cosmos client.Context and verify that we didn't skip any `With...`  & getter methods

github.com/cosmos/cosmos-sdk@v0.45.11/client/context.go


___
*[`pkg/tx/tx.go` line 20 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcxa5UDf0GnOzgLDFY:-NMcxa5UDf0GnOzgLDFZ:b4g7p7x) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/tx.go#L20)):*
> ```Go
> 	tmtypes "github.com/tendermint/tendermint/types"
> 
> 	"github.com/CoreumFoundation/coreum-tools/pkg/retry"
> ```

inside retry we depend on logger (mostly our logger)
IMO we should either remove dependency on logger or make it optional so users of client can proceed with their own context

```
func Do(ctx context.Context, retryAfter time.Duration, fn func() error) error {
	log := logger.Get(ctx)
...
```

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 8 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMczCr_8KAtgb0kkDCv:-NMczCr_8KAtgb0kkDCw:b-vgrvye) (no related file):*
After we finalize structure I want @silverspase to review it also
Keeping this unresolved for now

___

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 10 unresolved discussions (waiting on @miladz68, @silverspase, and @ysv)

___
*[`pkg/tx/context.go` line 72 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMctdK732AqQMZAnYHw:-NMd3En05qD10Q-Cj6Vw:b-svej0e) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L72)):*

Right, let's set the default in the config = `1.1` since `1` might not work by default. 
</blockquote></details>

where is the gasPriceMultiplier defined at the moment?

miladz68 (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 10 unresolved discussions (waiting on @silverspase, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 10 unresolved discussions (waiting on @silverspase and @ysv)

___
*[`pkg/tx/tx.go` line 20 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcxa5UDf0GnOzgLDFY:-NMd7hPX1XY8sft29zVl:b-3q7vg4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/tx.go#L20)):*

IMO we don't need that logger at all :-) It just spoils the output with the retries.
If we want it, then it should be the interface.
</blockquote></details>

And logger is also used inside `retry`. We may remove it from there as well

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 9 unresolved discussions (waiting on @dzmitryhil, @silverspase, and @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcoKKC8hDNlc_J3ob6:-NMddoeIDnCI-v72OVyQ:b-shv99m) (no related file):*

we discussed this a while ago and I think it is time to rename pkg/tx to pkg/client

Not sure if this is needed: Also we may consider restructuring it in a similar way as it is done in cosmos SDK
pkg/client/context.go & pkg/client/tx/tx.go
</blockquote></details>

on call decided:

`pkg/client`

___
___
*[`pkg/tx/context.go` line 32 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcrSqkDbqkXZx0FULZ:-NMdgnbh65uucNMVSqRQ:b1zsomv) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L32)):*

How about `ClientContextConfig` and `DefaultClientContextConfig` for the func, since it's the `ClientContext` config.
</blockquote></details>

ClientContext -> Context
Config -> ContextConfig
___
*[`pkg/tx/context.go` line 52 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMd0Bag7uHN4lA_j3h5:-NMdhGjG1DDDxQwhP1Cf:by62yd7) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L52)):*

I would pass the config as arg here, and use the `Default` in the caller.
</blockquote></details>

add clientConfig as a second arg
withConfig is not needed then
___
*[`pkg/tx/context.go` line 72 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMctdK732AqQMZAnYHw:-NMdjimN0PrrW1RDci_a:befluu0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L72)):*

where is the gasPriceMultiplier defined at the moment?
</blockquote></details>

Implement `withGasPriceAdjustment` & use 1.1 by default
___
*[`pkg/tx/tx.go` line 20 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcxa5UDf0GnOzgLDFY:-NMdkNFDAXaZlcr-Fdee:b-ags46x) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/tx.go#L20)):*

And logger is also used inside `retry`. We may remove it from there as well
</blockquote></details>

remove logger

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: 0 of 27 files reviewed, 9 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcoKKC8hDNlc_J3ob6:-NMhP_YiEVFI1IFtCCnG:b-896fix) (no related file):*

on call decided:

`pkg/client`
</blockquote></details>

Done.

___
___
*[`pkg/tx/context.go` line 3 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcsSTV0Bb_4OgiMt5e:-NMhMChU3HZWaNzawyih:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L3)):*

I think we should add top level package comment to describe in general modifications we did in our client utils, why we need them etc

So devs know what to expect from our client
</blockquote></details>

Done.
___
*[`pkg/tx/context.go` line 32 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcrSqkDbqkXZx0FULZ:-NMhGK4BEACwgHTWbWpJ:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L32)):*

ClientContext -> Context
Config -> ContextConfig
</blockquote></details>

Done.
___
*[`pkg/tx/context.go` line 52 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMd0Bag7uHN4lA_j3h5:-NMhGeuA65nm2FqRNaF_:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L52)):*

add clientConfig as a second arg
withConfig is not needed then
</blockquote></details>

Done.
___
*[`pkg/tx/context.go` line 72 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMd0fuJCfq5aUEZ4G3p:-NMhGjf_D5NXWnO0Qk_C:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L72)):*

I would propose the use the config as input in the constructor because in that case, the client will pay attention that we have defaults and what they are.
</blockquote></details>

Done.
___
*[`pkg/tx/context.go` line 83 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcqB0bDcSrjvTskGh5:-NMhJY-UDDyFXlMlfwiw:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L83)):*

can you please go through cosmos client.Context and verify that we didn't skip any `With...`  & getter methods

github.com/cosmos/cosmos-sdk@v0.45.11/client/context.go



</blockquote></details>

Done.
___
*[`pkg/tx/tx.go` line 20 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMcxa5UDf0GnOzgLDFY:-NMhJZuVASZgZJ7PbvXE:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/tx.go#L20)):*

remove logger
</blockquote></details>

Done.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: 0 of 27 files reviewed, 9 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`pkg/tx/context.go` line 72 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMctdK732AqQMZAnYHw:-NMhlhdj11fAnj09Sw3T:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/03da44bf9d7229255ad1740f2885adfb98e3db76/pkg/tx/context.go#L72)):*

Implement `withGasPriceAdjustment` & use 1.1 by default
</blockquote></details>

Done.

ysv (changes requested):
> Reviewed 25 of 27 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 8 unresolved discussions (waiting on @dzmitryhil, @silverspase, @wojtek-coreum, and @ysv)

___
*[`app/app.go` line 270 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk7PfFEAEAP4gfOp37:-NMk7PfFEAEAP4gfOp38:bsl5lqj) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/app/app.go#L270)):*
> ```Go
> 	baseAppOptions ...func(*baseapp.BaseApp),
> ) *App {
> 	deterministicGasRequirements := deterministicgas.DefaultConfig()
> ```

rebase on master ?
___
*[`integration-tests/chain.go` line 173 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMkE7Py6Hbm_FNd2eAZ:-NMkE7Py6Hbm_FNd2eA_:b-hl3ftd) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/integration-tests/chain.go#L173)):*
> ```Go
> 		panic(err)
> 	}
> 	clientCtx := client.NewContext(client.DefaultContextConfig(), app.ModuleBasics).
> ```

so do we use default multiplier of 1.1 in integration-tests also ?
___
*[`pkg/client/tx.go` line 4 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk8XNe3PpdzNp-9Fzs:-NMk8XNe3PpdzNp-9Fzt:b-bid2j2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L4)):*
> ```Go
> 
> // This file contains helper functions used to prepare and broadcast transactions.
> // Blocking broadcast was reimplemented to use polling instead of subscription to eliminate the case when
> ```

minor:
"Blocking broadcast mode was reimplemented to use polling instead of subscription to eliminate the case when transaction execution is missed due to broken websocket connection. For other broadcast modes we just call original Cosmos implementation.
For more details check BroadcastRawTx & broadcastTxCommit."
___
*[`pkg/client/tx.go` line 41 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk9_LEBLDHE0MWPjcv:-NMk9_LEBLDHE0MWPjcw:b-354iv9) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L41)):*
> ```Go
> var Sign = tx.Sign
> 
> var defaultGasPriceAdjustment = sdk.MustNewDecFromStr("1.1")
> ```

IMO this value should live in `pkg/client/context.go` and be set by default in NewContext function.
This value is like our request timeouts so lets group all of them there
___
*[`pkg/client/tx.go` line 64 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMkAxSg1g3P-xHVPkIf:-NMkAxSh9LO4ZngNA44G:bypuoxh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L64)):*
> ```Go
> 			gasPriceAdjustment = defaultGasPriceAdjustment
> 		}
> 		gasPrice.Amount.Mul(gasPriceAdjustment)
> ```

the problem here is that by blindly multiplying by this value we may go over `InitialGasPrice` which doesn't make sense in all cases except fee escalation
I'm not insisting on implementing this now but it might be beneficial to handle this.

Maybe request QueryMinGasPriceRequest should also return InitialGasPrice ?

Should leave todo or here ? WDYT ?

ysv (changes requested):
Reword commit name pls since it is not pkg/tx anymore

> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-:-NMkE_K28uEnRJ39-212:b-22cbm9)* status: all files reviewed, 8 unresolved discussions (waiting on @dzmitryhil, @silverspase, @wojtek-coreum, and @ysv)

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @silverspase, and @ysv)

___
*[`app/app.go` line 270 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk7PfFEAEAP4gfOp37:-NMm2aUaAL9GRd88bKO4:bk5212u) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/app/app.go#L270)):*

rebase on master ?
</blockquote></details>

WDYM? All the changes from master are here
___
*[`integration-tests/chain.go` line 173 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMkE7Py6Hbm_FNd2eAZ:-NMm2oFE26r0Yi8SR4fB:bt9u0k5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/integration-tests/chain.go#L173)):*

so do we use default multiplier of 1.1 in integration-tests also ?
</blockquote></details>

It depends on the test. Those who set `WithSimulateAndExecute(true)` (non-deterministic tests, testing WASM), use that multiplier. Others don't and they can't because it would break the balance calculations.
___
*[`pkg/client/tx.go` line 41 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk9_LEBLDHE0MWPjcv:-NMm3ZP686uF4vwJ6Qob:bdow49z) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L41)):*

IMO this value should live in `pkg/client/context.go` and be set by default in NewContext function.
This value is like our request timeouts so lets group all of them there


</blockquote></details>

I disagree. Standard client context does not set any magic values by default. Moreover, we don't do that with `WithGasAdjustment` either - we use default value only during gas estimation, and I follow the same pattern here.
___
*[`pkg/client/tx.go` line 64 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMkAxSg1g3P-xHVPkIf:-NMm4-Dy8oQ0a40CqtoC:b8ec3l8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L64)):*

the problem here is that by blindly multiplying by this value we may go over `InitialGasPrice` which doesn't make sense in all cases except fee escalation
I'm not insisting on implementing this now but it might be beneficial to handle this.

Maybe request QueryMinGasPriceRequest should also return InitialGasPrice ?

Should leave todo or here ? WDYT ?

</blockquote></details>

That query would have to return information about the region the price is in. Because in case of escalation region price should go above the initial price, but in others it should not. If we want to be smarter here, it should be a separate task as it requires changes in other component.

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: all files reviewed, 3 unresolved discussions (waiting on @silverspase and @wojtek-coreum)

___
*[`app/app.go` line 270 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk7PfFEAEAP4gfOp37:-NMmYjFA0WWJVDeRzAM0:b-oe5tl2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/app/app.go#L270)):*

WDYM? All the changes from master are here
</blockquote></details>

I still see gasRequirements while it should be degerministicGasConfig

There are still a lot of changes shown in diff which seems to be related to previous PR not this one
E.g: 


I'm wondering why they are here
___
*[`integration-tests/chain.go` line 173 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMkE7Py6Hbm_FNd2eAZ:-NMmeYkqBuFepsl0OYDV:b32wtd) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/integration-tests/chain.go#L173)):*

It depends on the test. Those who set `WithSimulateAndExecute(true)` (non-deterministic tests, testing WASM), use that multiplier. Others don't and they can't because it would break the balance calculations.
</blockquote></details>

ok
___
*[`pkg/client/tx.go` line 41 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk9_LEBLDHE0MWPjcv:-NMmc0BK1b33pcKtGB6I:bj8lrio) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L41)):*

I disagree. Standard client context does not set any magic values by default. Moreover, we don't do that with `WithGasAdjustment` either - we use default value only during gas estimation, and I follow the same pattern here.
</blockquote></details>

But we already do set magic default values in Context which are here:

```
func DefaultContextConfig() ContextConfig {
	return ContextConfig{
		RequestTimeout:           10 * time.Second,
		TxTimeout:                time.Minute,
		TxStatusPollInterval:     500 * time.Millisecond,
		TxNextBlocksTimeout:      time.Minute,
		TxNextBlocksPollInterval: time.Second,
	}
}
```

IMO we should add all default there so users can see what is set by default in context in a single place instead of looking for default value of `GasAdjustment` in `func CalculateGas` & for default for `GasPriceAdjustment` value in tx.go in `var defaultGasPriceAdjustment`

What is more I see that in original cosmos tx they don't set this magical value in `func CalculateGas` which I agree with. And if this value is not set it will error. So I propose to set all of them explicitly in DefaultContextConfig since we added this config but do not spread all over the code
```
func DefaultContextConfig() ContextConfig {
	return ContextConfig{
               TimeoutsConfig {
		 RequestTimeout:           10 * time.Second,
		 TxTimeout:                time.Minute,
		 TxStatusPollInterval:     500 * time.Millisecond,
		 TxNextBlocksTimeout:      time.Minute,
		 TxNextBlocksPollInterval: time.Second,
              },
            OtherConfig {
              GasPriceAdjustment: 1.1,
              GasAdjustment: 1.0,
              ...
            }
	}
}
```
___
*[`pkg/client/tx.go` line 64 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMkAxSg1g3P-xHVPkIf:-NMmdM5l4RagVWcz_fF-:b-emx108) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L64)):*

That query would have to return information about the region the price is in. Because in case of escalation region price should go above the initial price, but in others it should not. If we want to be smarter here, it should be a separate task as it requires changes in other component.
</blockquote></details>

yes of course not in a scope of this PR
Will add it to clickup & discuss later

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: 18 of 27 files reviewed, 3 unresolved discussions (waiting on @silverspase and @ysv)

___
*[`app/app.go` line 270 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk7PfFEAEAP4gfOp37:-NMmhukN2KZnzkIZSte5:bkdtil2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/app/app.go#L270)):*

I still see gasRequirements while it should be degerministicGasConfig

There are still a lot of changes shown in diff which seems to be related to previous PR not this one
E.g: 


I'm wondering why they are here
</blockquote></details>

Yes, yesterday I created a mess by pushing wrong commits to wrong PRs. It should be fine now.
___
*[`pkg/client/tx.go` line 41 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/362#-NMk9_LEBLDHE0MWPjcv:-NMmiCT34vnfBYqoGAXl:bh5pz36) ([raw file](https://github.com/CoreumFoundation/coreum/blob/374961474cde334044177eb28ef58df6dc4d8836/pkg/client/tx.go#L41)):*
> But we already do set magic default values in Context which are here:

Yes, because those variables require some values.
On the other hand gasPriceAdjustment does not. If it is not set and user does not use estimation it should use the exact price requested by the user. Otherwise it will  lead to unpredictable behavior. If you insist we may set it to 1.0 which is equivalent to not setting it.

We may do the same with gasAdjustment, so if we want to set defaults explicitly, then both should be set to 1.0

ysv (dismissed):
> Reviewed 9 of 9 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @silverspase)

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum)

wojtek-coreum (commented):
> Reviewed 16 of 27 files at r2, 2 of 9 files at r5, 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum)

wojtek-coreum (commented):
> Reviewed 7 of 9 files at r5.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/362)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum)

PR #382, opened 2023-02-08
Wait for one block in gas_price_test

miladz68:
the test fails intermittently because sometimes although the tx.BroadcastCommit returns and it means that the tx has been included in a block, querying immediate for the results returns not found error. we need to wait for one block so the state is fully synced.

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/382)


wojtek-coreum (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/382)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dzmitryhil, @silverspase, and @ysv)

PR #436, opened 2023-03-10
check there are no duplicate accounts in funded accounts list

miladz68:
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/436)

### DIFF
--- a/pkg/config/network.go
+++ b/pkg/config/network.go
@@ -92,6 +92,9 @@ func init() {
 	// where 4 * 100_000_000 is a balance of each of foundation-{1-4} wallets.
 	mainGenesisValidatorsCreatorBalance := sdk.NewCoins(sdk.NewCoin(constant.DenomMain, sdk.NewInt(120_000_000_000)))
 
+	mainFoundationZeroInitialBalance := sdk.NewCoins(sdk.NewCoin(constant.DenomMain, sdk.NewInt(99_880_000_000_000)))
+	mainFoundationOtherInitialBalance := sdk.NewCoins(sdk.NewCoin(constant.DenomMain, sdk.NewInt(100_000_000_000_000)))
+
 	// testnet vars
 
 	// TODO: Add test that total supply (sum of amounts funded) is always 500M.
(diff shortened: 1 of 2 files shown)


ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/436)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @miladz68, @silverspase, and @wojtek-coreum)

___
*[`pkg/config/network.go` line 444 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/436#-NQAEXx-5KZXwRxbXUyb:-NQAEXx-5KZXwRxbXUyc:bdwymb6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/530c94d5b1814f5b5951bbd33e9d22a43d5ec83a/pkg/config/network.go#L444)):*
> ```Go
> }
> 
> func MergeDuplicateFundedAccounts(accounts []FundedAccount) []FundedAccount {
> ```

I disagree with this approach I think we should proceed in a different way 

option 1:
Instead of initializing FundedAccounts in Network directly as array we can use function `func (n *Network) FundAccount` and add append them there one by one & during appending validate that there are no duplicates otherwise error. By the way we have this function but it is not used so if you don't proceed with this approach pls remove it

option 2:
We should merge duplicated funded accounts because if we have some duplicated address it means that we made a mistake. IMO we can change MergeDuplicateFundedAccounts -> validateFundedAccounts and return error if duplicate exists

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/436)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @silverspase, @wojtek-coreum, and @ysv)

___
*[`pkg/config/network.go` line 444 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/436#-NQAEXx-5KZXwRxbXUyb:-NQAG2ujAkME46OBKHB-:b-mw3gf5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/530c94d5b1814f5b5951bbd33e9d22a43d5ec83a/pkg/config/network.go#L444)):*

I disagree with this approach I think we should proceed in a different way 

option 1:
Instead of initializing FundedAccounts in Network directly as array we can use function `func (n *Network) FundAccount` and add append them there one by one & during appending validate that there are no duplicates otherwise error. By the way we have this function but it is not used so if you don't proceed with this approach pls remove it

option 2:
We should merge duplicated funded accounts because if we have some duplicated address it means that we made a mistake. IMO we can change MergeDuplicateFundedAccounts -> validateFundedAccounts and return error if duplicate exists
</blockquote></details>

the issue is that we have duplicate entries for mainnet, and I did not want to touch them. how should we approach that ?

ysv (dismissed):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/436)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dzmitryhil, @silverspase, and @wojtek-coreum)

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/436)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @silverspase and @wojtek-coreum)

PR #451, opened 2023-03-17
aligned spec with docs

vertex451:
When I moved the spec into docs as it is, it looked bad on UI, so I did some changes, and I would like to have the same files in both repositories, otherwise, it would be hard to support two different sets of files.
Approve it, please.

This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/451)


wojtek-coreum (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/451)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68 and @ysv)

PR #454, opened 2023-03-21
Makefile - simplify `cored` binary build

sryps:
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/454)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,64 @@
+#!/usr/bin/make -f
+
+BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
+COMMIT := $(shell git log -1 --format='%H')
+
+VERSION := $(shell git describe --tags)
+
+LEDGER_ENABLED ?= true
+TM_VERSION := $(shell go list -m github.com/tendermint/tendermint | sed 's:.* ::') # grab everything after the space in "github.com/tendermint/tendermint v0.34.7"
+BUILDDIR ?= $(CURDIR)/build
+
+GO_SYSTEM_VERSION = $(shell go version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f1-2)
+REQUIRE_GO_VERSION = 1.20
+
+export GO111MODULE = on
+
+
+build_tags = netgo
+build_tags += ledger
+
+build_tags += $(BUILD_TAGS)
+build_tags := $(strip $(build_tags))
+
+whitespace :=
+whitespace += $(whitespace)
+comma := ,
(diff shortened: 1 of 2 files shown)


wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/454)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @miladz68, @silverspase, @sryps, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/454#-NR55yBP9Kyy_KiDfMHQ:-NR55yBP9Kyy_KiDfMHR:b6a54zy) (no related file):*
Hi,

We already have building toolkit. Please check the repo: https://github.com/CoreumFoundation/crust.
You just checkout that and run `crust build/cored`.

___

greg-szabo:
I'm sad to see this closed. The build kit relies on Docker which makes building really inconvenient (and hard) if your own build system is also in Docker. (And it's in a different repository for some reason?)

Most of the Cosmos ecosystem (starting with the Cosmos SDK and CometBFT) have a simple `Makefile` that allows developers and especially validators build a local version of the source code. `make build` simply works in a regular case.

The build kit might be great for developers but validators usually have their own automation for building binaries, simply because they have to deal with multiple networks. The closer each build system is to each other, the easier life is for validators.

Looking at this change, it looked like the standard `Makefile` available at any other chain. I don't think it would hurt if it was here, it might even simplify the build kit somewhat.

wojtek-coreum:
1. Validators may download ready to use binary from our releases. They should not use any other version.
2. Developers may use toolkit we provide or create their own helper (at the end of the day it's just a single `go build` command, it's so trivial that it doesn't deserve Makefile at all).

greg-szabo:
Au contraire, most of the ecosystem encourages validators to build from the source code because that way they can check what they are running.

Tomorrow, there's a Cosmos Hub upgrade and a bunch of the big validators are waiting for the source code so they can review before they build. They've done this for years.

This PR came from a validator too, who runs cored (as well as other networks) from source. They did the PR to help the network and its validators.

And I'm building binaries from source within a "security incident response" effort. I'm going to prioritize chains that can be built via `make build` (or maybe `make install`) simply because I want to focus on the reports and not on build processes. 

There's over a 100 chains that we want to support, asking for a `Makefile` that was available for 5+ years already is not an unreasonable request. Especially, if the work is already done.

PR #457, opened 2023-03-24
Added functions to change gas adjustment on client context

miladz68:
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/457)

### DIFF
--- a/pkg/client/context.go
+++ b/pkg/client/context.go
@@ -119,8 +119,8 @@ func (c Context) GasPriceAdjustment() sdk.Dec {
 }
 
 // WithGasAdjustment returns context with new gas adjustment.
-func (c Context) WithGasAdjustment(adj sdk.Dec) Context {
-	c.config.GasConfig.GasPriceAdjustment = adj
+func (c Context) WithGasAdjustment(adj float64) Context {
+	c.config.GasConfig.GasAdjustment = adj
 	return c
 }
 
--- a/pkg/client/context.go
+++ b/pkg/client/context.go
@@ -118,6 +118,18 @@ func (c Context) GasPriceAdjustment() sdk.Dec {
 	return c.config.GasConfig.GasPriceAdjustment
 }
 
+// WithGasAdjustment returns context with new gas adjustment.
+func (c Context) WithGasAdjustment(adj sdk.Dec) Context {
+	c.config.GasConfig.GasPriceAdjustment = adj
+	return c
+}
+
+// WithGasPriceAdjustment returns context with new gas price adjustment.
+func (c Context) WithGasPriceAdjustment(adj sdk.Dec) Context {
+	c.config.GasConfig.GasPriceAdjustment = adj
+	return c
+}
+
 // WithRPCClient returns a copy of the context with an updated RPC client
 // instance.
 func (c Context) WithRPCClient(client rpcclient.Client) Context {


wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/457)* status: all files reviewed, 2 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/457#-NRJWD8cCXXd_-8yMdvg:-NRJWD8d8BNGt2G9ksB5:b-7id5aq) (no related file):*
When gas adjustment and gas price adjustment were added initialy some time ago, it was done like in this PR. But during the discussion IIRC @ysv and @dzmitryhil argued that it's better to set them in constructor as a config instead of adding those `With*` methods. Did the situation change that now you find it to be better to set it this way instead?

___

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/457)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase and @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/457#-NRJWD8cCXXd_-8yMdvg:-NRYqY34CuPZnAkxSLjG:b-2zn3cw) (no related file):*

When gas adjustment and gas price adjustment were added initialy some time ago, it was done like in this PR. But during the discussion IIRC @ysv and @dzmitryhil argued that it's better to set them in constructor as a config instead of adding those `With*` methods. Did the situation change that now you find it to be better to set it this way instead?
</blockquote></details>

I recall this discussion but don't remember why we decided to proceed in one or another way

I'm okay to keep this param required inside constructor but also have this setter 

___

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/457)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/457#-NRJWD8cCXXd_-8yMdvg:-NR_ckUb6QkcRpZ9HVfA:b-bohsg8) (no related file):*

I recall this discussion but don't remember why we decided to proceed in one or another way

I'm okay to keep this param required inside constructor but also have this setter 
</blockquote></details>

At that time you said that those params are so important that it's better to require them explicitely in the constructor. Setting them both in constructor and in `With*` is a bit redundant. If we want to go with `With*` then let's remove the constructor.

___

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/457)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase and @wojtek-coreum)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/457#-NRJWD8cCXXd_-8yMdvg:-NRbwZSp4frZaeXRYyO7:b-361ao7) (no related file):*

At that time you said that those params are so important that it's better to require them explicitely in the constructor. Setting them both in constructor and in `With*` is a bit redundant. If we want to go with `With*` then let's remove the constructor.
</blockquote></details>

they should be passed inside constructor because setting them to 0 doesn't make sense since it is obviously a wrong value which will not work.

Having them inside With is beneficial because then it gives ability to use DefaultCtx and override a single param using `With*`

```
ctx := DefaultContextConfig().WithGasAdjustment(1.2)
```
I think it is a reasonable case

___

PR #465, opened 2023-04-13
Upgrade cosmos sdk to v0.47.1 and wasm to v0.40.0-rc.

dzmitryhil:
* Update go.mod dependencies
* Regenerate protos
* Update wrapped modules
* Update app.go to be compatible with new SDK
* Remove the Block tx mode and add WithAwaitTx option

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)


wojtek-coreum (changes requested):
> Reviewed 1 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 17 of 96 files reviewed, 12 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`cmd/cored/main.go` line 16 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrJkVAVAhTiLFoBAX:-NSyrJkVAVAhTiLFoBAY:b-k6xr2z) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L16)):*
> ```Go
> 
> func main() {
> 	logger := log.NewNopLogger()
> ```

Why? Don't we like logs?
___
*[`cmd/cored/main.go` line 28 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrZtT3nM33-22h8by:-NSyrZtU46mT9SKQt_oG:bq3gyu7) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L28)):*
> ```Go
> 	cosmoscmd.OverwriteDefaultChainIDFlags(rootCmd)
> 	rootCmd.PersistentFlags().String(flags.FlagChainID, string(app.DefaultChainID), "The network chain ID")
> 	if err := svrcmd.Execute(rootCmd, "", app.DefaultNodeHome); err != nil {
> ```

I think we should use `CORED` for env prefix

wojtek-coreum (changes requested):
> Reviewed 1 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 18 of 96 files reviewed, 16 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`app/export.go` line 38 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSys-z5-o2xb-VmZ-4r:-NSys-z5-o2xb-VmZ-4s:bse1l1m) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/app/export.go#L38)):*
> ```Go
> 
> 	validators, err := staking.WriteValidators(ctx, app.StakingKeeper)
> 	if err != nil {
> ```

why removed? it was a good idea to return empty object in case of error
___
*[`app/export.go` line 80 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSys9SN4UwyuhPcio_D:-NSys9SN4UwyuhPcio_E:borsaf0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/app/export.go#L80)):*
> ```Go
> 	app.StakingKeeper.IterateValidators(ctx, func(_ int64, val stakingtypes.ValidatorI) (stop bool) {
> 		_, err := app.DistrKeeper.WithdrawValidatorCommission(ctx, val.GetOperator())
> 		if err != nil {
> ```

why removed?
___
*[`app/export.go` line 88 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysGzz9YK0sD9eqqRU:-NSysGzz9YK0sD9eqqRV:b-4vcodm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L88)):*
> ```Go
> 		delAddr := sdk.MustAccAddressFromBech32(delegation.DelegatorAddress)
> 
> 		_, _ = app.DistrKeeper.WithdrawDelegationRewards(ctx, delAddr, valAddr)
> ```

error is not handled
___
*[`app/export.go` line 180 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysmVD47VAOEhHq8Pk:-NSysmVEDIrah2Ktx0iT:ba74aa8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L180)):*
> ```Go
> 
> 	if err := iter.Close(); err != nil {
> 		app.Logger().Error("error while closing the key-value store reverse prefix iterator: ", err)
> ```

It seems like a case for `log.Fatal`

wojtek-coreum (changes requested):
> Reviewed 1 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 19 of 96 files reviewed, 17 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`cmd/cored/cosmoscmd/init.go` line 1 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytGqg12Jk16Zc1pKx:-NSytGqh1-ZT_P07dRIx:b-wtmsw0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/cmd/cored/cosmoscmd/init.go#L1)):*
> ```Go
> package cosmoscmd
> ```

why removed? How do we initialize genesis now?

wojtek-coreum (changes requested):
> Reviewed 1 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 21 of 96 files reviewed, 18 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`cmd/cored/cosmoscmd/root.go` line 53 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytjoeAsrzwxqrGsuf:-NSytjof0fSFo8Vcztom:b-adpcmv) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/cosmoscmd/root.go#L53)):*
> ```Go
> 		WithAccountRetriever(types.AccountRetriever{}).
> 		WithHomeDir(app.DefaultNodeHome).
> 		WithViper("")
> ```

`CORED`?

wojtek-coreum (changes requested):
> Reviewed 3 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 24 of 96 files reviewed, 19 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`pkg/client/tx.go` line 138 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyx5DkAHHyvL994H8-:-NSyx5DkAHHyvL994H80:bd7i548) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/pkg/client/tx.go#L138)):*
> ```Go
> 
> 	case flags.BroadcastAsync:
> 		txRes, err = broadcastTxAsync(ctx, clientCtx, txBytes)
> ```

Does it make sense to await tx in async mode? By using this mode you want to execute the tx **asynchronously**.

wojtek-coreum (commented):
> Reviewed 6 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 30 of 96 files reviewed, 19 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (commented):
> Reviewed 6 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 36 of 96 files reviewed, 19 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (changes requested):
> Reviewed 4 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 40 of 96 files reviewed, 20 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`pkg/config/genesis/genesis.tmpl.json` line 138 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz0nY62SdKs9FYLkCb:-NSz0nY62SdKs9FYLkCc:b-u41qhw) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/pkg/config/genesis/genesis.tmpl.json#L138)):*
> ```JSON
>         "threshold": "0.500000000000000000",
>         "veto_threshold": "0.334000000000000000",
>         "min_initial_deposit_ratio": "0.000000000000000000",
> ```

Do we understand these params?

wojtek-coreum (commented):
> Reviewed 2 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 42 of 96 files reviewed, 20 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (commented):
> Reviewed 2 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 44 of 96 files reviewed, 20 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (changes requested):
> Reviewed 3 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 47 of 96 files reviewed, 23 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`integration-tests/chain.go` line 23 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz4qTz4x84dEIc-Wb3:-NSz4qTz4x84dEIc-Wb4:b-7s8lnm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/chain.go#L23)):*
> ```Go
> // ChainContext is a types used to store the components required for the test chains subcomponents.
> type ChainContext struct {
> 	EncodingConfig         config.EncodingConfig
> ```

I think what `ChainContext` really depends on is the codec, used in `GenAccount()`, not entre encoding config
___
*[`integration-tests/chain.go` line 73 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz4WA04zLICYCdWrQN:-NSz4WA04zLICYCdWrQO:b-ef3890) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/chain.go#L73)):*
> ```Go
> 	address, err := keyInfo.GetAddress()
> 	if err != nil {
> 		panic(err)
> ```

I think we decided to return errors in tests. But might this operation ever fail?
___
*[`integration-tests/gov.go` line 205 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz5WDw7OzikmACGtfq:-NSz5WDw7OzikmACGtfr:b-6sijyh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/gov.go#L205)):*
> ```Go
> 
> 		clientCtx := g.chainCtx.ClientContext.
> 			WithBroadcastMode(flags.BroadcastSync).
> ```

Isn't this and await already there in client context by default?

wojtek-coreum (commented):
> Reviewed 11 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 58 of 96 files reviewed, 23 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (changes requested):
> Reviewed 1 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 59 of 96 files reviewed, 24 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`x/auth/ante/ante.go` line 98 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzEk9WAEuXdh0_8zZi:-NSzEk9WAEuXdh0_8zZj:b-2b3j4f) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/x/auth/ante/ante.go#L98)):*
> ```Go
> 		deterministicgasante.NewSetInfiniteGasMeterDecorator(options.DeterministicGasConfig),
> 		authante.NewRejectExtensionOptionsDecorator(),
> 		NewDenyMessagesDecorator(&crisistypes.MsgVerifyInvariant{}),
> ```

Why is this one removed? Does invariant verification tx work now? We have integration test checking that this message is denied.

wojtek-coreum (commented):
> Reviewed 4 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 63 of 96 files reviewed, 24 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (changes requested):
> Reviewed 7 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 70 of 96 files reviewed, 25 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`x/feemodel/ante/ante.go` line 30 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzGIHp-UUFCiqQDZDN:-NSzGIHp-UUFCiqQDZDO:btknqlr) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/x/feemodel/ante/ante.go#L30)):*
> ```Go
> // AnteHandle handles transaction in ante decorator.
> func (fd FeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
> 	// FIXME(v47-simulation): check that we need that additional check here `ctx.ChainID() == "simulation"`
> ```

You found that we need it, right?

wojtek-coreum (commented):
> Reviewed 6 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 76 of 96 files reviewed, 25 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (changes requested):
> Reviewed 5 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 81 of 96 files reviewed, 27 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`x/wbank/module.go` line 46 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzHQRc3CqKZDxUi0Gi:-NSzHQRc3CqKZDxUi0Gj:b-tx8ahl) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/x/wbank/module.go#L46)):*
> ```Go
> 	m := bankkeeper.NewMigrator(am.keeper.BaseKeeper, am.legacySubspace)
> 	if err := cfg.RegisterMigration(banktypes.ModuleName, 1, m.Migrate1to2); err != nil {
> 		panic(fmt.Sprintf("failed to migrate x/bank from version 1 to 2: %v", err))
> ```

Here and below: `errors.Errorf` to get full stack trace
___
*[`x/wstaking/module.go` line 55 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzHlMV9iMeuFMrRPe9:-NSzHlMV9iMeuFMrRPeA:b2nq8vw) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/x/wstaking/module.go#L55)):*
> ```Go
> 	m := stakingkeeper.NewMigrator(am.stakingKeeper, am.legacySubspace)
> 	if err := cfg.RegisterMigration(stakingtypes.ModuleName, 1, m.Migrate1to2); err != nil {
> 		panic(fmt.Sprintf("failed to migrate x/%s from version 1 to 2: %v", stakingtypes.ModuleName, err))
> ```

errors.Errorf

wojtek-coreum (commented):
> Reviewed 6 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 87 of 96 files reviewed, 27 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (commented):
> Reviewed 2 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 89 of 96 files reviewed, 27 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

wojtek-coreum (commented):
> Reviewed 7 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 27 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 27 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

___
*[`go.mod` line 65 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyj56q8gOtHeuIv33x:-NTIqzbVAcB7R8mjYgos:b2nh4j5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/go.mod#L65)):*

funny
</blockquote></details>

:-)
___
*[`app/app.go` line 19 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyjZqT1xQKgWdviLET:-NTIeFkK8WfSbU4Sfou1:bdw6ybz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L19)):*

using `tm` prefix for cometbft imports doesn't seem to make sense
</blockquote></details>

The cosmos team updated it, but still uses the Prev keys, everywhere, that's why I do the same. Let's discuss with the team.
___
*[`app/app.go` line 154 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyk2DSDzni3jeRCHrz:-NTIeVlL9wlXXzvpAy6C:b-sufmpv) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/app/app.go#L154)):*

Does it mean that there are no CLI subcommands for those specific proposals?
</blockquote></details>

Yes, you just build the proposal file. That's how I understand it now. But we will get more clarity once starts working on updates for our proposals.
___
*[`app/app.go` line 256 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyla-G3NJqlECvNJmK:-NTIf83F8h31v4wXPaux:b-r101q6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L256)):*

If `mm` was renamed, then let's rename `sm` too
</blockquote></details>

I did it to simplify the merge, since it's `ModuleManager` in the SDK now. I can't update the `sm` because there is a method (interface impl ) SimulationManager. Returned the name to mm back. 
___
*[`app/app.go` line 309 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSymNLg7lE1vM85u3C5:-NTIfsn94gCGjHq_RohL:b8nrop) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L309)):*

Funny, that params are encoded using `legacyAmino` internally and `appCodec` is unused.
</blockquote></details>

Seem like used.



_Code quote:_
```Go
initParamsKeeper(
```
___
*[`app/app.go` line 468 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyoNoXBMCasxeRtjqw:-NTIg7MQAyGq07GG4UEo:b8wvs8u) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L468)):*

What does it mean? Aren't they set in genesis.json?
</blockquote></details>

They will be set. This is the default. 



_Code quote:_
```Go
govConfig.MaxMetadataLen = 10000
```
___
*[`app/app.go` line 538 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSypD0h-4Cpmp-xj5RW:-NTIlJyN3aNcJhxlW6JZ:bomio4f) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L538)):*

add "Deprecated" amrk
</blockquote></details>

Updated with `FIXME(v47-legacy)`
___
*[`app/app.go` line 568 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSypaN0678QnDv8xd-R:-NTIlXA-2fM9enbF-8i2:b-6wxon6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L568)):*

comment says it should be the last module in the list
</blockquote></details>

Interesting things :-) In the SDK now such comment, but there is such in the WASM. Moved to be the last.



_Code quote:_
```Go
always be last to make sure that it
```
___
*[`app/app.go` line 954 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyqshODGEG08AMwIgJ:-NTIm1xe2F9zRj4-nvEF:b5tlwln) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L954)):*

Should it be renamed? We no longer use tendermint
</blockquote></details>

It implements interface. 



_Code quote:_
```Go
RegisterTendermintSe
```
___
*[`app/app.go` line 970 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyr6k8E6ZPM25kKRAq:-NTImAbzEET3IWGPInrj:bd7t2it) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L970)):*

Together with `BlockedAddresses` function it is not used anywhere
</blockquote></details>

Correct. Removed.



_Code quote:_
```Go
GetMaccPerms() map[stri
```
___
*[`app/export.go` line 38 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSys-z5-o2xb-VmZ-4r:-NTImJsY9ql4WgLepll9:bd3tvvj) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/app/export.go#L38)):*

why removed? it was a good idea to return empty object in case of error
</blockquote></details>

I took that latest version of the `export.go`, and it was done that way there. Agree, better prev version, fixed.



_Code quote:_
```Go
validators, err := staking.WriteValida
```
___
*[`app/export.go` line 80 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSys9SN4UwyuhPcio_D:-NTImd5F2DVsKajCO7GM:byudb4t) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/app/export.go#L80)):*

why removed?
</blockquote></details>

Same as prev comment, copy/paste. Updated.



_Code quote:_
```Go
_, _ = app.DistrKeepe
```
___
*[`app/export.go` line 88 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysGzz9YK0sD9eqqRU:-NTImpd-2CshT_sw64ou:bt2vr1b) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L88)):*

error is not handled
</blockquote></details>

Fixed.



_Code quote:_
```Go
_, _ = app.DistrKeeper.WithdrawDel
```
___
*[`app/export.go` line 180 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysmVD47VAOEhHq8Pk:-NTImyqp-GOtHiCotRCY:b-c528zh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L180)):*

It seems like a case for `log.Fatal`
</blockquote></details>

Not sure. Looks like the `Error` is enough.



_Code quote:_
```Go
app.Logger().Error
```
___
*[`cmd/cored/main.go` line 16 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrJkVAVAhTiLFoBAX:-NTIojSjCNZLpV71_RX6:b-lx4fj3) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L16)):*

Why? Don't we like logs?
</blockquote></details>

Fixed. 



_Code quote:_
```Go
	logger := log.NewNopLogg
```
___
*[`cmd/cored/main.go` line 28 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrZtT3nM33-22h8by:-NTIolcCEB7qI6TUY01l:b19r99c) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L28)):*

I think we should use `CORED` for env prefix
</blockquote></details>

Not sure, haven't found real example. The SDK uses empty.
___
*[`cmd/cored/cosmoscmd/init.go` line 1 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytGqg12Jk16Zc1pKx:-NTIou0i9ryAm9-9f1hQ:brrrj89) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/cmd/cored/cosmoscmd/init.go#L1)):*

why removed? How do we initialize genesis now?
</blockquote></details>

We stil have it, but instead of copy we use the SDK. 
___
*[`cmd/cored/cosmoscmd/root.go` line 53 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytjoeAsrzwxqrGsuf:-NTIp0ObA5ueKqTM0WxI:b-4lqfav) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/cosmoscmd/root.go#L53)):*

`CORED`?
</blockquote></details>

Not sure. Why `CORED` ?
___
*[`integration-tests/chain.go` line 23 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz4qTz4x84dEIc-Wb3:-NTIpTov42q3EbQ1Fux5:bms36gs) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/chain.go#L23)):*

I think what `ChainContext` really depends on is the codec, used in `GenAccount()`, not entre encoding config
</blockquote></details>

But we need the `EncodingConfig` here only.
___
*[`integration-tests/chain.go` line 73 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz4WA04zLICYCdWrQN:-NTIpAqDCfKchnYwLZJy:bsxn550) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/chain.go#L73)):*

I think we decided to return errors in tests. But might this operation ever fail?
</blockquote></details>

This situation is almost unachievable, but if achievable we should panic. Since it's unexpected.
___
*[`integration-tests/gov.go` line 205 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz5WDw7OzikmACGtfq:-NTIpYwE3tsYq2vRq9Hz:bj6hnuz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/gov.go#L205)):*

Isn't this and await already there in client context by default?
</blockquote></details>

I haven't found. 
___
*[`pkg/client/tx.go` line 138 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyx5DkAHHyvL994H8-:-NTIpd2LE40NOri7MYzW:b-2kzsew) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/pkg/client/tx.go#L138)):*

Does it make sense to await tx in async mode? By using this mode you want to execute the tx **asynchronously**.
</blockquote></details>

I would not mix up the SDK and our additional functionality. We did it before and now had to remove the compatibility. I would go with custom options fir custom features.
___
*[`pkg/config/genesis/genesis.tmpl.json` line 138 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz0nY62SdKs9FYLkCb:-NTIpsCr4QuOU1w0ffL6:b-m8wwgb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/pkg/config/genesis/genesis.tmpl.json#L138)):*

Do we understand these params?
</blockquote></details>

We have task to set the correct values for them.
___
*[`x/auth/ante/ante.go` line 98 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzEk9WAEuXdh0_8zZi:-NTIpzvuC2TJuw9sVyXQ:bicv8qk) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/x/auth/ante/ante.go#L98)):*

Why is this one removed? Does invariant verification tx work now? We have integration test checking that this message is denied.
</blockquote></details>

You are right. Restored.
___
*[`x/feemodel/ante/ante.go` line 30 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzGIHp-UUFCiqQDZDN:-NTIqZC-0IvYKxt8-70Z:b-cszld4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/x/feemodel/ante/ante.go#L30)):*

You found that we need it, right?
</blockquote></details>

Yes, and it will be restored in the tests PR. 
___
*[`x/wbank/module.go` line 46 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzHQRc3CqKZDxUi0Gi:-NTIqkHADXWHzHV1L5ZI:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/x/wbank/module.go#L46)):*

Here and below: `errors.Errorf` to get full stack trace
</blockquote></details>

Done.
___
*[`x/wstaking/module.go` line 55 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSzHlMV9iMeuFMrRPe9:-NTIqxRSA5DvTgfYUsIl:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/x/wstaking/module.go#L55)):*

errors.Errorf
</blockquote></details>

Done



_Code quote:_
```Go
	panic(fmt.Sprintf("failed to migrate x/%s from version 1 to 2: %v", stakingtypes.ModuleName, e
```

wojtek-coreum (changes requested):
> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 9 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`app/app.go` line 19 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyjZqT1xQKgWdviLET:-NTIuqUs7oyqUZXV4DWF:b5giokv) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L19)):*

The cosmos team updated it, but still uses the Prev keys, everywhere, that's why I do the same. Let's discuss with the team.
</blockquote></details>

I don't care much. After moving to sometbft, the word "tendermint" means nothing
___
*[`app/export.go` line 180 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysmVD47VAOEhHq8Pk:-NTIw-NU0Tc5LYvmHYvT:b4t79v1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L180)):*

Not sure. Looks like the `Error` is enough.
</blockquote></details>

Probably this operation never fails, but if it does, application should not continue.
___
*[`cmd/cored/main.go` line 28 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrZtT3nM33-22h8by:-NTIwZKV7Pfl-dQVAeJJ:b2nw4hi) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L28)):*

Not sure, haven't found real example. The SDK uses empty.
</blockquote></details>

IMO it is better to have env variables like CORED_WHATEVER to avoid clashes 
___
*[`cmd/cored/main.go` line 18 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTIwPQrA3gRmsE7oAck:-NTIwPQrA3gRmsE7oAcl:b934ft3) ([raw file](https://github.com/CoreumFoundation/coreum/blob/078947240866d40111bd9495d245dc80aad58021/cmd/cored/main.go#L18)):*
> ```Go
> 	network, err := cosmoscmd.PreProcessFlags()
> 	if err != nil {
> 		fmt.Printf("Error processing chain id flag, err: %s", err)
> ```

But logger should be used here, to generate logs in proper format which are then sent to grafana
___
*[`cmd/cored/main.go` line 28 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTIwkdC1Km7YSxcYCkb:-NTIwkdC1Km7YSxcYCkc:bihqbda) ([raw file](https://github.com/CoreumFoundation/coreum/blob/078947240866d40111bd9495d245dc80aad58021/cmd/cored/main.go#L28)):*
> ```Go
> 	rootCmd.PersistentFlags().String(flags.FlagChainID, string(app.DefaultChainID), "The network chain ID")
> 	if err := svrcmd.Execute(rootCmd, "", app.DefaultNodeHome); err != nil {
> 		fmt.Printf("Error executing cmd, err: %s", err)
> ```

use logger
___
*[`cmd/cored/cosmoscmd/init.go` line 1 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytGqg12Jk16Zc1pKx:-NTIx6Pp0BGARmsgaOpg:bpinuw8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/cmd/cored/cosmoscmd/init.go#L1)):*

We stil have it, but instead of copy we use the SDK. 
</blockquote></details>

But sdk does not dump the genesis based on chain ID. That's why we implemented our own version
___
*[`integration-tests/chain.go` line 23 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz4qTz4x84dEIc-Wb3:-NTIxK1g1MY_bNOZq8an:b-ttmoyo) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/chain.go#L23)):*

But we need the `EncodingConfig` here only.
</blockquote></details>

That's my point, we need only `Codec` - not entire `EncodingConfig`
___
*[`integration-tests/gov.go` line 205 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz5WDw7OzikmACGtfq:-NTIxdYV8prD9uVlxkJi:b-2e8izf) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/gov.go#L205)):*

I haven't found. 
</blockquote></details>

Inside `NewChain` function we do:

```go
		WithBroadcastMode(flags.BroadcastSync).
		WithAwaitTx(true)
```

That's why I think we don't need to set them here again

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 95 of 98 files reviewed, 9 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

___
*[`app/export.go` line 180 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysmVD47VAOEhHq8Pk:-NTJVwa2-SR96A8dEFcK:b-ipwngb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L180)):*

Probably this operation never fails, but if it does, application should not continue.
</blockquote></details>

Actually, let's add the `log.Fatal(err)`, added. At least I don't see reason not to add it.



_Code quote:_
```Go
app.Logger().Error("error while closing the key-value store reverse prefix iterator: ", err)
```
___
*[`cmd/cored/main.go` line 28 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrZtT3nM33-22h8by:-NTJWmOeEu-JMzAZmUSS:b74g2tk) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L28)):*

IMO it is better to have env variables like CORED_WHATEVER to avoid clashes 
</blockquote></details>

@ysv how about add a task to v0.47 migration, to investigate, that "env" config? 
___
*[`cmd/cored/main.go` line 18 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTIwPQrA3gRmsE7oAck:-NTJW5bx3qJU0V4SDZSY:b-um0lpo) ([raw file](https://github.com/CoreumFoundation/coreum/blob/078947240866d40111bd9495d245dc80aad58021/cmd/cored/main.go#L18)):*

But logger should be used here, to generate logs in proper format which are then sent to grafana
</blockquote></details>

I thought it from the users prospective. The problem is that logs configs (for example JSON) will be sent after the `start` command, so even if we change the logger, it won't use the same format. Also even string logs will the sent to the grafana.
___
*[`cmd/cored/main.go` line 28 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTIwkdC1Km7YSxcYCkb:-NTJX4T82gr_IGI0MuxW:bgweya7) ([raw file](https://github.com/CoreumFoundation/coreum/blob/078947240866d40111bd9495d245dc80aad58021/cmd/cored/main.go#L28)):*

use logger
</blockquote></details>

See prev reply
___
*[`integration-tests/chain.go` line 23 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz4qTz4x84dEIc-Wb3:-NTJZ-BRCZ2WKrNjgPmB:bpa9qvx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/chain.go#L23)):*

That's my point, we need only `Codec` - not entire `EncodingConfig`
</blockquote></details>

Now clear. Updated. 



_Code quote:_
```Go
EncodingConfig         config.EncodingConfig
```
___
*[`integration-tests/gov.go` line 205 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz5WDw7OzikmACGtfq:-NTJZLDu1go3sCRa4Cuj:b-q731an) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/gov.go#L205)):*

Inside `NewChain` function we do:

```go
		WithBroadcastMode(flags.BroadcastSync).
		WithAwaitTx(true)
```

That's why I think we don't need to set them here again
</blockquote></details>

Updated with `WithAwaitTx(false)`. Since we need it here. 



_Code quote:_
```Go
 g.chainCtx.ClientContext.
```

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 93 of 98 files reviewed, 9 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

___
*[`cmd/cored/cosmoscmd/init.go` line 1 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytGqg12Jk16Zc1pKx:-NTK5qzC2qgN512pmsPh:bqvf9i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/4976aa4c2b7c0e3834edd3ce611c207506849dae/cmd/cored/cosmoscmd/init.go#L1)):*

But sdk does not dump the genesis based on chain ID. That's why we implemented our own version
</blockquote></details>

You are right, missed that. Fixed.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 93 of 98 files reviewed, 9 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

___
*[`app/app.go` line 19 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyjZqT1xQKgWdviLET:-NTK5vvi19kcnWRsvDZp:bodbizq) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/app.go#L19)):*

I don't care much. After moving to sometbft, the word "tendermint" means nothing
</blockquote></details>

Agreed to use tm.

wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`app/export.go` line 180 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysmVD47VAOEhHq8Pk:-NTNbeTb9AzB0z9F7zmN:bk4d9ak) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L180)):*

Actually, let's add the `log.Fatal(err)`, added. At least I don't see reason not to add it.
</blockquote></details>

Now you use both `app.Logger().Error()` and `log.Fatal()`. The latter is enough.
___
*[`cmd/cored/main.go` line 18 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTIwPQrA3gRmsE7oAck:-NTNc8dA3nvGLMpusHRR:bm7kaib) ([raw file](https://github.com/CoreumFoundation/coreum/blob/078947240866d40111bd9495d245dc80aad58021/cmd/cored/main.go#L18)):*

I thought it from the users prospective. The problem is that logs configs (for example JSON) will be sent after the `start` command, so even if we change the logger, it won't use the same format. Also even string logs will the sent to the grafana.
</blockquote></details>

I see, but it still seems weird tht some logs are just printed and others use logger. I think logs should be logged by the logger everywhere.
___
*[`cmd/cored/cosmoscmd/init.go` line 79 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTNcbjEBEpVRt9mQ7nv:-NTNcbjEBEpVRt9mQ7nw:b-p3w9r7) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b87d8199a81c1c8eff1c9ada54741c81a341fc64/cmd/cored/cosmoscmd/init.go#L79)):*
> ```Go
> 				chainID = clientCtx.ChainID
> 			default:
> 				chainID = fmt.Sprintf("test-chain-%v", tmrand.Str(6))
> ```

Wouldn't it be better to return an error in this case. This version creates an unexpected mess.
___
*[`cmd/cored/cosmoscmd/init.go` line 140 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTNczCQ3rxS0i1FsU8G:-NTNczCQ3rxS0i1FsU8H:b-rffimi) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b87d8199a81c1c8eff1c9ada54741c81a341fc64/cmd/cored/cosmoscmd/init.go#L140)):*
> ```Go
> 	cmd.Flags().Bool(genutilcli.FlagRecover, false, "provide seed phrase to recover existing key instead of creating")
> 	cmd.Flags().String(flags.FlagChainID, "", "genesis file chain-id, if left blank will be randomly created")
> 	cmd.Flags().Int64(flags.FlagInitHeight, 1, "specify the initial block height at genesis")
> ```

Does it mean that we could start a blockchain from the height 5? Did Cosmos SDK team invent this?
___
*[`integration-tests/gov.go` line 205 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz5WDw7OzikmACGtfq:-NTNdSFGAajYszdHNYHK:bb7r2f0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/gov.go#L205)):*

Updated with `WithAwaitTx(false)`. Since we need it here. 
</blockquote></details>

why?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 7 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

___
*[`app/export.go` line 180 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSysmVD47VAOEhHq8Pk:-NTO60e57e2OY3S-y5V3:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/app/export.go#L180)):*

Now you use both `app.Logger().Error()` and `log.Fatal()`. The latter is enough.
</blockquote></details>

Done.



_Code quote:_
```Go
app.Logger().Error("error while clos
```
___
*[`cmd/cored/main.go` line 18 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTIwPQrA3gRmsE7oAck:-NTO67_U-4eqWnpVHm3T:byeftic) ([raw file](https://github.com/CoreumFoundation/coreum/blob/078947240866d40111bd9495d245dc80aad58021/cmd/cored/main.go#L18)):*

I see, but it still seems weird tht some logs are just printed and others use logger. I think logs should be logged by the logger everywhere.
</blockquote></details>

This is the CLI logger. For example when you query for balance, it uses the same out as here. 
But for the node it's different. So it can't be the same.
___
*[`cmd/cored/cosmoscmd/init.go` line 79 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTNcbjEBEpVRt9mQ7nv:-NTO6qTV2_qQsW4VXdtx:b-8ezbo2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b87d8199a81c1c8eff1c9ada54741c81a341fc64/cmd/cored/cosmoscmd/init.go#L79)):*

Wouldn't it be better to return an error in this case. This version creates an unexpected mess.
</blockquote></details>

Agree, updated.



_Code quote:_
```Go
chainID = fmt.Sprintf("test-chain-%v", 
```
___
*[`cmd/cored/cosmoscmd/init.go` line 140 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTNczCQ3rxS0i1FsU8G:-NTO6thgEPIu-tJjBysY:bajbxg2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b87d8199a81c1c8eff1c9ada54741c81a341fc64/cmd/cored/cosmoscmd/init.go#L140)):*

Does it mean that we could start a blockchain from the height 5? Did Cosmos SDK team invent this?
</blockquote></details>

This is the config for the genesis. You can can set the height when you init the chain.
___
*[`integration-tests/gov.go` line 205 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSz5WDw7OzikmACGtfq:-NTO7124-kN_RkgwFRXO:bgep80m) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/integration-tests/gov.go#L205)):*

why?
</blockquote></details>

We execute the TXs here in parallel, and later await for the txs.

wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 4 unresolved discussions (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

___
*[`cmd/cored/cosmoscmd/init.go` line 140 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTNczCQ3rxS0i1FsU8G:-NTOBq6xEd_PD6jjMDmG:bt9kx1d) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b87d8199a81c1c8eff1c9ada54741c81a341fc64/cmd/cored/cosmoscmd/init.go#L140)):*

This is the config for the genesis. You can can set the height when you init the chain.
</blockquote></details>

I understand it. But what happens if I create a genesis wit initial height equal to 5? Will the genesis block have height 5?
___
*[`cmd/cored/cosmoscmd/init.go` line 78 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTOBhjXCH2W-Y2zhnqz:-NTOBhjXCH2W-Y2zhnr-:b-e4gz9l) ([raw file](https://github.com/CoreumFoundation/coreum/blob/f23d367167ce78ffc69b014713a23da4d85111f7/cmd/cored/cosmoscmd/init.go#L78)):*
> ```Go
> 				chainID = clientCtx.ChainID
> 			default:
> 				return errors.Errorf("undefined %s", flags.FlagChainID)
> ```

`undefined chain ID %s`

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 4 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

___
*[`cmd/cored/cosmoscmd/init.go` line 140 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTNczCQ3rxS0i1FsU8G:-NTT3dVX2L2fm5kB-JsE:b-y55s2i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b87d8199a81c1c8eff1c9ada54741c81a341fc64/cmd/cored/cosmoscmd/init.go#L140)):*

I understand it. But what happens if I create a genesis wit initial height equal to 5? Will the genesis block have height 5?
</blockquote></details>

Good point - not for our implementation. Removed that flag.
___
*[`cmd/cored/cosmoscmd/init.go` line 78 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTOBhjXCH2W-Y2zhnqz:-NTT3acx4RD9JnjryF-A:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/f23d367167ce78ffc69b014713a23da4d85111f7/cmd/cored/cosmoscmd/init.go#L78)):*

`undefined chain ID %s`
</blockquote></details>

Done.

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 2 unresolved discussions (waiting on @miladz68, @silverspase, and @ysv)

miladz68 (changes requested):
> Reviewed 1 of 96 files at r1, 1 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 3 unresolved discussions (waiting on @dzmitryhil, @silverspase, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTmUqaF4aRaRk710jTR:-NTmUqaF4aRaRk710jTS:bhy679k) (no related file):*
This code does not compile, does it ?

___

miladz68 (approved):

miladz68 (commented):
> Reviewed 1 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 3 unresolved discussions (waiting on @dzmitryhil, @silverspase, @wojtek-coreum, and @ysv)

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 3 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTmUqaF4aRaRk710jTR:-NTn2t-87Jf-7unAY6lC:b-mwbsgk) (no related file):*

This code does not compile, does it ?
</blockquote></details>

The code compiles, but the tests don't. And that is expected. In the tests PR all are fixed.

___

miladz68 (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 2 unresolved discussions (waiting on @silverspase, @wojtek-coreum, and @ysv)

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: 95 of 98 files reviewed, 8 unresolved discussions (waiting on @miladz68, @silverspase, @wojtek-coreum, and @ysv)

___
*[`app/app.go` line 228 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTZQWez5aFKdnk0HG7S:-NTsGfwF5NXbH4Y86lAx:bdjnzym) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d907c3690cddbb793a525b35ebc963978d30aa7e/app/app.go#L228)):*

question: does anyone know why they pass some keepers by value and others by reference ?
</blockquote></details>

Because they initialize them with the hooks later.
___
*[`app/app.go` line 414 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTZRjJ06UYpyMRIqa0j:-NTsGsuI4XKaQk_X5ctF:b7abvlb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d907c3690cddbb793a525b35ebc963978d30aa7e/app/app.go#L414)):*

so skip upgrade heights is a new feature ?
</blockquote></details>

No, why new? 
___
*[`app/app.go` line 683 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTZT1Ob3AWVxn7DLsyc:-NTsGuo288olcUPSHdXd:b-dru2a5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d907c3690cddbb793a525b35ebc963978d30aa7e/app/app.go#L683)):*

so deterministicgas is fully disabled currently ?
</blockquote></details>

Yes, it's fully disabled now.
___
*[`cmd/cored/main.go` line 28 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrZtT3nM33-22h8by:-NTsHAgVAZXE4JaPiS3p:b-mbsx0q) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L28)):*

I thought that we already have it like Wojtek mentioned `CORED_SMTH`. I recall we had this discussions

pls make a quick check for this & add to the task  if needed
</blockquote></details>

```
The method SetEnvPrefix sets a prefix for the key to load config from environment variable. Both BindEnv and AutomaticEnv uses this function to load configuration. EnvPrefix helps when you want to load only few environment variables specific to your application and not all. By setting a prefix, Viper can load the environment variables with the configured prefix. For example, if you configure Viper to have TGC as the prefix, then Viper will only load the environment variables that start with TGC_.

```
IMO we don't need to set it.



_Code quote:_
```Go
 svrcmd.Execute(rootCmd, "", app.DefaultNodeHo
```
___
*[`cmd/cored/cosmoscmd/root.go` line 53 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytjoeAsrzwxqrGsuf:-NTsI-kfDTOsCd-J6QmM:b5l76kl) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/cosmoscmd/root.go#L53)):*

Before we had envPrefix here which was probably equal to CORED but now it is empty string
</blockquote></details>

But in that case, we can break compatibility. 
___
*[`cmd/cored/cosmoscmd/root.go` line 198 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTs7sOh7rAczCYTAMAq:-NTsI6MQ-QQema6WOwq7:b-otbcyb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d907c3690cddbb793a525b35ebc963978d30aa7e/cmd/cored/cosmoscmd/root.go#L198)):*

minor: why can't we use subCmd var name here ?
</blockquote></details>

This is the copied code.
___
*[`integration-tests/gov.go` line 10 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTs9tU82UIh8TiNX64Q:-NTsICmxAvGX9XKj4BzF:bkciu39) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d907c3690cddbb793a525b35ebc963978d30aa7e/integration-tests/gov.go#L10)):*

v1beta1 is old and v1 is new or vice versa ?
</blockquote></details>

v1beta1 - old
___
*[`integration-tests/modules/authz_test.go` line 65 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTsDEr71Z_6HQJCgsiV:-NTsN-h71a__xXzcDEJ1:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d907c3690cddbb793a525b35ebc963978d30aa7e/integration-tests/modules/authz_test.go#L65)):*

minor: to avoid these change everywhere you could use `lo.ToPtr(time.Now().Add(time.Minute))`
</blockquote></details>

Done
___
*[`x/feemodel/ante/ante.go` line 30 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NTsF9yf3WpAq0VRjQq4:-NTsN1qY1OLluE-hk6eO:b9o6gyb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d907c3690cddbb793a525b35ebc963978d30aa7e/x/feemodel/ante/ante.go#L30)):*

check should be `ctx.ChainID() == helpers.SimAppChainID` ?
Or this const doesn't exist anymore
</blockquote></details>

We check it in the tests PR.

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 2 unresolved discussions (waiting on @silverspase and @wojtek-coreum)

___
*[`cmd/cored/main.go` line 28 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrZtT3nM33-22h8by:-NTtMlXNBSPxVWl7fh6o:b-1cgzll) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L28)):*

```
The method SetEnvPrefix sets a prefix for the key to load config from environment variable. Both BindEnv and AutomaticEnv uses this function to load configuration. EnvPrefix helps when you want to load only few environment variables specific to your application and not all. By setting a prefix, Viper can load the environment variables with the configured prefix. For example, if you configure Viper to have TGC as the prefix, then Viper will only load the environment variables that start with TGC_.

```
IMO we don't need to set it.

</blockquote></details>

well I see this as beneficial because then we have all our env variables prefixed with `CORED`
___
*[`cmd/cored/cosmoscmd/root.go` line 53 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytjoeAsrzwxqrGsuf:-NTtNEpi31kdarKrxLee:b-xrivy5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/cosmoscmd/root.go#L53)):*

But in that case, we can break compatibility. 
</blockquote></details>

it was my guess that rootOptions.envPrefix was equal to CORED 
NO ?

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 2 unresolved discussions (waiting on @silverspase and @wojtek-coreum)

___
*[`cmd/cored/main.go` line 28 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSyrZtT3nM33-22h8by:-NTx_8NH1SgrL8XGzt59:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/main.go#L28)):*

well I see this as beneficial because then we have all our env variables prefixed with `CORED`
</blockquote></details>

Done.



_Code quote:_
```Go
svrcmd.Execute(rootCmd, "", app.DefaultNodeH
```
___
*[`cmd/cored/cosmoscmd/root.go` line 53 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/465#-NSytjoeAsrzwxqrGsuf:-NTxaspX9JGEfYlyv6P5:bwe6iji) ([raw file](https://github.com/CoreumFoundation/coreum/blob/cbf701ffca33d64ed248a228963ab7768bea9deb/cmd/cored/cosmoscmd/root.go#L53)):*

it was my guess that rootOptions.envPrefix was equal to CORED 
NO ?
</blockquote></details>

We used `rootOptions.envPrefix` but was it CORED ? The issue is that we don't use the `rootOptions` anymore here. And we set create the rootCmd first and then pass it to the `svrcmd.Execute(rootCmd, coreumEnvPrefix, app.DefaultNodeHome); ` Where the `coreumEnvPrefix` is `CORE` now.



_Code quote:_
```Go
WithViper("")
```

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase)

ysv (approved):
> Reviewed 1 of 96 files at r1, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/465)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase)

PR #531, opened 2023-06-21
Remove proto files which fail generation.

vertex451:
Just checking if we need those proto files if no - lets remove them, otherwise I need to spend more time to understand why proto doc generation fails

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/531)


wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/531)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @miladz68, @silverspase, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/531#-NYX2H972x9T2ztIylfB:-NYX2H972x9T2ztIylfC:b-fvjnz8) (no related file):*
@dzmitryhil IIRC you initially copied cosmos protos to verify in CI that backward compatibility is preserved, right? If this is true then removing them is not a good idea I guess.

___

PR #575, opened 2023-07-17
Add deterministic_gas_factor_exceed_expected_max metric.

dzmitryhil:


miladz68 (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-:-N_YBjTg8-0zdFVnuSh6:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @silverspase, @wojtek-coreum, and @ysv)

wojtek-coreum (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @silverspase, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 144 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_bfVgF7a70wNYSaeuR:-N_bfVgF7a70wNYSaeuS:b-vmcra3) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3038746ee0d37297b82ff29b7c145e9689ecb326/x/deterministicgas/types/gas.go#L144)):*
> ```Go
> 		{Name: "msg_name", Value: msgURL},
> 	})
> 	if gasFactor > expectedMaxGasFactor {
> ```

`reportDeterministicGasMetric` is called only if message handler succeeded. IIRC we ent this metric to be reported even (and especially) when handler fails with "out of gas" error.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase, @wojtek-coreum, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 144 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_bfVgF7a70wNYSaeuR:-N_btExG3TFQXXeatbZ2:b-me0yzs) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3038746ee0d37297b82ff29b7c145e9689ecb326/x/deterministicgas/types/gas.go#L144)):*

`reportDeterministicGasMetric` is called only if message handler succeeded. IIRC we ent this metric to be reported even (and especially) when handler fails with "out of gas" error.
</blockquote></details>

The issue with that approach is that lack of gas causes panic which is handled by the SDK recovery middleware which is unexported. So if we want to report that we will have to copy that part and control the changes after each update which IMO doesn't worth the output we get. 

SDK sources: https://github.com/cosmos/cosmos-sdk/blob/v0.45.16/baseapp/baseapp.go#L629

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase, @wojtek-coreum, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 144 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_bfVgF7a70wNYSaeuR:-N_c-Cyy-pANdMaPLgrO:bcvjgzp) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3038746ee0d37297b82ff29b7c145e9689ecb326/x/deterministicgas/types/gas.go#L144)):*

We could catch the panic here, do the magic and panic again
</blockquote></details>

Good idea. Done.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: all files reviewed, 2 unresolved discussions (waiting on @dzmitryhil, @silverspase, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 104 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_c4nWiA3_X3ZKgyQ-6:-N_c4nWiA3_X3ZKgyQ-7:b3cdevd) ([raw file](https://github.com/CoreumFoundation/coreum/blob/605e795f903444bdedd6ea738ee51e7591b23fa2/x/deterministicgas/types/gas.go#L104)):*
> ```Go
> 							_, isOutOfGasError := recoveryObj.(sdk.ErrorOutOfGas)
> 							if isOutOfGasError &&
> 								isDeterministic &&
> ```

maybe some code deduplication could be applied?
___
*[`x/deterministicgas/types/gas.go` line 107 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_c3M--6Lq9W-EZiAKA:-N_c3M--6Lq9W-EZiAKB:b-1qvup0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/605e795f903444bdedd6ea738ee51e7591b23fa2/x/deterministicgas/types/gas.go#L107)):*
> ```Go
> 								!newSDKCtx.IsCheckTx() &&
> 								!newSDKCtx.IsReCheckTx() {
> 								reportDeterministicGasFactorExceedMetric(fuseGasMultiplier, proto.MessageName(msg))
> ```

I think the metric reported here should not report the factor because it is unknown. It should only report the fact that our estimation is wrong and causes panics.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: all files reviewed, 2 unresolved discussions (waiting on @silverspase, @wojtek-coreum, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 104 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_c4nWiA3_X3ZKgyQ-6:-N_c6niR3z8U61PNq5BK:b-bw8e14) ([raw file](https://github.com/CoreumFoundation/coreum/blob/605e795f903444bdedd6ea738ee51e7591b23fa2/x/deterministicgas/types/gas.go#L104)):*

maybe some code deduplication could be applied?
</blockquote></details>

Updated.
___
*[`x/deterministicgas/types/gas.go` line 107 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_c3M--6Lq9W-EZiAKA:-N_c5SwM0cU8mR8swg1d:bpqale0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/605e795f903444bdedd6ea738ee51e7591b23fa2/x/deterministicgas/types/gas.go#L107)):*

I think the metric reported here should not report the factor because it is unknown. It should only report the fact that our estimation is wrong and causes panics.
</blockquote></details>

I thought about it, but in terms of metric, we can use one instead of two, since the meaning of it is the same. And we know if the value is 10 we exceeded the max.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @silverspase, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 107 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_c3M--6Lq9W-EZiAKA:-N_cAVvs9Bf9YmENRHHB:b-c2qw1i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/605e795f903444bdedd6ea738ee51e7591b23fa2/x/deterministicgas/types/gas.go#L107)):*

I thought about it, but in terms of metric, we can use one instead of two, since the meaning of it is the same. And we know if the value is 10 we exceeded the max. 
</blockquote></details>

I got you but still think it's misleading. We don't know it's 10, it might be 1000. If we do it this way then someone might think, by looking at metrics that increasing deterministic gas 10 times is the right thing to do.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: all files reviewed, 1 unresolved discussion (waiting on @silverspase, @wojtek-coreum, and @ysv)

___
*[`x/deterministicgas/types/gas.go` line 107 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/575#-N_c3M--6Lq9W-EZiAKA:-N_cBqzk1EKC15YovTmH:bgmmj8c) ([raw file](https://github.com/CoreumFoundation/coreum/blob/605e795f903444bdedd6ea738ee51e7591b23fa2/x/deterministicgas/types/gas.go#L107)):*

I got you but still think it's misleading. We don't know it's 10, it might be 1000. If we do it this way then someone might think, by looking at metrics that increasing deterministic gas 10 times is the right thing to do.
</blockquote></details>

Added new metric: deterministic_gas_factor_exceed_fuse_gas_multiplier

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/575)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @miladz68 and @silverspase)

PR #681, opened 2023-10-20
New feature allowing to disable sending FT to smart contracts

wojtek-coreum:
# Description

There are 4 entry points now, which are handled to detect if sender or recipient is a smart contract:
- single send in bank keeper
- multi send in bank keeper
- token transferrer in WASM
- handler of messages executed by smart contract

In those entry points, addresses are added to appropriate list in context (senders or recipients) and then interested parties may verify if sender or recipient exists in the list, meaning that this is smart contract.

It might happen that one address is handled by two entry points. For example, if previously instantiated smart contract is executed with coins attached, the address of smart contract will be marked in two places:
- WASM token transferrer
- bank keeper

# Reviewers checklist:

# Authors checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/681)


dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/681)* status: 0 of 27 files reviewed, 2 unresolved discussions (waiting on @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/wasm/handler/message.go` line 2 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/681#-NhQI7Xm2bSiypq_UNOS:-NhQI7Xm2bSiypq_UNOT:bra3ait) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fea49cd5a4f88903093cd08b8e8076fecf076c2e/x/wasm/handler/message.go#L2)):*
> ```Go
> package handler
> ```

The wrapper part should in the `wwasm` package keeping the same structure as in the original wasm. That package contains just standard wasm supported extensions. 
___
*[`x/wasm/types/purpose.go` line 7 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/681#-NhQIsjkDwoy7ZJrd5S1:-NhQIsjkDwoy7ZJrd5S2:b55gf87) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fea49cd5a4f88903093cd08b8e8076fecf076c2e/x/wasm/types/purpose.go#L7)):*
> ```Go
> )
> 
> type smartContractRecipientKey struct{}
> ```

You can group them with the 
```go
type (  
    smartContractRecipientKey struct{}  
    smartContractSenderKey    struct{}  
)
```
___
*[`x/wasm/types/purpose.go` line 31 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/681#-NhQJL2n0NAC-YL1SwrP:-NhQJL2n0NAC-YL1SwrQ:bth0vn4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fea49cd5a4f88903093cd08b8e8076fecf076c2e/x/wasm/types/purpose.go#L31)):*
> ```Go
> 
> func add(ctx sdk.Context, addr string, key struct{}) sdk.Context {
> 	set, ok := ctx.Value(key).(map[string]struct{})
> ```

Could you please explain why don't you use the string directly here and in the `has`?

PR #812, opened 2024-04-26
Modify NFT contract to use new protos for updatable NFTs and include examples

keyleu:
# Description
This PR uses the new generated rust proto types (which are in a temporary branch of the coreum-wasm-sdk until we upgrade) in the integration tests contracts to mint and update NFT data. Remarks:

1) The idea is to move away from the wasm_handler (deprecate it) and start using the protos with grpc directly (in cosmwasm 2.0, Stargate is renamed to grpc) so that we don't have to rely in custom bindings, keep consistency between all procedures and also not need to keep bloating the wasm_handler when it's not necessary at all because we can interact directly with grpc from the smart contracts.
2) For queries we can keep adding them for convenience because they still need to be whitelisted.
3) One thing I realized is an inconsistent behavior between wasm (due to the handler's code) and directly using grpc. When we don't pass data in grpc, the NFT is created with a DataBytes structure with an empty byte array inside, however, when we were doing it with wasm, there was no DataBytes structure at all (no data field). It's just a small thing and not really important but shows why we should try to avoid complicated logic and extending the handler to avoid such behaviors (It would also have avoided all the Base64 problem we had 1 year ago).
4) Left one TODO to replace the contract dependency to the new coreum-wasm-sdk version when the final version is uploaded to crates.io

# Reviewers checklist:

# Authors checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)


dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @keyleu, @masihyeganeh, @miladz68, and @ysv)

___
*[`integration-tests/contracts/modules/nft/src/contract.rs` line 174 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/812#-Nwd8e3a6qprhOxWQbbV:-Nwd8e3a6qprhOxWQbbW:bh6odhz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/e2f38f122afde38fdf1c2ebb2b5e8e78e4c48f5b/integration-tests/contracts/modules/nft/src/contract.rs#L174)):*
> ```XML
>     let mint_bytes = mint.to_proto_bytes();
> 
>     let msg = CosmosMsg::Stargate {
> ```

What about the migration of other messages to use the `CosmosMsg::Stargate`?
And what about the coreum SDK, what will it provide?

keyleu (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`integration-tests/contracts/modules/nft/src/contract.rs` line 174 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/812#-Nwd8e3a6qprhOxWQbbV:-Nwdg0GK3W5_M4W2toja:b-anbiyt) ([raw file](https://github.com/CoreumFoundation/coreum/blob/e2f38f122afde38fdf1c2ebb2b5e8e78e4c48f5b/integration-tests/contracts/modules/nft/src/contract.rs#L174)):*

What about the migration of other messages to use the `CosmosMsg::Stargate`?
And what about the coreum SDK, what will it provide?
</blockquote></details>

1) There is no need for migration, all messages work using Stargate/Grpc using the right proto structures, which we will have in the wasm sdk (we have them there already because they are used for test tube)

2) Coreum wasm SDK provides all the proto structures required and will be updated/regenerated with every upgrade (we have a tool for that), and also api for queries (unless you guys want to start whitelisting like other chains do then we don't even need this).  And also some constants that are useful, for example the values of the features for both assetft and nft  (minting, burning, soulbound etc).
All current msg structures can be kept there for people who are already using them and since we will keep the wasm handler because we need backward compatibility we can keep those messages in the wasm sdk too.

Actually the wasm handler is not needed for anything, because queries can be whitelisted.

keyleu (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`integration-tests/contracts/modules/nft/src/contract.rs` line 174 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/812#-Nwd8e3a6qprhOxWQbbV:-NwdvAlO-R0ENUzm2dNH:bosenx8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/e2f38f122afde38fdf1c2ebb2b5e8e78e4c48f5b/integration-tests/contracts/modules/nft/src/contract.rs#L174)):*
<details><summary><i>Previously, keyleu (Keyne) wrote…</i></summary><blockquote>

1) There is no need for migration, all messages work using Stargate/Grpc using the right proto structures, which we will have in the wasm sdk (we have them there already because they are used for test tube)

2) Coreum wasm SDK provides all the proto structures required and will be updated/regenerated with every upgrade (we have a tool for that), and also api for queries (unless you guys want to start whitelisting like other chains do then we don't even need this).  And also some constants that are useful, for example the values of the features for both assetft and nft  (minting, burning, soulbound etc).
All current msg structures can be kept there for people who are already using them and since we will keep the wasm handler because we need backward compatibility we can keep those messages in the wasm sdk too.

Actually the wasm handler is not needed for anything, because queries can be whitelisted.
</blockquote></details>

I meant the wasm handler was not really needed for anything, now it's needed because we need backward compatibility

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @keyleu, @masihyeganeh, @miladz68, and @ysv)

___
*[`integration-tests/modules/wasm_test.go` line 1687 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/812#-NwhVU8q8gzZbb9G7lVu:-NwhVU8q8gzZbb9G7lVv:b3fupmj) ([raw file](https://github.com/CoreumFoundation/coreum/blob/e2f38f122afde38fdf1c2ebb2b5e8e78e4c48f5b/integration-tests/modules/wasm_test.go#L1687)):*
> ```Go
> 	requireT.NoError(err)
> 
> 	// FIXME: When minting an NFT with empty data it creates an []byte{} in the data
> ```

Here is the fix for the test.
```go

	emptyData, err := codectypes.NewAnyWithValue(&assetnfttypes.DataBytes{Data: nil})
	requireT.NoError(err)
	expectedNFT := &nfttypes.NFT{
		ClassId: classIDNoWhitelist,
		Id:      mintImmutableNFTReq.ID,
		Data:    emptyData,
	}

	gotNFT := nftResp.Nft
	// encode the data `from` and `to` proto.Any to load same state as `codectypes.NewAnyWithValue` does
	var gotNFTDataBytes assetnfttypes.DataBytes
	requireT.NoError(gotNFTDataBytes.Unmarshal(gotNFT.Data.Value))
	gotNFTData, err := codectypes.NewAnyWithValue(&gotNFTDataBytes)
	requireT.NoError(err)
	gotNFT.Data = gotNFTData

	requireT.Equal(
		expectedNFT, gotNFT,
	)
```

miladz68 (approved):
> Reviewed 4 of 6 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: all files reviewed, 1 unresolved discussion (waiting on @keyleu, @masihyeganeh, and @ysv)

keyleu (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`integration-tests/modules/wasm_test.go` line 1687 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/812#-NwhVU8q8gzZbb9G7lVu:-NwnVHuH5dw2c-7OFmG8:bx7dunl) ([raw file](https://github.com/CoreumFoundation/coreum/blob/e2f38f122afde38fdf1c2ebb2b5e8e78e4c48f5b/integration-tests/modules/wasm_test.go#L1687)):*

Here is the fix for the test.
```go

	emptyData, err := codectypes.NewAnyWithValue(&assetnfttypes.DataBytes{Data: nil})
	requireT.NoError(err)
	expectedNFT := &nfttypes.NFT{
		ClassId: classIDNoWhitelist,
		Id:      mintImmutableNFTReq.ID,
		Data:    emptyData,
	}

	gotNFT := nftResp.Nft
	// encode the data `from` and `to` proto.Any to load same state as `codectypes.NewAnyWithValue` does
	var gotNFTDataBytes assetnfttypes.DataBytes
	requireT.NoError(gotNFTDataBytes.Unmarshal(gotNFT.Data.Value))
	gotNFTData, err := codectypes.NewAnyWithValue(&gotNFTDataBytes)
	requireT.NoError(err)
	gotNFT.Data = gotNFTData

	requireT.Equal(
		expectedNFT, gotNFT,
	)
```
</blockquote></details>

Thanks for this! Changed it.

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: 5 of 6 files reviewed, all discussions resolved (waiting on @masihyeganeh, @miladz68, and @ysv)

miladz68 (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/812)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @masihyeganeh and @ysv)

PR #818, opened 2024-05-01
Implementing features in extension contract

masihyeganeh:
# Description

# Reviewers checklist:

# Authors checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,48 +19,28 @@ jobs:
           "integration tests coreum-modules",
           "integration tests coreum-ibc",
           "integration tests coreum-upgrade-v4.0.0",
-          "integration tests faucet",
         ]
         include:
           - ci_step: "lint"
-            command: "coreum-builder lint"
+            command: make lint
             linter-cache: true
             codecov: false
           - ci_step: "test"
-            command: "coreum-builder wasm test"
+            command: make test
             linter-cache: false
             codecov: true
             codecov-name: coreum
           - ci_step: "integration tests coreum-modules"
-            command: |
-              coreum-builder wasm build images
-              crust znet start --profiles=3cored --timeout-commit 0.5s
-              coreum-builder integration-tests-unsafe/modules
-              crust znet stop
-              crust znet coverage-convert
+            command: make integration-tests-modules
             linter-cache: false
             codecov: true
             codecov-name: coreum-integration-tests-modules
           - ci_step: "integration tests coreum-ibc"
-            command: |
-              coreum-builder build images
-              crust znet start --profiles=3cored,ibc --timeout-commit 1s
-              coreum-builder integration-tests-unsafe/ibc
+            command: make integration-tests-ibc
             linter-cache: false
             codecov: false
           - ci_step: "integration tests coreum-upgrade-v4.0.0"
-            command: |
-              coreum-builder wasm build images
-              crust znet start --cored-version=v3.0.3 --profiles=3cored,ibc --timeout-commit 1s
-              coreum-builder integration-tests-unsafe/upgrade integration-tests-unsafe/ibc integration-tests-unsafe/modules
-            linter-cache: false
-            codecov: false
-          - ci_step: "integration tests faucet"
-            command: |
-              coreum-builder build images
-              faucet-builder images
-              crust znet start --profiles=faucet --timeout-commit 0.5s
-              faucet-builder integration-tests
+            command: make integration-tests-upgrade
             linter-cache: false
             codecov: false
 
@@ -73,26 +53,11 @@ jobs:
         with:
           persist-credentials: false
           path: coreum
-      - name: Checkout faucet
-        uses: actions/checkout@v4
-        with:
-          repository: CoreumFoundation/faucet
-          path: faucet
       - name: Checkout crust
         uses: actions/checkout@v4
         with:
           repository: CoreumFoundation/crust
           path: crust
-      - name: Set up build system
-        run: |
-          echo "$(pwd)/coreum/bin" >> $GITHUB_PATH
-          echo "$(pwd)/faucet/bin" >> $GITHUB_PATH
-          echo "$(pwd)/crust/bin" >> $GITHUB_PATH
-      - name: Retrieve go version
-        id: goversion
-        run: echo "GO_VERSION=$(coreum/bin/go version)" >> $GITHUB_OUTPUT
-      - name: Print go version
-        run: echo ${{ steps.goversion.outputs.GO_VERSION }}
       - name: Setup go cache
         uses: actions/cache@v4
         with:
@@ -122,16 +87,10 @@ jobs:
         with:
           key: ${{ runner.os }}-docker-${{ steps.get-year-week.outputs.date }} # year-week key
       - name: Prepare dependencies
-        # FIXME (wojciech): Execute `download` commands for crust builder once available
-        run: |
-          crust build/me
-          crust build
-          crust znet remove
-          coreum-builder build/me
-          coreum-builder download
-          faucet-builder build/me
-          faucet-builder download
+        working-directory: ./coreum
+        run: make dependencies
       - name: Run ${{ matrix.ci_step }}
+        working-directory: ./coreum
         run: ${{ matrix.command }}
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v3
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,71 @@
+CoreumBuilder = ./bin/coreum-builder
+CrustBuilder = ../crust/bin/crust
+
+.PHONY: lint
+lint:
+	$(CoreumBuilder) lint
+
+.PHONY: test
+test:
+	$(CoreumBuilder) test
+
+.PHONY: build
+build:
+	$(CoreumBuilder) build
+
+.PHONY: images
+images:
+	$(CoreumBuilder) images
+
+.PHONY: wasm
+wasm:
+	$(CoreumBuilder) wasm
+
+.PHONY: generate
+generate:
+	$(CoreumBuilder) generate
+
+.PHONY: release
+release:
+	$(CoreumBuilder) release
+
+.PHONY: release-images
+release-images:
+	$(CoreumBuilder) release/images
+
+.PHONY: dependencies
+dependencies: check-crust-builder
+	$(CrustBuilder) download
+	$(CoreumBuilder) download
+
+.PHONY: integration-tests-modules
+integration-tests-modules: check-crust-builder
+	$(CrustBuilder) znet remove
+	$(CoreumBuilder) wasm build images
+	$(CrustBuilder) znet start --profiles=3cored --timeout-commit 0.5s
+	$(CoreumBuilder) integration-tests-unsafe/modules
+	$(CrustBuilder) znet stop
+	$(CrustBuilder) znet coverage-convert
+	$(CrustBuilder) znet remove
+
+.PHONY: integration-tests-ibc
+integration-tests-ibc: check-crust-builder
+	$(CrustBuilder) znet remove
+	$(CoreumBuilder) build images
+	$(CrustBuilder) znet start --profiles=3cored,ibc --timeout-commit 1s
+	$(CoreumBuilder) integration-tests-unsafe/ibc
+	$(CrustBuilder) znet remove
+
+.PHONY: integration-tests-upgrade
+integration-tests-upgrade: check-crust-builder
+	$(CrustBuilder) znet remove
+	$(CoreumBuilder) wasm build images
+	$(CrustBuilder) znet start --cored-version=v3.0.3 --profiles=3cored,ibc --timeout-commit 1s
+	$(CoreumBuilder) integration-tests-unsafe/upgrade integration-tests-unsafe/ibc integration-tests-unsafe/modules
+	$(CrustBuilder) znet remove
+
+### Helpers go below this line
+
+.PHONY: check-crust-builder
+check-crust-builder:
+	test -f $(CrustBuilder) || (echo "You need to checkout crust repository" && exit 1)
--- a/build/coreum/build.go
+++ b/build/coreum/build.go
@@ -134,6 +134,8 @@ func Lint(ctx context.Context, deps build.DepsFunc) error {
 
 // Test run unit tests in coreum repo.
 func Test(ctx context.Context, deps build.DepsFunc) error {
+	deps(CompileAllSmartContracts)
+
 	return golang.Test(ctx, repoPath, deps)
 }
 
--- a/build/coreum/generate-proto-breaking.go
+++ b/build/coreum/generate-proto-breaking.go
@@ -99,7 +99,7 @@ func breakingProto(ctx context.Context, deps build.DepsFunc) error {
 	args := []string{
 		"--buf-breaking_out=.",
 		fmt.Sprintf("--buf-breaking_opt=%s", configBuf),
-		"--plugin", must.String(filepath.Abs("bin/protoc-gen-buf-breaking")),
+		fmt.Sprintf("--plugin=%s", tools.Path("bin/protoc-gen-buf-breaking", tools.TargetPlatformLocal)),
 	}
 
 	args = append(args, includeArgs...)
--- a/build/coreum/generate-proto-docs.go
+++ b/build/coreum/generate-proto-docs.go
@@ -49,8 +49,9 @@ func executeProtocCommand(ctx context.Context, deps build.DepsFunc, includeDirs,
 	deps(tools.EnsureProtoc, tools.EnsureProtocGenDoc)
 
 	args := []string{
-		fmt.Sprintf("%s=%s", "--doc_out", "docs"),
-		fmt.Sprintf("%s=%s,api.md", "--doc_opt", filepath.Join("docs", "api.tmpl.md")),
+		"--doc_out=docs",
+		fmt.Sprintf("--plugin=%s", tools.Path("bin/protoc-gen-doc", tools.TargetPlatformLocal)),
+		"--doc_opt=docs/api.tmpl.md,api.md",
 	}
 
 	for _, path := range includeDirs {
--- a/build/coreum/generate-proto-go.go
+++ b/build/coreum/generate-proto-go.go
@@ -3,6 +3,7 @@ package coreum
 import (
 	"bufio"
 	"context"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
@@ -14,7 +15,6 @@ import (
 
 	"github.com/CoreumFoundation/coreum-tools/pkg/build"
 	"github.com/CoreumFoundation/coreum-tools/pkg/libexec"
-	"github.com/CoreumFoundation/coreum-tools/pkg/must"
 	"github.com/CoreumFoundation/crust/build/golang"
 	"github.com/CoreumFoundation/crust/build/tools"
 )
@@ -61,8 +61,8 @@ func executeGoProtocCommand(ctx context.Context, deps build.DepsFunc, includeDir
 	args := []string{
 		"--gocosmos_out=plugins=interfacetype+grpc,Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:.",
 		"--grpc-gateway_out=logtostderr=true:.",
-		"--plugin", must.String(filepath.Abs("bin/protoc-gen-gocosmos")),
-		"--plugin", must.String(filepath.Abs("bin/protoc-gen-grpc-gateway")),
+		fmt.Sprintf("--plugin=%s", tools.Path("bin/protoc-gen-gocosmos", tools.TargetPlatformLocal)),
+		fmt.Sprintf("--plugin=%s", tools.Path("bin/protoc-gen-grpc-gateway", tools.TargetPlatformLocal)),
 	}
 
 	for _, path := range includeDirs {
--- a/build/coreum/generate-proto-lint.go
+++ b/build/coreum/generate-proto-lint.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/CoreumFoundation/coreum-tools/pkg/build"
 	"github.com/CoreumFoundation/coreum-tools/pkg/libexec"
-	"github.com/CoreumFoundation/coreum-tools/pkg/must"
 	"github.com/CoreumFoundation/crust/build/tools"
 )
 
@@ -51,7 +50,7 @@ func executeLintProtocCommand(ctx context.Context, deps build.DepsFunc, includeD
 	args := []string{
 		"--buf-lint_out=.",
 		fmt.Sprintf("--buf-lint_opt=%s", configLint),
-		"--plugin", must.String(filepath.Abs("bin/protoc-gen-buf-lint")),
+		fmt.Sprintf("--plugin=%s", tools.Path("bin/protoc-gen-buf-lint", tools.TargetPlatformLocal)),
 	}
 
 	for _, path := range includeDirs {
--- a/build/coreum/generate-proto-openapi.go
+++ b/build/coreum/generate-proto-openapi.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/CoreumFoundation/coreum-tools/pkg/build"
 	"github.com/CoreumFoundation/coreum-tools/pkg/libexec"
-	"github.com/CoreumFoundation/coreum-tools/pkg/must"
 	"github.com/CoreumFoundation/crust/build/tools"
 )
 
@@ -95,7 +94,7 @@ func executeOpenAPIProtocCommand(ctx context.Context, deps build.DepsFunc, inclu
 
 	args := []string{
 		"--openapiv2_out=logtostderr=true,allow_merge=true,json_names_for_fields=false,fqn_for_openapi_name=true,simple_operation_ids=true,Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:.", //nolint:lll // breaking down this string will make it more complicated.
-		"--plugin", must.String(filepath.Abs("bin/protoc-gen-openapiv2")),
+		fmt.Sprintf("--plugin=%s", tools.Path("bin/protoc-gen-openapiv2", tools.TargetPlatformLocal)),
 	}
 
 	for _, path := range includeDirs {
--- a/build/go.mod
+++ b/build/go.mod
@@ -13,7 +13,7 @@ replace (
 require (
 	github.com/CoreumFoundation/coreum-tools v0.4.1-0.20240321120602-0a9c50facc68
 	github.com/CoreumFoundation/coreum/v4 v4.0.0-20240213123712-d7d6a45ddb8f
-	github.com/CoreumFoundation/crust/build v0.0.0-20240503123939-79485cb9b136
+	github.com/CoreumFoundation/crust/build v0.0.0-20240509142425-5ef601a5fee4
 	github.com/iancoleman/strcase v0.3.0
 	github.com/pkg/errors v0.9.1
 )
--- a/build/go.sum
+++ b/build/go.sum
@@ -2,8 +2,8 @@ github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8
 github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
 github.com/CoreumFoundation/coreum-tools v0.4.1-0.20240321120602-0a9c50facc68 h1:Tb9avuSQW3smVGrUnDh/Y+ML4eK802UsvJNZHsBgOGg=
 github.com/CoreumFoundation/coreum-tools v0.4.1-0.20240321120602-0a9c50facc68/go.mod h1:VD93vCHkxYaT/RhOesXTFgd/GQDW54tr0BqGi5JU1c0=
-github.com/CoreumFoundation/crust/build v0.0.0-20240503123939-79485cb9b136 h1:B8v0bW8BiHNpBWAc6KILVewx6H20DHLYc6r3JiOhlto=
-github.com/CoreumFoundation/crust/build v0.0.0-20240503123939-79485cb9b136/go.mod h1:2jqmpvPVpODkfLXUWc2Fj9voIMloobAHFcbhNxbwGSc=
+github.com/CoreumFoundation/crust/build v0.0.0-20240509142425-5ef601a5fee4 h1:8CgQaP8DsBHu9JSJz2+IbYVRzwqPotzS2CnZKGApmaI=
+github.com/CoreumFoundation/crust/build v0.0.0-20240509142425-5ef601a5fee4/go.mod h1:2jqmpvPVpODkfLXUWc2Fj9voIMloobAHFcbhNxbwGSc=
 github.com/benbjohnson/clock v1.1.0/go.mod h1:J11/hYXuz8f4ySSvYwY0FKfm+ezbsZBKZxNJlLklBHA=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
--- a/integration-tests/contracts/modules/ft/Cargo.lock
+++ b/integration-tests/contracts/modules/ft/Cargo.lock
@@ -108,9 +108,8 @@ checksum = "28c122c3980598d243d63d9a704629a2d748d101f278052ff068be5a4423ab6f"
 
 [[package]]
 name = "coreum-wasm-sdk"
-version = "0.2.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "de2207655b1d9ddbb228d96b9900d8726ebead2d382a893492806230ead2daa0"
+version = "0.3.0"
+source = "git+https://github.com/CoreumFoundation/coreum-wasm-sdk.git?branch=keyne/upgrade-wasm-sdk#e1c571716e4cb3c286b320c0d65530dd28af8390"
 dependencies = [
  "chrono",
  "cosmwasm-schema",
@@ -148,9 +147,9 @@ dependencies = [
 
 [[package]]
 name = "cosmwasm-schema"
-version = "1.5.0"
+version = "1.5.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0df41ea55f2946b6b43579659eec048cc2f66e8c8e2e3652fc5e5e476f673856"
+checksum = "8467874827d384c131955ff6f4d47d02e72a956a08eb3c0ff24f8c903a5517b4"
 dependencies = [
  "cosmwasm-schema-derive",
  "schemars",
@@ -161,9 +160,9 @@ dependencies = [
 
 [[package]]
 name = "cosmwasm-schema-derive"
-version = "1.5.0"
+version = "1.5.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43609e92ce1b9368aa951b334dd354a2d0dd4d484931a5f83ae10e12a26c8ba9"
+checksum = "f6db85d98ac80922aef465e564d5b21fa9cfac5058cb62df7f116c3682337393"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -600,9 +599,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.69"
+version = "1.0.82"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "134c189feb4956b20f6f547d2cf727d4c0fe06722b20a0eec87ed445a97f92da"
+checksum = "8ad3d49ab951a01fbaafe34f2ec74122942fe18a3f9814c3268f1bb72042131b"
 dependencies = [
  "unicode-ident",
 ]
@@ -650,7 +649,7 @@ dependencies = [
  "itertools 0.11.0",
  "proc-macro2",
  "quote",
- "syn 2.0.39",
+ "syn 2.0.61",
 ]
 
 [[package]]
@@ -673,9 +672,9 @@ dependencies = [
 
 [[package]]
 name = "quote"
-version = "1.0.33"
+version = "1.0.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5267fca4496028628a95160fc423a33e8b2e6af8a5302579e322e4b520293cae"
+checksum = "0fa76aaf39101c457836aec0ce2316dbdc3ab723cdda1c6bd4e6ad4208acaca7"
 dependencies = [
  "proc-macro2",
 ]
@@ -713,9 +712,9 @@ checksum = "1ad4cc8da4ef723ed60bced201181d83791ad433213d8c24efffda1eec85d741"
 
 [[package]]
 name = "schemars"
-version = "0.8.15"
+version = "0.8.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1f7b0ce13155372a76ee2e1c5ffba1fe61ede73fbea5630d61eee6fac4929c0c"
+checksum = "fc6e7ed6919cb46507fb01ff1654309219f62b4d603822501b0b80d42f6f21ef"
 dependencies = [
  "dyn-clone",
  "schemars_derive",
@@ -725,14 +724,14 @@ dependencies = [
 
 [[package]]
 name = "schemars_derive"
-version = "0.8.15"
+version = "0.8.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e85e2a16b12bdb763244c69ab79363d71db2b4b918a2def53f80b02e0574b13c"
+checksum = "185f2b7aa7e02d418e453790dde16890256bbd2bcd04b7dc5348811052b53f49"
 dependencies = [
  "proc-macro2",
  "quote",
  "serde_derive_internals",
- "syn 1.0.109",
+ "syn 2.0.61",
 ]
 
 [[package]]
@@ -757,9 +756,9 @@ checksum = "836fa6a3e1e547f9a2c4040802ec865b5d85f4014efe00555d7090a3dcaa1090"
 
 [[package]]
 name = "serde"
-version = "1.0.190"
+version = "1.0.200"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91d3c334ca1ee894a2c6f6ad698fe8c435b76d504b13d436f0685d648d6d96f7"
+checksum = "ddc6f9cc94d67c0e21aaf7eda3a010fd3af78ebf6e096aa6e2e13c79749cce4f"
 dependencies = [
  "serde_derive",
 ]
@@ -784,24 +783,24 @@ dependencies = [
 
 [[package]]
 name = "serde_derive"
-version = "1.0.190"
+version = "1.0.200"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "67c5609f394e5c2bd7fc51efda478004ea80ef42fee983d5c67a65e34f32c0e3"
+checksum = "856f046b9400cee3c8c94ed572ecdb752444c24528c035cd35882aad6f492bcb"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.39",
+ "syn 2.0.61",
 ]
 
 [[package]]
 name = "serde_derive_internals"
-version = "0.26.0"
+version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85bf8229e7920a9f636479437026331ce11aa132b4dde37d121944a44d6e5f3c"
+checksum = "330f01ce65a3a5fe59a60c82f3c9a024b573b8a6e875bd233fe5f934e71d54e3"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 1.0.109",
+ "syn 2.0.61",
 ]
 
 [[package]]
@@ -884,9 +883,9 @@ dependencies = [
 
 [[package]]
 name = "syn"
-version = "2.0.39"
+version = "2.0.61"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23e78b90f2fcf45d3e842032ce32e3f2d1545ba6636271dcbf24fa306d87be7a"
+checksum = "c993ed8ccba56ae856363b1845da7266a7cb78e1d146c8a32d54b45a8b831fc9"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -910,7 +909,7 @@ checksum = "266b2e40bc00e5a6c09c3584011e08b06f123c00362c92b975ba9843aaaa14b8"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.39",
+ "syn 2.0.61",
 ]
 
 [[package]]
--- a/integration-tests/contracts/modules/ft/Cargo.lock
+++ b/integration-tests/contracts/modules/ft/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.7.7"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a824f2aa7e75a0c98c5a504fceb80649e9c35265d44525b5f94de4771a395cd"
+checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
  "getrandom",
  "once_cell",
@@ -15,15 +15,15 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.75"
+version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
+checksum = "25bdb32cbbdce2b519a9cd7df3a678443100e265d5e25ca763b7572a5104f5f3"
 
 [[package]]
 name = "autocfg"
-version = "1.1.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
+checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
 
 [[package]]
 name = "base16ct"
@@ -33,9 +33,9 @@ checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
 
 [[package]]
 name = "base64"
-version = "0.21.5"
+version = "0.21.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "35636a1494ede3b646cc98f74f8e62c773a38a659ebc777a2cf26b9b74171df9"
+checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
 name = "base64ct"
@@ -81,9 +81,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.5.0"
+version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2bd12c1caf447e69cd4528f47f94d203fd2582878ecb9e9465484c4148a8223"
+checksum = "514de17de45fdb8dc022b1a7975556c53c86f9f0aa5f534b98977b171857c2c9"
 
 [[package]]
 name = "cfg-if"
@@ -93,30 +93,30 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.31"
+version = "0.4.38"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f2c685bad3eb3d45a01354cedb7d5faa66194d1d58ba6e267a8de788f79db38"
+checksum = "a21f936df1771bf62b77f047b726c4625ff2e8aa607c01ec06e5a05bd8463401"
 dependencies = [
  "num-traits",
 ]
 
 [[package]]
 name = "const-oid"
-version = "0.9.5"
+version = "0.9.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28c122c3980598d243d63d9a704629a2d748d101f278052ff068be5a4423ab6f"
+checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
 
 [[package]]
 name = "coreum-wasm-sdk"
 version = "0.3.0"
-source = "git+https://github.com/CoreumFoundation/coreum-wasm-sdk.git?branch=keyne/upgrade-wasm-sdk#e1c571716e4cb3c286b320c0d65530dd28af8390"
+source = "git+https://github.com/CoreumFoundation/coreum-wasm-sdk.git?branch=keyne/upgrade-wasm-sdk#766bd3c618c8f5404135d7829a60673ed23b2f3c"
 dependencies = [
  "chrono",
  "cosmwasm-schema",
  "cosmwasm-std",
  "osmosis-std-derive",
- "prost 0.11.9",
- "prost-types 0.12.1",
+ "prost 0.12.4",
+ "prost-types 0.12.4",
  "schemars",
  "serde",
  "serde-cw-value",
@@ -193,9 +193,9 @@ dependencies = [
 
 [[package]]
 name = "cosmwasm-storage"
-version = "1.5.0"
+version = "1.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bd2b4ae72a03e8f56c85df59d172d51d2d7dc9cec6e2bc811e3fb60c588032a4"
+checksum = "66de2ab9db04757bcedef2b5984fbe536903ada4a8a9766717a4a71197ef34f6"
 dependencies = [
  "cosmwasm-std",
  "serde",
@@ -203,18 +203,18 @@ dependencies = [
 
 [[package]]
 name = "cpufeatures"
-version = "0.2.11"
+version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ce420fe07aecd3e67c5f910618fe65e94158f6dcc0adf44e00d69ce2bdfe0fd0"
+checksum = "53fe5e26ff1b7aef8bca9c6080520cfb8d9333c7568e1829cef191a9723e5504"
 dependencies = [
  "libc",
 ]
 
 [[package]]
 name = "crypto-bigint"
-version = "0.5.3"
+version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "740fe28e594155f10cfc383984cbefd529d7396050557148f79cb0f621204124"
+checksum = "0dc92fb57ca44df6db8059111ab3af99a63d5d0f8375d9972e319a379c6bab76"
 dependencies = [
  "generic-array",
  "rand_core 0.6.4",
@@ -282,9 +282,9 @@ dependencies = [
 
 [[package]]
 name = "cw-storage-plus"
-version = "1.1.0"
+version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f0e92a069d62067f3472c62e30adedb4cab1754725c0f2a682b3128d2bf3c79"
+checksum = "d5ff29294ee99373e2cd5fd21786a3c0ced99a52fec2ca347d565489c61b723c"
 dependencies = [
  "cosmwasm-std",
  "schemars",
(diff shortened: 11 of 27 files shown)


masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 3 of 15 files reviewed, 25 unresolved discussions (waiting on @dzmitryhil, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 270 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxH8bLq0hxCT5VX6vta:-NxOfVne7J-JA-SgLpeZ:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L270)):*

extension is not related to signature verification, I think this step can be removed.
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 551 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxHBIjn3Af0RQWLM-Dy:-NxOfO63EijYJQ9LIq7m:b9vi3kb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L551)):*

you don't need this feature.
</blockquote></details>

Done. But it was in the original test
___
*[`integration-tests/modules/assetft_extension_test.go` line 587 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxHBf8G7DPzzGbs_T4_:-NxOfGuX7squZbRbWHLh:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L587)):*

same as above, signature verification is not part of the extension, it is verified in the other test already.
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 829 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxLzb9ZDJTA1Fe6g-NZ:-NxOfCDw4iWNVRgK2-7T:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L829)):*

this test does not interact with extesions, so it seems useless to be here. 2.6315789
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 897 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM-0po6SRXjg3Q2Z_p:-NxOesd12pABEek29Ymr:bktsyd) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L897)):*

the extension contract must have the same burn privileges as the admin, meaning that it can burn tokens even if the feature is not enabled. What we want to test here is only that. I think a lot of test cases that are not related. 
I can only think of a single meaningful test case here, that the contract is able to burn tokens and assert that burning took place as intended.
</blockquote></details>

Our goal was duplicating all tests we already have, but just add the extension to them. I don't think that it hurts if we have those
___
*[`integration-tests/modules/assetft_extension_test.go` line 1105 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1rkd1Uy7Jql2Vfov:-NxOeTXV7mTAG_TlBTCS:b-hta3jh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1105)):*

As discussed for minting, there should be only one test case here, that the contract is able to mint tokens if feature is enabled, and assert that minting happened.
</blockquote></details>

I believe that we should test mintable and non-mintable tokens here. But I agree that tests in which the extension is not involved are not needed
___
*[`integration-tests/modules/assetft_extension_test.go` line 1245 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM-jw2-oOwxfjuBh4_:-NxOeDbe7JHOfDQ1Gq8t:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1245)):*

this check does not interact with extensions, it is not a meaningful check IMO.
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 1259 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1Xjf8qAmFhrorK7F:-NxOdSWD5vmX9RAmkAW5:b-jsfnxg) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1259)):*

we don't want to test minting, we know that it works. we just want to check that the extension contract has privileges to mint tokens if the feature is enabled.
</blockquote></details>

I don't agree. It's an integration test, so if the feature is enabled, it should be able to mint which also means that the contract had the privileges to mint
___
*[`x/asset/ft/keeper/keeper.go` line 690 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM2rauAlmfvStkI5wM:-NxOh32VDYwxNYXO3rUT:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper.go#L690)):*

only admin can transfer admin role, the extension contract is not allowed to do that.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper.go` line 722 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM2yLkAVYKIMiCsn6j:-NxOgzoq18GMHnx7eSJb:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper.go#L722)):*

only admin can do this, not the extension contract.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 160 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM3jNS7W5q27qmDCms:-NxOhnleAWdv-tCEAWVs:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L160)):*

this case does not interact with extensions, it should be removed
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 164 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM3rpW4B6_dx3QqI-l:-NxOhnaq87JzDjs_bLPV:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L164)):*

this case does not interact with extensions, it should be removed
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 169 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM3tB-Cf5B_rs6DWXL:-NxOhmwl2-Y4UVK3LVNG:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L169)):*

this case does not interact with extensions, it should be removed
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 174 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM3uQ26drI0jcva4jz:-NxOhmki9ey8LlElWx6-:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L174)):*

this case does not interact with extensions, it should be removed
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 178 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM3vlZ56YxWu2sbXyC:-NxOhlAy5r5CTw0r24I6:b-iyz07i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L178)):*

this case does not interact with extensions, it should be removed
</blockquote></details>

This one does, doesn't it?
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 54 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM5U6vEyw3mkvXfYks:-NxOnF4E0Q9n7b5oIxwU:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L54)):*

please add this check now. it must check that the attached funds contain the amount.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 122 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM5mc50Rdag4P_-7dw:-NxOp5yC-25QjvC8-5W5:bieggq8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L122)):*

I guess your Admin PR is merged so you should be able to fix it now.
</blockquote></details>

It is not merged yet:
[https://github.com/CoreumFoundation/coreum/pull/827](https://github.com/CoreumFoundation/coreum/pull/827)
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 176 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM66pd97s1OF_zUqRo:-NxOnXBs4RnnpDimejVx:b-drx83d) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L176)):*

this is not accurate, the condition is that the user is not allowed to burn their frozen balance. 
here if you only need check that sender's current balance is above their freezing balance and reject the transaction if it is not. because all the money that must be deduced from the sender's balance is already deducted and given to smart contract.
</blockquote></details>

The TODO is exactly copied from [the doc](https://test.docs.coreum.dev/docs/modules/coreum-fungible-token#:~:text=The%20user%20cannot%20burn%20the%20frozen%20amount%20if%20both%20freezing%20and%20burning%20is%20enabled.)
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 178 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM6eSh0_NJWWyPLkYo:-NxOoaNJ59e8QhBKluhp:b4xbaep) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L178)):*

the smart contract will be able to burn, no matter the feature is enable or not.
</blockquote></details>

It is exactly what you wrote. Isn't it?

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 3 of 15 files reviewed, 30 unresolved discussions (waiting on @masihyeganeh, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 293 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQlQY-CxNQh3sdP3jx:-NxQlQY-CxNQh3sdP3jy:bpae5cm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/integration-tests/modules/assetft_extension_test.go#L293)):*
> ```Go
> 	multiSendMsg := &banktypes.MsgMultiSend{
> 		Inputs:  []banktypes.Input{{Address: issuer.String(), Coins: coinsToSend}},
> 		Outputs: []banktypes.Output{{Address: recipient.String(), Coins: coinsToSend}},
> ```

If you want to test the `MsgMultiSend` use multiple recipients. 
I would propose to use both `MsgSend` with single recipient and `MsgMultiSend` with multiple. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 69 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQjTzW2v98Y25mh1wl:-NxQjTzW2v98Y25mh1wm:bit2ujh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L69)):*
> ```XML
>     // check that amount is present in the attached funds, and attached funds is enough
>     // to cover the transfer.
>     let attached_fund = query_bank_balance(
> ```

The variable name is not correct, the `contract_balance` is better name here. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 89 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQjjA951eY2JxnogZt:-NxQjjAA1jk9lYXa-LDH:bg4vkzb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L89)):*
> ```XML
>         }
> 
>         if features.contains(&assetft::BLOCK_SMART_CONTRACTS) {
> ```

IMO no need to check that feature, since you won't be able to call the contract if the feature is enabled :-)
@miladz68 @ysv @wojtek-coreum WDYT ?
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 99 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQkjis9JbIqUyCQ2Y1:-NxQkjis9JbIqUyCQ2Y2:b-x5w0ww) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L99)):*
> ```XML
>         // meaningful check.
>         if amount == Uint128::new(101) {
>             return assert_burning(
> ```

Could you please explain the `assert_burning` and `assert_minting` usage here?
Because if the burning/minting is disabled no matter whether you do that check here or not, the transaction should fail. So seems that part is not needed.
___
*[`x/asset/ft/types/token.go` line 206 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQl2pl3Czbh5g72omr:-NxQl2pl3Czbh5g72oms:b-g4h40b) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/types/token.go#L206)):*
> ```Go
> 
> // HasAdminPrivileges returns true if the addr is the admin or the asset extension contract address.
> func (def Definition) HasAdminPrivileges(addr sdk.Address) bool {
> ```

nit: What was wrong with prev name?

dzmitryhil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 3 of 15 files reviewed, 31 unresolved discussions (waiting on @masihyeganeh, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 89 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQjjA951eY2JxnogZt:-NxQnynGC0BjsAGf2Y4F:bowp642) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L89)):*

IMO no need to check that feature, since you won't be able to call the contract if the feature is enabled :-)
@miladz68 @ysv @wojtek-coreum WDYT ?
</blockquote></details>

I mean if the extensions are enabled it makes no sense to allow the `BLOCK_SMART_CONTRACTS` feature to be enabled. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 100 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQmw0o0CwLTHKu1e35:-NxQmw0o0CwLTHKu1e36:b6u0uge) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L100)):*
> ```XML
>         if amount == Uint128::new(101) {
>             return assert_burning(
>                 info.sender.as_ref(), amount, &token, features.contains(&assetft::BURNING)
> ```

IMO what is possible to test here is that you can mint/burn when you intercept the send. For example, you can define at the time of the installation and addresses, and if you send from that address you burn the tokens if you send to that address mint `x2`. Or any other example. You can ask @miladz68 probably it's a part of a different task?!

miladz68 (changes requested):
> Reviewed 12 of 12 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 22 unresolved discussions (waiting on @masihyeganeh, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 897 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM-0po6SRXjg3Q2Z_p:-NxRBer13KtiMhQejei7:b-87x5op) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L897)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

Our goal was duplicating all tests we already have, but just add the extension to them. I don't think that it hurts if we have those
</blockquote></details>

Actually our goal was to test extensions, and we took the code from previous tests so we don't write it from scratch. 
___
*[`integration-tests/modules/assetft_extension_test.go` line 1105 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1rkd1Uy7Jql2Vfov:-NxREpZ2BXMktMdO7bDR:b7zu936) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1105)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

I believe that we should test mintable and non-mintable tokens here. But I agree that tests in which the extension is not involved are not needed
</blockquote></details>

agreed mintable and non-mintable tokens should both be tested. 
___
*[`integration-tests/modules/assetft_extension_test.go` line 1259 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1Xjf8qAmFhrorK7F:-NxREzN85vILwXb2vLoP:bhhzlrp) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1259)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

I don't agree. It's an integration test, so if the feature is enabled, it should be able to mint which also means that the contract had the privileges to mint
</blockquote></details>

the minting part is already tested, and the mint message does not interact with extensions. 
___
*[`integration-tests/modules/assetft_extension_test.go` line 1221 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRFCLH-gDvgvh0vN7j:-NxRFCLH-gDvgvh0vN7k:boovmc5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/integration-tests/modules/assetft_extension_test.go#L1221)):*
> ```Go
> // TestAssetFTExtensionSendingToSmartContractIsDenied verifies that this is not possible to send token to smart contract
> // if issuer blocked this operation.
> func TestAssetFTExtensionSendingToSmartContractIsDenied(t *testing.T) {
> ```

Just a reminder, if you keep adding features on top of this PR, it will get so big that you will have hard time finalizing it. 
___
*[`integration-tests/modules/assetft_extension_test.go` line 1329 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRFfsG4v4269B1oYKQ:-NxRFfsG4v4269B1oYKR:b-6iarpi) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/integration-tests/modules/assetft_extension_test.go#L1329)):*
> ```Go
> // TestAssetFTExtensionAttachingToSmartContractIsDenied verifies that this is not possible to attach token to smart
> // contract call if issuer blocked this operation.
> func TestAssetFTExtensionAttachingToSmartContractCallIsDenied(t *testing.T) {
> ```

smart test case :thumbsup: 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 176 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM66pd97s1OF_zUqRo:-NxRICBv-Q2bD_CnELSv:b-pzjw51) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L176)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

The TODO is exactly copied from [the doc](https://test.docs.coreum.dev/docs/modules/coreum-fungible-token#:~:text=The%20user%20cannot%20burn%20the%20frozen%20amount%20if%20both%20freezing%20and%20burning%20is%20enabled.)
</blockquote></details>

yes, the external behavior is as described in the doc, meaning cannot burn. but this is not the user burning the token, it is extension burning the token. the extension can burn the tokens that is in its custody as it sees fit. since the user transferred their tokens to the custody of the extension, the only condition that should be checked here, is that the balance of the user after this transfer is not below their frozen balance. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 178 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM6eSh0_NJWWyPLkYo:-NxRHxD69LOX3QSvzAxx:b-2g7vw1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L178)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

It is exactly what you wrote. Isn't it?
</blockquote></details>

what part did I write ? could you be more specific ?
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 69 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRH2TyBh5w2arBVZSg:-NxRH2TyBh5w2arBVZSh:b-os8xiw) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L69)):*
> ```XML
>     // check that amount is present in the attached funds, and attached funds is enough
>     // to cover the transfer.
>     let attached_fund = query_bank_balance(
> ```

don't query bank balance of the contract, inspect attached funds. it is `info.funds`.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 89 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQjjA951eY2JxnogZt:-NxRD94CCKKQo9jbEmMv:bkejajx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L89)):*

I mean if the extensions are enabled it makes no sense to allow the `BLOCK_SMART_CONTRACTS` feature to be enabled.
</blockquote></details>

it does make sense, we block all other contracts, except the extension contract.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 191 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRDysQ869UaVxkOSKH:-NxRDysQ869UaVxkOSKI:bep31v2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L191)):*
> ```XML
>     // TODO(masih): The user cannot burn the frozen amount if both
>     // freezing and burning is enabled.
>     if !burning_enabled && sender != token.issuer {
> ```

you don't need this check to be able to check that burning works ok. just send 101 tokens with 2 scenarios, burnalbe and non-burnable. In both cases burning should have been done, make sure it was burnt correctly by inspecting the events and total supply. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 249 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRGGhCDEhX7_54p-XY:-NxRGGhCDEhX7_54p-XZ:b-lsngte) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L249)):*
> ```XML
> 
>     // TODO: Do we need this?
>     let issued_from_smart_contract = is_smart_contract(deps, &token.issuer);
> ```

I don't think you need this check, but the extension smart contract must be an exception just like the admin. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/error.rs` line 27 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRGQltDhGIpcyoEJDn:-NxRGQltDhGIpcyoEJDo:b5lfbpa) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/error.rs#L27)):*
> ```XML
>     InsufficientFunds {},
> 
>     // TODO: Delete this one
> ```

apply the TODO.

masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 22 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 1105 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1rkd1Uy7Jql2Vfov:-NxRjVdL1p7AXa8fb9gp:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1105)):*

agreed mintable and non-mintable tokens should both be tested.
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 1259 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1Xjf8qAmFhrorK7F:-NxRjvW1C8iMZBfT9ojk:b-4ve6wx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1259)):*

the minting part is already tested, and the mint message does not interact with extensions.
</blockquote></details>

But it is not calling the native mint method of token. It is simulated with a magic amount and it is actually happened in the extension. So, I believe we need this test. Let me know if you insist on removing it
___
*[`integration-tests/modules/assetft_extension_test.go` line 1221 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRFCLH-gDvgvh0vN7j:-NxRk_Tn7GJfi44Waf6c:b-4a9yqn) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/integration-tests/modules/assetft_extension_test.go#L1221)):*

Just a reminder, if you keep adding features on top of this PR, it will get so big that you will have hard time finalizing it.
</blockquote></details>

You're right. I was actually implementing it while you were reviewing. I won't implement other features in this branch
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 176 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM66pd97s1OF_zUqRo:-NxRobUIBq6vZYZ5g_OA:ba4f2nx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L176)):*

yes, the external behavior is as described in the doc, meaning cannot burn. but this is not the user burning the token, it is extension burning the token. the extension can burn the tokens that is in its custody as it sees fit. since the user transferred their tokens to the custody of the extension, the only condition that should be checked here, is that the balance of the user after this transfer is not below their frozen balance.
</blockquote></details>

So, I'll delete the TODO
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 178 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM6eSh0_NJWWyPLkYo:-NxRoycB741EN7mCsh3k:bvw6i2d) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L178)):*

what part did I write ? could you be more specific ?
</blockquote></details>

I mean, if the sender of the bankSend is the issuer (or admin) of the token, the smart contract can burn it. If I let all of the smart contract calls be able to burn, then any user that sends a bankSend can burn. Let me know if I'm not clear enough
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 69 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQjTzW2v98Y25mh1wl:-NxRnBwzBxOQo4l9nVMT:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L69)):*

The variable name is not correct, the `contract_balance` is better name here.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 69 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRH2TyBh5w2arBVZSg:-NxRnCDr3D0qWPsWCuw3:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L69)):*

don't query bank balance of the contract, inspect attached funds. it is `info.funds`.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 89 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQjjA951eY2JxnogZt:-NxRnT_K3H_2oJJ9ZW6C:bh40hm1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L89)):*

it does make sense, we block all other contracts, except the extension contract.
</blockquote></details>

I agree with Milad
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 99 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQkjis9JbIqUyCQ2Y1:-NxRnfDQ351FmwPlmqGY:bsrlejz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L99)):*

Could you please explain the `assert_burning` and `assert_minting` usage here?
Because if the burning/minting is disabled no matter whether you do that check here or not, the transaction should fail. So seems that part is not needed.
</blockquote></details>

If the asset extension feature is enabled, those checks should happen in the extension, but enabling or disabling them still happens using those feature flags
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 100 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQmw0o0CwLTHKu1e35:-NxRoJSe5RSEhnRgXMiz:biub566) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L100)):*

IMO what is possible to test here is that you can mint/burn when you intercept the send. For example, you can define at the time of the installation and addresses, and if you send from that address you burn the tokens if you send to that address mint `x2`. Or any other example. You can ask @miladz68 probably it's a part of a different task?!
</blockquote></details>

This is exactly what you suggest but when the amount is equal to a magic number, instead of a specific account. Both is possible, but this one lets minting to more accounts and you can test more situations
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 191 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRDysQ869UaVxkOSKH:-NxRpTnW4C6ijHuhbgDe:b-1ezjhc) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L191)):*

you don't need this check to be able to check that burning works ok. just send 101 tokens with 2 scenarios, burnalbe and non-burnable. In both cases burning should have been done, make sure it was burnt correctly by inspecting the events and total supply.
</blockquote></details>

But it is testing more than that. It tests that non-burnables won't burn, but for burnable, you can decide who can burn it. Let me know if you insist on removing it
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/error.rs` line 27 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRGQltDhGIpcyoEJDn:-NxRptxV2lawIm6huuco:bxx12lk) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/error.rs#L27)):*

apply the TODO.
</blockquote></details>

Will do then I finalize this feature
___
*[`x/asset/ft/types/token.go` line 206 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQl2pl3Czbh5g72omr:-NxRpzCT6W2pQaXQ4GvT:bh5taun) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/types/token.go#L206)):*

nit: What was wrong with prev name?
</blockquote></details>

It now checks for both token admin and asset extension admin. It's a new method. We will have the `IsAdmin` one

miladz68 (changes requested):
> Reviewed 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 20 unresolved discussions (waiting on @dzmitryhil, @i, @masihyeganeh, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRqEgcEkAWSHzKMbF1:-NxRqEgcEkAWSHzKMbF2:b-enzhbs) (no related file):*
Run `cargo fmt` at the root of the cargo project to make sure that the files are properly formatted. 

___
___
*[`integration-tests/modules/assetft_extension_test.go` line 1105 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1rkd1Uy7Jql2Vfov:-NxWHhCI-125yndOHdcv:b2pv6mx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1105)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

you are not testing that trying to mint non-mintable tokens are correctly rejected by the go code. you only reject it in the smart contract which is not what we intend to test. 
___
*[`integration-tests/modules/assetft_extension_test.go` line 1259 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1Xjf8qAmFhrorK7F:-NxWI1dR8J9sz9uTkEKx:b-5lz99o) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1259)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

But it is not calling the native mint method of token. It is simulated with a magic amount and it is actually happened in the extension. So, I believe we need this test. Let me know if you insist on removing it
</blockquote></details>

but the code below the comment `mint tokens and check balance and total supply` only tests the native code, which is already tested. I am not saying to remove the entire test, but the parts that don't affect and are not affected by extensions don't need to be tested. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 178 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM6eSh0_NJWWyPLkYo:-NxWHG6h98TQtYyqfUAT:b-bi8pzm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L178)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

I mean, if the sender of the bankSend is the issuer (or admin) of the token, the smart contract can burn it. If I let all of the smart contract calls be able to burn, then any user that sends a bankSend can burn. Let me know if I'm not clear enough
</blockquote></details>

you are clear, and that is the behavior that we want to test, so remove this check. 
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 191 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRDysQ869UaVxkOSKH:-NxWHRex1vAKoVM3n-ju:b-i21nbn) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L191)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

But it is testing more than that. It tests that non-burnables won't burn, but for burnable, you can decide who can burn it. Let me know if you insist on removing it
</blockquote></details>

asset extension must be able to burn both burnables and non-burnables, so the check should be removed.

masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 10 of 16 files reviewed, 19 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRqEgcEkAWSHzKMbF1:-NxXFrX69bIOmCn8ClmY:b-896fix) (no related file):*

Run `cargo fmt` at the root of the cargo project to make sure that the files are properly formatted.
</blockquote></details>

Done.

___
___
*[`integration-tests/modules/assetft_extension_test.go` line 897 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM-0po6SRXjg3Q2Z_p:-NxX1ESd5ZzNwiw1jRNQ:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L897)):*

Actually our goal was to test extensions, and we took the code from previous tests so we don't write it from scratch.
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 1105 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1rkd1Uy7Jql2Vfov:-NxX1bmE-xHuLD9FOVeG:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1105)):*

you are not testing that trying to mint non-mintable tokens are correctly rejected by the go code. you only reject it in the smart contract which is not what we intend to test.
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 293 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQlQY-CxNQh3sdP3jx:-NxTPZoHDDcc1vuqBrxm:b-rurq59) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/integration-tests/modules/assetft_extension_test.go#L293)):*

If you want to test the `MsgMultiSend` use multiple recipients. 
I would propose to use both `MsgSend` with single recipient and `MsgMultiSend` with multiple.
</blockquote></details>

To be honest, this test is copied from the whitelist one to make sure the asset extension is compatible with that
___
*[`x/asset/ft/keeper/keeper.go` line 1073 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM3TZo8L6f9zU9uveS:-NxWfdfMAveyV9KGK9-5:b-jqkpq) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper.go#L1073)):*

I think it should be possible to clawback extension contract balance. let's dicuss in the team.
</blockquote></details>

Done. But in another PR
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 235 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM4Ec_BFmLqNh6B5EC:-NxWi--R8lPcWK3ZjuFQ:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L235)):*

add a special condition into the smart contract to ignore default whitelisting, so we will know that we are able to overwrite the default behavior via smart contract.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 238 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM4QreD2NNiXSE-vis:-NxX7xtI7-OxT30UxolJ:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L238)):*

same as whitelisting, remove the conditions that do not interact with the extension. and add a special condition that shows we are able to overwrite the default behavior via smart contract. 
Same goes for minting and burning.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 438 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM5KZ83bDoeRi-kHkY:-NxX7yZ57Kq-kgqYS5uw:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L438)):*

Should be fixed as commented above.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 597 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM5Nd26pEGtfatRtAH:-NxX7zPR3v7uOyTK8Brm:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/keeper_extension_test.go#L597)):*

Should be fixed as commented above.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 178 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM6eSh0_NJWWyPLkYo:-NxX9T5G4-d_GR9oRu3Z:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L178)):*

you are clear, and that is the behavior that we want to test, so remove this check.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 199 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM6t4CCAVH0MPL2dAy:-NxX9Z8tEbCI16LEUzIa:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L199)):*

maybe remove this check, and let the go code catch and return this error, and also test for it. it will be an interesting test case.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 191 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRDysQ869UaVxkOSKH:-NxX9VxB3Y1gKtE9Ybh8:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L191)):*

asset extension must be able to burn both burnables and non-burnables, so the check should be removed.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/error.rs` line 27 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRGQltDhGIpcyoEJDn:-NxXBNyr5769Td6oHWuH:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/error.rs#L27)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

Will do then I finalize this feature
</blockquote></details>

Done.

miladz68 (changes requested):
> Reviewed 2 of 5 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 15 unresolved discussions (waiting on @dzmitryhil, @i, @masihyeganeh, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 1105 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1rkd1Uy7Jql2Vfov:-NxXWjoP-_Dq4t6fOzYF:b711qe4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1105)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

you are still returning the error on the contract level and not the go level. 
___
*[`integration-tests/modules/assetft_extension_test.go` line 293 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQlQY-CxNQh3sdP3jx:-NxXULFQ6OVDrJXXek_P:b-a0ltzw) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/integration-tests/modules/assetft_extension_test.go#L293)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

To be honest, this test is copied from the whitelist one to make sure the asset extension is compatible with that
</blockquote></details>

but I guess it makes sense to test multiple tokens via multisend, one with extension and one without.
___
*[`integration-tests/modules/assetft_extension_test.go` line 31 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXT06_6xzoJuVrAA0l:-NxXT06_6xzoJuVrAA0m:b-kyqzlx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/integration-tests/modules/assetft_extension_test.go#L31)):*
> ```Go
> 
> const (
> 	MagicAmountDisallowed         = 7
> ```

magic is not a good name maybe `TriggerAmountDisallowed` or `AmountDisallowedTrigger`
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 26 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXUqaC2GhS1rbKmkbV:-NxXUqaC2GhS1rbKmkbW:b-2y8pei) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L26)):*
> ```Go
> 
> const (
> 	MagicAmountDisallowed         = 7
> ```

import these in integration tests if they are accessible, so you are not duplicating them. 
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 223 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXX1Ry9GrwSSE2ihl4:-NxXX1Ry9GrwSSE2ihl5:b-nyli5q) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L223)):*
> ```Go
> 	// sending magic amount will be transferred despise whitelisted amount being exceeded
> 	err = bankKeeper.SendCoins(ctx, issuer, recipient, sdk.NewCoins(
> 		sdk.NewCoin(denom, sdkmath.NewInt(MagicAmountIgnoreWhitelisting))),
> ```

what is the value of  `balance`  and `whitelisted` balance here ?
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 324 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXW3a44wGOqIZdpctg:-NxXW3a44wGOqIZdpcth:b-218m2b) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L324)):*
> ```Go
> 
> 	// send magic amount to transfer despite freezing
> 	err = bankKeeper.SendCoins(ctx, recipient, issuer, sdk.NewCoins(
> ```

what is the value of `(balance - frozen)` here ?
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 438 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXWKvY0dmctHQjk1ol:-NxXWKvY0dmctHQjk1om:b-2uk4a1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L438)):*
> ```Go
> 
> 	// try to burn as non-issuer
> 	err = bankKeeper.SendCoins(ctx, recipient, issuer, sdk.NewCoins(
> ```

also check total supply.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 458 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXWWXd76xlyaFOBkpU:-NxXWWXd76xlyaFOBkpV:b-2uk4a1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L458)):*
> ```Go
> 	// burn tokens and check balance and total supply
> 	err = bankKeeper.SendCoins(ctx, issuer, issuer, sdk.NewCoins(
> 		sdk.NewCoin(burnableDenom, sdkmath.NewInt(MagicAmountBurning))),
> ```

also check total supply.
___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 214 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXVUaD3QuAyNJW_0Gw:-NxXVUaD3QuAyNJW_0Gx:b-asjjsj) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L214)):*
> ```XML
>     });
> 
>     if token.admin != Some(sender.to_string()) {
> ```

remove this check, let the error to be thrown by the go code, so we will know that interface works.

masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 15 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 31 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXT06_6xzoJuVrAA0l:-NxXgeIk-Bn2DeiE8bxB:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/integration-tests/modules/assetft_extension_test.go#L31)):*

magic is not a good name maybe `TriggerAmountDisallowed` or `AmountDisallowedTrigger`
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 26 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXUqaC2GhS1rbKmkbV:-NxXiJzh-q0KkkWp6OoY:b-s97txx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L26)):*

import these in integration tests if they are accessible, so you are not duplicating them.
</blockquote></details>

I don't think I can import them from keeper test. I can have them in a separate file somewhere that both of them can access, but it is not a good idea I think
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 223 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXX1Ry9GrwSSE2ihl4:-NxXv8U10rhR8CzhJuI1:b-bebhis) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L223)):*

what is the value of  `balance`  and `whitelisted` balance here ?
</blockquote></details>

Done. Added the check for those values
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 324 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXW3a44wGOqIZdpctg:-NxXo3aG6TJVPyFNgYA8:b-bebhis) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L324)):*

what is the value of `(balance - frozen)` here ?
</blockquote></details>

Done. Added the check for those values
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 438 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXWKvY0dmctHQjk1ol:-NxXx_0y4XwoLd3Kaltl:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L438)):*

also check total supply.
</blockquote></details>

Done.
___
*[`x/asset/ft/keeper/keeper_extension_test.go` line 458 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXWWXd76xlyaFOBkpU:-NxXxwjgBsps74ZvFTSY:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/keeper_extension_test.go#L458)):*

also check total supply.
</blockquote></details>

Done.

masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 13 of 16 files reviewed, 15 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 214 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxXVUaD3QuAyNJW_0Gw:-NxY0g3ZDByT2YfMV38X:b6j0b4q) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c1331ab3e9468385c46f3d94ca819e3323b175fa/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L214)):*

remove this check, let the error to be thrown by the go code, so we will know that interface works.
</blockquote></details>

It will not throw any error, because the extension will mint the amount and it has privigates to do it. Am I wrong here?

masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 13 of 16 files reviewed, 15 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 1105 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1rkd1Uy7Jql2Vfov:-NxY5Vt6EQ55aL9m9bru:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1105)):*

you are still returning the error on the contract level and not the go level.
</blockquote></details>

Done.
___
*[`integration-tests/modules/assetft_extension_test.go` line 1259 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxM1Xjf8qAmFhrorK7F:-NxY69_6D8lbuQVv2TpF:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/3fc63549dd76169db275760b3cc845aec1e0dff6/integration-tests/modules/assetft_extension_test.go#L1259)):*

but the code below the comment `mint tokens and check balance and total supply` only tests the native code, which is already tested. I am not saying to remove the entire test, but the parts that don't affect and are not affected by extensions don't need to be tested.
</blockquote></details>

Done.

miladz68 (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @i, @masihyeganeh, @wojtek-coreum, and @ysv)

masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 15 of 16 files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs` line 249 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxRGGhCDEhX7_54p-XY:-NxYCiA67A-c4qDF_EXi:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/x/asset/ft/keeper/test-contracts/asset-extension/src/contract.rs#L249)):*

I don't think you need this check, but the extension smart contract must be an exception just like the admin.
</blockquote></details>

Done.

masihyeganeh (commented):
> Reviewed 1 of 3 files at r2, 6 of 12 files at r3, 3 of 6 files at r4, 2 of 5 files at r5, 1 of 4 files at r6, 2 of 3 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

masihyeganeh (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: 15 of 16 files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

___
*[`integration-tests/modules/assetft_extension_test.go` line 293 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/818#-NxQlQY-CxNQh3sdP3jx:-NxYJUVUDEwdX-fk9Ifc:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/b5222a448f733cd22a55fe7bfa4f9588a6e6f7eb/integration-tests/modules/assetft_extension_test.go#L293)):*

but I guess it makes sense to test multiple tokens via multisend, one with extension and one without.
</blockquote></details>

Done.

masihyeganeh (commented):
> Reviewed 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 6 unresolved discussions (waiting on @dzmitryhil, @i, @miladz68, @wojtek-coreum, and @ysv)

miladz68 (approved):
> Reviewed 1 of 1 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: all files reviewed, 5 unresolved discussions (waiting on @dzmitryhil, @i, @wojtek-coreum, and @ysv)

dzmitryhil (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/818)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @wojtek-coreum and @ysv)

PR #825, opened 2024-05-07
Order matching sample with the `filling a small order with a lower price` strategy.

dzmitryhil:
# Description

Implement order matching sample with the `filling a small order with a lower price` strategy.

# Reviewers checklist:

# Authors checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)


ysv (changes requested):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: 1 of 3 files reviewed, 10 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @wojtek-coreum)

ysv (changes requested):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 15 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/keeper/keeper_test.go` line 63 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIlcdX85BZDR0t5t13:-NxIlcdX85BZDR0t5t14:bltdjur) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L63)):*
> ```Go
> 					BuyDenom:     denom1,
> 					SellQuantity: sdkmath.NewInt(30),
> 					Price:        sdkmath.LegacyMustNewDecFromStr("5.1"),
> ```

visibly checking if prices match is a bit complicated.
Because to understand if they match you should calculate `1/Price` using calculator and then compare.

## Approach 1
Maybe we can just put a comment nearby with approximate number so it is easier to compare them visually. Smth like

```
				ID:           "order4",
				SellDenom:    denom2,
				BuyDenom:     denom1,
				SellQuantity: sdkmath.NewInt(30),
				Price:        sdkmath.LegacyMustNewDecFromStr("5.1"), // reversed price: ~0.196

```
It is not mandatory for all cases but in most cases e.g here it make visual comparison more clear. 
Actually that is what I do for each case during review.

## Approach 2:

Maybe to improve readability we can define one side as decimal number and another as fraction. 
E.g here

```
// order1
Price: "0.2"

// order 2
Price: "0.19" ^-1
```
this way we can clearly see that 0.19 is less than 0.2 so no match.
But it is problematic that 0.19^-1 is `5,2631578947... ` so I'm not sure if it makes it more clear or more confusing.


I think I like 1st approach unless we find better DSL for approach 2.
WDYT ?
___
*[`x/dex/keeper/keeper_test.go` line 64 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIq-tZ8kSdNasG7lfN:-NxIq-tZ8kSdNasG7lfO:btot0c4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L64)):*
> ```Go
> 					SellQuantity: sdkmath.NewInt(30),
> 					Price:        sdkmath.LegacyMustNewDecFromStr("5.1"),
> 				},
> ```

related to prev comment. I think 
___
*[`x/dex/keeper/keeper_test.go` line 154 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIomG88NiVCrAzLFvH:-NxIomG88NiVCrAzLFvI:bm66shq) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L154)):*
> ```Go
> 				},
> 			},
> 			expectedBalances: map[string]sdk.Coins{
> ```

nit: I'm not sure if expectedBalances is 100% correct name here since it shows balance increases as result of matching but not full balances as they are.

Maybe smth like: 
1. expectedBalanceChanges but then we should show both increases and decreases which makes sense because then we verify who gets/spends how much. But this complicates tests a bit.
2. expectedBalanceIncreases then we can keep current logic.
___
*[`x/dex/keeper/keeper_test.go` line 247 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxItKkEDYJ_ano3rJPB:-NxItKkEDYJ_ano3rJPC:b-tyuqdl) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L247)):*
> ```Go
> 		},
> 		{
> 			name: "order_rounding_issue_smaller_order_filled_with_lower_than_expected_amount",
> ```

nit: I'm not rounding logic in this case in scope of current PR since rounding will be implemented in the next one.

ok ?
___
*[`x/dex/keeper/keeper_test.go` line 287 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxItUw96BnLKKvtJioh:-NxItUw96BnLKKvtJioi:b-tyuqdl) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L287)):*
> ```Go
> 		},
> 		{
> 			name: "order_rounding_issue_denom_with_high_price_rounded_in_favor_or_higher_volume",
> ```

nit: I'm not rounding logic in this case in scope of current PR since rounding will be implemented in the next one.

ok ?
___
*[`x/dex/keeper/matching_sample.go` line 128 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIqNt_9lci70UFheYx:-NxIqNt_9lci70UFheYy:b-1m04j3) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/matching_sample.go#L128)):*
> ```Go
> LOOP:
> 	for i, makerOrder := range makerOB {
> 		takerBuyPrice := (&big.Rat{}).SetFrac(DecPrecisionReuse, takerOrder.Price.BigInt())
> ```

I would do this a bit differently in terms of names (this improves readability IMO)

```
takerReversePrice := takerOrder.RatReversePrice() // this one returns 1/Price as rational
makerPrice := makerOrder.RatPrice() // this one returns Price as rational number
```

and I would use these 2 names everywhere. More clear IMO & again we avoid buy & sell names which are confusing.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: 2 of 3 files reviewed, 13 unresolved discussions (waiting on @masihyeganeh, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/dex/keeper/keeper_test.go` line 1 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHQXaSCCjomryLqFlX:-NxN80-X4dEFzy3DpZoQ:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/keeper_test.go#L1)):*

nit: more proper name for file is matching_sample_test ?
</blockquote></details>

Done
___
*[`x/dex/keeper/keeper_test.go` line 63 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIlcdX85BZDR0t5t13:-NxN87bR1YDd-B6tg7oL:b-w1642q) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L63)):*

visibly checking if prices match is a bit complicated.
Because to understand if they match you should calculate `1/Price` using calculator and then compare.

## Approach 1
Maybe we can just put a comment nearby with approximate number so it is easier to compare them visually. Smth like

```
				ID:           "order4",
				SellDenom:    denom2,
				BuyDenom:     denom1,
				SellQuantity: sdkmath.NewInt(30),
				Price:        sdkmath.LegacyMustNewDecFromStr("5.1"), // reversed price: ~0.196

```
It is not mandatory for all cases but in most cases e.g here it make visual comparison more clear. 
Actually that is what I do for each case during review.

## Approach 2:

Maybe to improve readability we can define one side as decimal number and another as fraction. 
E.g here

```
// order1
Price: "0.2"

// order 2
Price: "0.19" ^-1
```
this way we can clearly see that 0.19 is less than 0.2 so no match.
But it is problematic that 0.19^-1 is `5,2631578947... ` so I'm not sure if it makes it more clear or more confusing.


I think I like 1st approach unless we find better DSL for approach 2.
WDYT ?
</blockquote></details>

Since we don't have the `side` IMO better to just print the reverted value. Updated.
___
*[`x/dex/keeper/keeper_test.go` line 64 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIq-tZ8kSdNasG7lfN:-NxNA7a8DzQXP8kBtJQi:b9t9ex9) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L64)):*

related to prev comment. I think
</blockquote></details>

Didn't get the comment, explain it please. 
___
*[`x/dex/keeper/keeper_test.go` line 154 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIomG88NiVCrAzLFvH:-NxNAZe3BG03NZa0ACbg:b-s51sgx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L154)):*

nit: I'm not sure if expectedBalances is 100% correct name here since it shows balance increases as result of matching but not full balances as they are.

Maybe smth like: 
1. expectedBalanceChanges but then we should show both increases and decreases which makes sense because then we verify who gets/spends how much. But this complicates tests a bit.
2. expectedBalanceIncreases then we can keep current logic.
</blockquote></details>

Updated to `expectedBalanceIncrease`
___
*[`x/dex/keeper/keeper_test.go` line 247 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxItKkEDYJ_ano3rJPB:-NxNAk9S4TgeXVr7Qocs:b-bsvn8w) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L247)):*

nit: I'm not rounding logic in this case in scope of current PR since rounding will be implemented in the next one.

ok ?
</blockquote></details>

This one with the rounding already.
___
*[`x/dex/keeper/keeper_test.go` line 287 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxItUw96BnLKKvtJioh:-NxNAnIOAx3CGRcgoe-V:b-bsvn8w) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L287)):*

nit: I'm not rounding logic in this case in scope of current PR since rounding will be implemented in the next one.

ok ?
</blockquote></details>

This one with the rounding already.
___
*[`x/dex/keeper/matching_sample.go` line 31 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHVV7rDIRt4n4ZhBXb:-NxNAoLm1ikbpwDP1swP:bby9yuj) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L31)):*

I think we can get rid of usage of LegacyDec completely either in this PR of as a next step.
I suggest to use string to represent price (btw, as I know it is community standard).

And as I can see here we just need a few simple methods on top of this string: convert to BigInt, get length of decimal part.

I'm ok to do this in a separate PR, just want to make sure we are on the same page here.
</blockquote></details>

We need to discuss it, since the topic is quite wide. For the current PR I would keep it as is.
___
*[`x/dex/keeper/matching_sample.go` line 35 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHSZl0A-e-iYDgLZcz:-NxNB5Xy8ebl2GHlxw9P:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L35)):*

isn't it more correct to name these 2 methods as `OrderBookKey` & `ReversedOrderBookKey` ?
These names seems to match business logic better.
</blockquote></details>

Done.
___
*[`x/dex/keeper/matching_sample.go` line 44 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHTzBXBdt2lJAcjqrH:-NxNBD6W2icIxivIyw_J:b-bw8e14) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L44)):*

less than 1.0 you mean ?
</blockquote></details>

Updated.
___
*[`x/dex/keeper/matching_sample.go` line 48 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHTWok-SjbHvPHdaSN:-NxNBHzX2364MS4hV4U6:b-w3a18c) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L48)):*

nit: such statements are a bit complicated to understand as formulas.
I see 2 ways to improve it here:
1. Add comment with formula & explanation.
2. Define helper funcs to do these kind of multiplications

WDYT ?
</blockquote></details>

Added the formula. 
___
*[`x/dex/keeper/matching_sample.go` line 101 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHj66K0Pll2lpxqEQp:-NxNBnAS0elcrkFlsFfx:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L101)):*

nit: I would move this part to the beginning of func so if order is not valid we return fast and do nothing.
</blockquote></details>

Done
___
*[`x/dex/keeper/matching_sample.go` line 124 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHR2qPDmBdlqLD2N3f:-NxNMwoi0tC3yQtIf384:b-ido6ff) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L124)):*

nit: is there a reason to use map here instead of slice ?
</blockquote></details>

Becase it's simplier in that case to skip them when I update the result order book. 
___
*[`x/dex/keeper/matching_sample.go` line 164 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHPOHu-uiPVCYoW0Sd:-NxNMryi3dklnSo8NGSZ:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L164)):*

labeled break (`break LOOP`) complicates code readability here.
Also usually it is a bad practice and we can avoid it here IMO.

I suggest to introduce separate func smth like `app.MatchOrders(takerOrder, makerOrder...)`.
All the logic from this switch case could be moved there.
</blockquote></details>

Done
___
*[`x/dex/keeper/matching_sample.go` line 258 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHiCbA3eoJkLQQZm7a:-NxNE57q3t6AmzWvl-m9:b-xwplht) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L258)):*

where is the round down specified ? or it is default behaviour of Quo ?
</blockquote></details>

Correct. Similar to truncate.
___
*[`x/dex/keeper/matching_sample.go` line 128 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIqNt_9lci70UFheYx:-NxNC3GeBysU5BZ3lvYx:bhxcmdz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/matching_sample.go#L128)):*

I would do this a bit differently in terms of names (this improves readability IMO)

```
takerReversePrice := takerOrder.RatReversePrice() // this one returns 1/Price as rational
makerPrice := makerOrder.RatPrice() // this one returns Price as rational number
```

and I would use these 2 names everywhere. More clear IMO & again we avoid buy & sell names which are confusing.
</blockquote></details>

We anyways have sell and buy in the code, for example for the `RemainingSellQuantity` , so adding additional words doesn't help.

ysv (changes requested):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 4 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/keeper/keeper_test.go` line 63 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIlcdX85BZDR0t5t13:-NxNajbQ64OypAggbDAR:buyht8j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L63)):*

Since we don't have the `side` IMO better to just print the reverted value. Updated.
</blockquote></details>

ok, could you just add reversed price to it pls so it is more clear ?

I would also add ~ (approximate sign) in places where value is not exact
___
*[`x/dex/keeper/matching_sample.go` line 31 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHVV7rDIRt4n4ZhBXb:-NxNb8Hm2Nb2sIFRm7FQ:bafchg) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L31)):*

We need to discuss it, since the topic is quite wide. For the current PR I would keep it as is.
</blockquote></details>

lets discuss
Or if you leave todo here I'm ok to merge as it is now
___
*[`x/dex/keeper/matching_sample.go` line 128 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIqNt_9lci70UFheYx:-NxNa-Cc-c-dFxWBMB5g:b-3chfd8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/matching_sample.go#L128)):*

We anyways have sell and buy in the code, for example for the `RemainingSellQuantity` , so adding additional words doesn't help.
</blockquote></details>

yes, we have sell and buy but both within single order.
While here you name one order takerBuyPrice makerSellPrice which seems like we have buy & sell.
IMO price & reversed price is much more clear because it 100% matches matching algorithm logic.
Also I think we should use it more during discussions because it is very simple & clean to understand. While buy/sell are confusing since we don't have buy & sell sides.

If you want we can discuss
___
*[`x/dex/keeper/matching_sample_test.go` line 126 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxNbtDk8Vqzhkl91Pkq:-NxNbtDk8Vqzhkl91Pkr:b79zmit) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L126)):*
> ```Go
> 
> func (app *App) iterateMakerOrderBook(takerOrder Order, makerOB, takerOB []Order) ([]Order, []Order) {
> 	makerOB, takerOB, makerOBIndexesToRemove := app.matchOrders(takerOrder, makerOB, takerOB)
> ```

I didn't get why do we need to return both makerOB & makerOBIndexesToRemove ?

Why can't we just return makerOB with removed orders or just return makerOBIndexesToRemove and remove from initial value ?
It is better to modify it in a single place either everything inside matchOrders or everything here.

Also you pass both of them into func & slices point to the same memory so I think they will be changed as result. I mean probably there is no need to return them.

ysv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 5 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/keeper/keeper_test.go` line 1 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxNn6Gj8VTH9rFAQH0b:-NxNn6Gj8VTH9rFAQH0c:bb2vqyh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/keeper_test.go#L1)):*
> ```Go
> package keeper_test
> ```

still confusing filenames.
matching_sample_test -> matching_sample // implementation
keeper_test -> matching_sample_test // tests

wojtek-coreum (changes requested):
> Reviewed 1 of 3 files at r1, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 16 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, and @miladz68)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQzfcpAa3fj-VGBwqX:-NxQzfcpAa3fj-VGBwqY:butnn0v) (no related file):*
I think `sdk.Dec` does everything what is required. I don't see reason to use `big`, especially `big.Rat`.

___
___
*[`x/dex/keeper/matching_sample_test.go` line 85 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQynGx8muwRSUCRrk3:-NxQynGx8muwRSUCRrk4:b3dhd2i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L85)):*
> ```Go
> 
> // PlaceOrder places and matches the order into the order book.
> func (app *App) PlaceOrder(takerOrder Order) {
> ```

when order is placed it is not known yet, if it is a taker or maker order, so the variable should just be called `order`.
___
*[`x/dex/keeper/matching_sample_test.go` line 85 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR2JZVEreHIXEFC2d9:-NxR2JZVEreHIXEFC2dA:b-fl1v2u) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L85)):*
> ```Go
> 
> // PlaceOrder places and matches the order into the order book.
> func (app *App) PlaceOrder(takerOrder Order) {
> ```

In this function, it would be good to to lock the funds (take them from `Balances` and store inside `LockedBalances`). And then, in unit tests verify that `LockedBalances` stores expected amoutns.
___
*[`x/dex/keeper/matching_sample_test.go` line 86 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQz0tB1VNjXT_eBCYC:-NxQz0tB1VNjXT_eBCYD:b-9f3jyt) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L86)):*
> ```Go
> // PlaceOrder places and matches the order into the order book.
> func (app *App) PlaceOrder(takerOrder Order) {
> 	fmt.Printf("\nAdding new taker order: %s\n", takerOrder.String())
> ```

taker order -> order
___
*[`x/dex/keeper/matching_sample_test.go` line 91 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR-7Li1iZrHRY_80uR:-NxR-7Li1iZrHRY_80uS:bdlvttb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L91)):*
> ```Go
> 	takerOrder.RemainingSellQuantity = takerOrder.SellQuantity
> 	if takerOrder.IsRemainingBuyQuantityLessThanOne() {
> 		app.CancelOrder(takerOrder)
> ```

There is nothing to cancel at this stage because:
- order hasn't been added to the order book yet
- funds hasn't been locked, so you are sending free coins to the account
___
*[`x/dex/keeper/matching_sample_test.go` line 95 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR3APQ6zr0SCE1S6Qz:-NxR3APQ6zr0SCE1S6R-:b-7m9urp) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L95)):*
> ```Go
> 	}
> 
> 	takerOKKey := takerOrder.OrderBookKey()
> ```

`takerOKKey` -> `takerOBKey`
___
*[`x/dex/keeper/matching_sample_test.go` line 106 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR63BC4G-GIjDqcdkG:-NxR63BC4G-GIjDqcdkH:b-fnvtfv) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L106)):*
> ```Go
> 	}
> 	if len(makerOB) == 0 {
> 		takerOB = append(takerOB, takerOrder)
> ```

I would prefer to move this condition to the `matchOrders` function because then, both standard and special cases are handled next to each other, making everything more clear. Now, when inspecting `matchOrders` I must keep in mind that the special case is handled many lines above.
___
*[`x/dex/keeper/matching_sample_test.go` line 113 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxRBtqB5VJ6OtztD9lY:-NxRBtqC0huyv-Swt9O3:b2v88q6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L113)):*
> ```Go
> 	// sort orders by price
> 	sort.Slice(takerOB, func(i, j int) bool {
> 		return takerOB[i].Price.LTE(takerOB[j].Price)
> ```

I would prefer to sort both order books inside `matchOrders` function. Otherwise, that function returns "not so correct" order books, which are then fixed here. I would prefer order books to be correct from the very beginning.
___
*[`x/dex/keeper/matching_sample_test.go` line 125 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR5Z31EdQ_QT8ihmnV:-NxR5Z31EdQ_QT8ihmnW:b-qk4on7) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L125)):*
> ```Go
> }
> 
> func (app *App) iterateMakerOrderBook(takerOrder Order, makerOB, takerOB []Order) ([]Order, []Order) {
> ```

This function doesn't do what its name says. The logic of removing records from makerOB could be moved to `matchOrders` and then this function might be removed completely.
___
*[`x/dex/keeper/matching_sample_test.go` line 141 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxRChDE7adYVDf_duV_:-NxRChDF0IgG8PhIdmaG:b1t4cim) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L141)):*
> ```Go
> 	makerOBIndexesToRemove := make(map[int]struct{})
> 	for i, makerOrder := range makerOB {
> 		takerBuyPrice := (&big.Rat{}).SetFrac(DecPrecisionReuse, takerOrder.Price.BigInt())
> ```

You may do it once, before the loop
___
*[`x/dex/keeper/matching_sample_test.go` line 201 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxREALW8BJkP5-OghQo:-NxREALW8BJkP5-OghQp:b-ddvbmi) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L201)):*
> ```Go
> 			}
> 			// if nothing to match with add remaining taker order
> 			if len(makerOB) == len(makerOBIndexesToRemove) {
> ```

If this is true it means you iterated all the records in `makerOB`. It means you may do it outside the loop, without `if`

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 16 unresolved discussions (waiting on @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/keeper/keeper_test.go` line 63 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIlcdX85BZDR0t5t13:-NxQqc1J2QT8GL3Bx7LZ:bax2gw6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/keeper_test.go#L63)):*

ok, could you just add reversed price to it pls so it is more clear ?

I would also add ~ (approximate sign) in places where value is not exact
</blockquote></details>

OK. Updated.
___
*[`x/dex/keeper/keeper_test.go` line 1 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxNn6Gj8VTH9rFAQH0b:-NxQox6FCLqJUaRYu5l1:b-8ex2al) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/keeper_test.go#L1)):*

still confusing filenames.
matching_sample_test -> matching_sample // implementation
keeper_test -> matching_sample_test // tests
</blockquote></details>

Clear now. Updated.
___
*[`x/dex/keeper/matching_sample.go` line 31 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxHVV7rDIRt4n4ZhBXb:-NxQqlaZ9QY-4Mof2RFP:bjsof1w) ([raw file](https://github.com/CoreumFoundation/coreum/blob/eeab5397dab1fa3708fc2c68f062f5ccf8a871a2/x/dex/keeper/matching_sample.go#L31)):*

lets discuss
Or if you leave todo here I'm ok to merge as it is now
</blockquote></details>

Left TODO.
___
*[`x/dex/keeper/matching_sample_test.go` line 126 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxNbtDk8Vqzhkl91Pkq:-NxQr6oJ4CsCbDYqovYu:b-jguv3i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L126)):*

I didn't get why do we need to return both makerOB & makerOBIndexesToRemove ?

Why can't we just return makerOB with removed orders or just return makerOBIndexesToRemove and remove from initial value ?
It is better to modify it in a single place either everything inside matchOrders or everything here.

Also you pass both of them into func & slices point to the same memory so I think they will be changed as result. I mean probably there is no need to return them.
</blockquote></details>

The focus of the PR was the matching so the order book manipulation was done in the fastest way. But since it distracts from the matching I've refactored the `App` completely. Check it now.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: 1 of 4 files reviewed, 16 unresolved discussions (waiting on @masihyeganeh, @miladz68, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQzfcpAa3fj-VGBwqX:-NxRMBJXEhEvCG-4ojTI:bbdvq1p) (no related file):*

I think `sdk.Dec` does everything what is required. I don't see reason to use `big`, especially `big.Rat`.
</blockquote></details>

We decided to start writing without the sdk.Dec. The problem with the dec is that it truncates or rounds the numbers, and can't keep it precise all the time we do the calculation. Also some sdk.Dec operations might panic. Also, it doesn't give enough precision 18 is the max, but we might need more. So the price type is the topic we will need to re-discuss.

___
___
*[`x/dex/keeper/matching_sample.go` line 128 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxIqNt_9lci70UFheYx:-NxRPg53DdsWVKUDgNHe:b-3znowh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6dcce52be8ee3cddd8f3fee654cdcb4eec0861ad/x/dex/keeper/matching_sample.go#L128)):*

yes, we have sell and buy but both within single order.
While here you name one order takerBuyPrice makerSellPrice which seems like we have buy & sell.
IMO price & reversed price is much more clear because it 100% matches matching algorithm logic.
Also I think we should use it more during discussions because it is very simple & clean to understand. While buy/sell are confusing since we don't have buy & sell sides.

If you want we can discuss
</blockquote></details>

Updated. Now we use order and revOrder.
___
*[`x/dex/keeper/matching_sample_test.go` line 85 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQynGx8muwRSUCRrk3:-NxROvClEboLfcT5OZB_:b-dlnt20) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L85)):*

when order is placed it is not known yet, if it is a taker or maker order, so the variable should just be called `order`.
</blockquote></details>

Updated. 
___
*[`x/dex/keeper/matching_sample_test.go` line 85 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR2JZVEreHIXEFC2d9:-NxROzci5WOMzJzy6JnX:b-t3xvrx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L85)):*

In this function, it would be good to to lock the funds (take them from `Balances` and store inside `LockedBalances`). And then, in unit tests verify that `LockedBalances` stores expected amoutns.
</blockquote></details>

It works with the assumption that when an order is created the amount is locked in the order and remaining user balance is zero.
___
*[`x/dex/keeper/matching_sample_test.go` line 86 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQz0tB1VNjXT_eBCYC:-NxRPDDH7tkXDVJVFt-o:b-bw8e14) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L86)):*

taker order -> order
</blockquote></details>

Updated.
___
*[`x/dex/keeper/matching_sample_test.go` line 91 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR-7Li1iZrHRY_80uR:-NxRPKAhEa1ex8UZsB9u:b-dlnt20) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L91)):*

There is nothing to cancel at this stage because:
- order hasn't been added to the order book yet
- funds hasn't been locked, so you are sending free coins to the account
</blockquote></details>

Updated. 
___
*[`x/dex/keeper/matching_sample_test.go` line 95 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR3APQ6zr0SCE1S6Qz:-NxRPR_g0U-DqloJQc-m:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L95)):*

`takerOKKey` -> `takerOBKey`
</blockquote></details>

Done.
___
*[`x/dex/keeper/matching_sample_test.go` line 106 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR63BC4G-GIjDqcdkG:-NxRRD8p8MzGaxd-LhBE:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L106)):*

I would prefer to move this condition to the `matchOrders` function because then, both standard and special cases are handled next to each other, making everything more clear. Now, when inspecting `matchOrders` I must keep in mind that the special case is handled many lines above.
</blockquote></details>

Done.
___
*[`x/dex/keeper/matching_sample_test.go` line 113 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxRBtqB5VJ6OtztD9lY:-NxRPT9Q1PqgeWDDUYZV:b-6nq3fc) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L113)):*

I would prefer to sort both order books inside `matchOrders` function. Otherwise, that function returns "not so correct" order books, which are then fixed here. I would prefer order books to be correct from the very beginning.
</blockquote></details>

Updated. Now they are sorted in `OrderBook.AddOrder`.
___
*[`x/dex/keeper/matching_sample_test.go` line 125 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR5Z31EdQ_QT8ihmnV:-NxRPblv6XUFaah84Y8_:byg433c) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L125)):*

This function doesn't do what its name says. The logic of removing records from makerOB could be moved to `matchOrders` and then this function might be removed completely.
</blockquote></details>

Refactored.
___
*[`x/dex/keeper/matching_sample_test.go` line 141 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxRChDE7adYVDf_duV_:-NxRPoAP7sUW6nBNi4t4:bx3t7u0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L141)):*

You may do it once, before the loop
</blockquote></details>

It'll need to take first anyway. Also, the code is refactored, so the `Iterate` is the required part now. 
___
*[`x/dex/keeper/matching_sample_test.go` line 201 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxREALW8BJkP5-OghQo:-NxRQMbY4zVxeYGn8o8M:bgrphqe) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L201)):*

If this is true it means you iterated all the records in `makerOB`. It means you may do it outside the loop, without `if`
</blockquote></details>

I need to add if it's empty and `buyAmount > revSellAmount`, that's why I use that if.

wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 21 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQzfcpAa3fj-VGBwqX:-NxWCA_HCCl60MN4dcDZ:b-yajpad) (no related file):*

We decided to start writing without the sdk.Dec. The problem with the dec is that it truncates or rounds the numbers, and can't keep it precise all the time we do the calculation. Also some sdk.Dec operations might panic. Also, it doesn't give enough precision 18 is the max, but we might need more. So the price type is the topic we will need to re-discuss.
</blockquote></details>

Let's discuss then

___
___
*[`x/dex/keeper/matching_sample.go` line 83 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxW_5Jh-Ioc9yC3IKiT:-NxW_5Jh-Ioc9yC3IKiU:b-gzc43w) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L83)):*
> ```Go
> 
> // OrderBook is order book.
> type OrderBook struct {
> ```

In general, I think the term "order book" is used incorrectly in this PR. "Order book" is made of two sides: "buy" and "sell", or "taker" and "maker" in terms of this PR. So what you call "order book", in reality is "order book side".
___
*[`x/dex/keeper/matching_sample.go` line 102 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWDAUHAND3k9ZUekax:-NxWDAUHAND3k9ZUekay:b-pjpp9v) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L102)):*
> ```Go
> 	i := ob.findRecordIndex(order.Account, order.ID)
> 	if i >= 0 {
> 		panic(fmt.Sprintf("Record with the same account and orderID is already exists in the order book, order:%s", order))
> ```

remove "is"
___
*[`x/dex/keeper/matching_sample.go` line 102 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWE2T07SpJHFwBf9Id:-NxWE2T07SpJHFwBf9Ie:b-ihvssb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L102)):*
> ```Go
> 	i := ob.findRecordIndex(order.Account, order.ID)
> 	if i >= 0 {
> 		panic(fmt.Sprintf("Record with the same account and orderID is already exists in the order book, order:%s", order))
> ```

here and everywhere: Return error instead of panicking
___
*[`x/dex/keeper/matching_sample.go` line 110 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWDTNF-5pU03JcdFYa:-NxWDTNF-5pU03JcdFYb:b5fkzv4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L110)):*
> ```Go
> 		Price:                 order.Price,
> 	}
> 	fmt.Printf("Adding record to the order book, %s/%s, record:%s\n",
> ```

please use the logger to print logs
___
*[`x/dex/keeper/matching_sample.go` line 114 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWFz_QCbVIurU0JTVX:-NxWFz_QCbVIurU0JTVY:b-3hvgox) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L114)):*
> ```Go
> 	)
> 	ob.Records = append(ob.Records, record)
> 	ob.Sort()
> ```

Possible optimization: `ob` is sorted all the time, so the only thing to do is:
1. find the index for new record
2. move all the following records by 1
3. insert new record at the right position

This algorithm is `O(n)` instead of `O(nlogn)` or `O(n^2)` required by sorting.
___
*[`x/dex/keeper/matching_sample.go` line 149 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWHihl2ezllCQFd-2m:-NxWHihl2ezllCQFd-2n:bkdkoy8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L149)):*
> ```Go
> func (ob *OrderBook) Iterate(iterator func(record OrderBookRecord) bool) {
> 	// use copy to allow removal at the time of the iterating
> 	recordCopy := lo.Map(ob.Records, func(record OrderBookRecord, _ int) OrderBookRecord {
> ```

You may remove from the map when iterating. Copy is not needed.
___
*[`x/dex/keeper/matching_sample.go` line 175 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWFRHp1Ou7E9ZcCO6S:-NxWFRHp1Ou7E9ZcCO6T:b5n02mq) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L175)):*
> ```Go
> 
> func (ob *OrderBook) findRecordIndex(account, orderID string) int {
> 	for i, record := range ob.Records {
> ```

This iteration is called a lot in order matching, even when record index is (or might be) known already. I don't know if this implementation is just for testing the logic or it is going to be like this. If it's for testing then it's ok.
___
*[`x/dex/keeper/matching_sample.go` line 180 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWEQmO5ApbpOERxA0c:-NxWEQmO5ApbpOERxA0d:bgt4f4c) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L180)):*
> ```Go
> 		}
> 	}
> 	return -1
> ```

would be better to return NotFound error instead of magic number
___
*[`x/dex/keeper/matching_sample.go` line 184 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWDvIA7Q3dCpLZ2uIq:-NxWDvIA7Q3dCpLZ2uIr:b-dsx9f3) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L184)):*
> ```Go
> 
> // Sort sorts order book records by price desc.
> func (ob *OrderBook) Sort() {
> ```

this method should be private
___
*[`x/dex/keeper/matching_sample.go` line 212 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxW_S_K14ERNmEI8FAW:-NxW_S_K14ERNmEI8FAX:bywx27a) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L212)):*
> ```Go
> 	if order.IsBuyQuantityLessThanOne() {
> 		fmt.Printf("\nOrder cancelled, buy quantity < 1, %s\n", order.String())
> 		app.SendCoin(order.Account, sdk.NewCoin(order.SellDenom, order.SellQuantity))
> ```

Do you assume that at this point, funds have been already locked?
___
*[`x/dex/keeper/matching_sample.go` line 236 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxW_oxQ3dD_QW6qWmeZ:-NxW_oxQ3dD_QW6qWme_:bir8qzl) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L236)):*
> ```Go
> func (app *App) matchOrder(order Order, revOB, ob *OrderBook) {
> 	if revOB.IsEmpty() {
> 		ob.AddOrder(order)
> ```

and return?
___
*[`x/dex/keeper/matching_sample.go` line 240 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWaOi1BfsakwidiF1K:-NxWaOi1BfsakwidiF1L:bevm72v) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L240)):*
> ```Go
> 
> 	revOB.Iterate(func(revOBRecord OrderBookRecord) bool {
> 		buyPrice := (&big.Rat{}).SetFrac(DecPrecisionReuse, order.Price.BigInt())
> ```

this thing might be done once, before calling `Iterate`
___
*[`x/dex/keeper/matching_sample.go` line 302 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWbo3v4xsIIZbLx5ok:-NxWbo3v4xsIIZbLx5ol:b-lg2cbg) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L302)):*
> ```Go
> 				ob.AddOrder(order)
> 				return true
> 			}
> ```

add `default: panic("impossible case")`
___
*[`x/dex/keeper/matching_sample.go` line 305 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWbtc8CZ8efFhTVl16:-NxWbtc8CZ8efFhTVl17:bpf1lxw) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L305)):*
> ```Go
> 		}
> 
> 		return false
> ```

after applying the comment above, this line might be removed
___
*[`x/dex/keeper/matching_sample_test.go` line 85 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR2JZVEreHIXEFC2d9:-NxWcHqu9UBB7VdsMUla:bmqqdjk) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L85)):*

It works with the assumption that when an order is created the amount is locked in the order and remaining user balance is zero.
</blockquote></details>

If this is the assumption, then unlocking should also happen in the external code. Locking and unlocking should be done in the same layer.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 21 unresolved discussions (waiting on @masihyeganeh, @miladz68, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxQzfcpAa3fj-VGBwqX:-NxX9KPJ4SjW-kSbr_-I:b32w0x) (no related file):*

Let's discuss then
</blockquote></details>

OK

___
___
*[`x/dex/keeper/matching_sample.go` line 83 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxW_5Jh-Ioc9yC3IKiT:-NxX9M1aCIjMib3FxwCf:b-xeg9yp) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L83)):*

In general, I think the term "order book" is used incorrectly in this PR. "Order book" is made of two sides: "buy" and "sell", or "taker" and "maker" in terms of this PR. So what you call "order book", in reality is "order book side".
</blockquote></details>

Agree, let's discuss and find a better name to satisfy everyone.
___
*[`x/dex/keeper/matching_sample.go` line 102 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWDAUHAND3k9ZUekax:-NxX9XTV3kCdbg7Jwm7R:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L102)):*

remove "is"
</blockquote></details>

Done.
___
*[`x/dex/keeper/matching_sample.go` line 102 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWE2T07SpJHFwBf9Id:-NxX9Xg292xAKNxRTC66:b-145jcz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L102)):*

here and everywhere: Return error instead of panicking
</blockquote></details>

For the POC of the matching we don't need it, it will just add more lines of code. 
___
*[`x/dex/keeper/matching_sample.go` line 110 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWDTNF-5pU03JcdFYa:-NxX7S4l-20hdlWtJsUB:bkywgh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L110)):*

please use the logger to print logs
</blockquote></details>

This is a sample that shows the matching, so all printing parts will be removed in the implementation. So I don't see a reason to introduce the logger now to remove it soon.
___
*[`x/dex/keeper/matching_sample.go` line 114 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWFz_QCbVIurU0JTVX:-NxX7owZ1bDcDA1YfqlO:b-euzysm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L114)):*

Possible optimization: `ob` is sorted all the time, so the only thing to do is:
1. find the index for new record
2. move all the following records by 1
3. insert new record at the right position

This algorithm is `O(n)` instead of `O(nlogn)` or `O(n^2)` required by sorting.
</blockquote></details>

This the matching POC, not app. The app will be completely reworked using the SDK, so no need to pay attention on it now.
___
*[`x/dex/keeper/matching_sample.go` line 149 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWHihl2ezllCQFd-2m:-NxX8BqgEJfgHHp_jKbX:bb522oz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L149)):*

You may remove from the map when iterating. Copy is not needed.
</blockquote></details>

Didn't get the comment. From which map?
___
*[`x/dex/keeper/matching_sample.go` line 175 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWFRHp1Ou7E9ZcCO6S:-NxX8MbK-mykGdKrwPI1:b-euzysm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L175)):*

This iteration is called a lot in order matching, even when record index is (or might be) known already. I don't know if this implementation is just for testing the logic or it is going to be like this. If it's for testing then it's ok.
</blockquote></details>

This the matching POC, not app. The app will be completely reworked using the SDK, so no need to pay attention on it now.
___
*[`x/dex/keeper/matching_sample.go` line 180 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWEQmO5ApbpOERxA0c:-NxX8TfZDYBTQPWWw5-4:b-euzysm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L180)):*

would be better to return NotFound error instead of magic number
</blockquote></details>

This the matching POC, not app. The app will be completely reworked using the SDK, so no need to pay attention on it now.
___
*[`x/dex/keeper/matching_sample.go` line 184 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWDvIA7Q3dCpLZ2uIq:-NxX8VfUE4fJZTtxratV:b-euzysm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L184)):*

this method should be private
</blockquote></details>

This the matching POC, not app. The app will be completely reworked using the SDK, so no need to pay attention on it now.
___
*[`x/dex/keeper/matching_sample.go` line 212 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxW_S_K14ERNmEI8FAW:-NxX8_Cr0Io0yFNMiN58:btqqu5l) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L212)):*

Do you assume that at this point, funds have been already locked?
</blockquote></details>

Correct.
___
*[`x/dex/keeper/matching_sample.go` line 236 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxW_oxQ3dD_QW6qWmeZ:-NxX8iYK3TbkodqWvyCC:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L236)):*

and return?
</blockquote></details>

Done.
___
*[`x/dex/keeper/matching_sample.go` line 240 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWaOi1BfsakwidiF1K:-NxX8pXy13svMufvMYvx:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L240)):*

this thing might be done once, before calling `Iterate`
</blockquote></details>

Done.
___
*[`x/dex/keeper/matching_sample.go` line 302 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWbo3v4xsIIZbLx5ok:-NxX8sIoCTEKhw-RXekH:b89z8c7) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L302)):*

add `default: panic("impossible case")`
</blockquote></details>

But it's impossible to have the default here.
___
*[`x/dex/keeper/matching_sample.go` line 305 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWbtc8CZ8efFhTVl16:-NxX92Ek55QlaepKFsy9:brm5tsr) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L305)):*

after applying the comment above, this line might be removed
</blockquote></details>

Actually not.
___
*[`x/dex/keeper/matching_sample_test.go` line 85 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR2JZVEreHIXEFC2d9:-NxX9Byp7DthxPKp-Wd9:baz6740) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L85)):*

If this is the assumption, then unlocking should also happen in the external code. Locking and unlocking should be done in the same layer.
</blockquote></details>

We don't have any locking and unlocking :-) 
Only matching, and the balances here show the result of matching.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/825)* status: all files reviewed, 11 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/keeper/matching_sample.go` line 102 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWE2T07SpJHFwBf9Id:-NxkdI7qDVbyeN5b0x4W:b-tajw6l) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L102)):*

For the POC of the matching we don't need it, it will just add more lines of code.
</blockquote></details>

But I believe that this PoC defines the interface which will later be implemented using the real storage. So that interface should be as correct as possible from the very beginning.
___
*[`x/dex/keeper/matching_sample.go` line 149 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWHihl2ezllCQFd-2m:-NxkdgKb5-KqlNs3mCeB:bu3b5v2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L149)):*

Didn't get the comment. From which map?
</blockquote></details>

It's about the comment "// use copy to allow removal at the time of the iterating". It is possible to remove records while iterating.
___
*[`x/dex/keeper/matching_sample.go` line 184 at r6](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxWDvIA7Q3dCpLZ2uIq:-NxkdtMHBsZyPLD9B76m:b-lu92mm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3b9d759b2b6f215f3011cdca938385484f3c718/x/dex/keeper/matching_sample.go#L184)):*

This the matching POC, not app. The app will be completely reworked using the SDK, so no need to pay attention on it now.
</blockquote></details>

okay, but still, this method might be private :-)
___
*[`x/dex/keeper/matching_sample_test.go` line 85 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/825#-NxR2JZVEreHIXEFC2d9:-Nxkec2d5tJygtyH9frp:b-56tf4c) ([raw file](https://github.com/CoreumFoundation/coreum/blob/7c455c4dc4b4f82a231b25336bceca723394f340/x/dex/keeper/matching_sample_test.go#L85)):*

We don't have any locking and unlocking :-) 
Only matching, and the balances here show the result of matching.
</blockquote></details>

You have unlocking whenever you cancel remaining part of the order, e.g. here:
```
if order.IsBuyQuantityLessThanOne() {
				// cancel since nothing to use for the next iteration
				app.SendCoin(order.Account, sdk.NewCoin(order.SellDenom, order.SellQuantity))
				return true
			}
```

If you don't want to lock inside the `App`, then remove those unlockings too.

PR #842, opened 2024-05-16
DRAFT: POC Order matching with ticks

ysv:
# Description

# Reviewers checklist:

# Authors checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/842)

### DIFF
--- a/x/dex/keeper/matching_sample.go
+++ b/x/dex/keeper/matching_sample.go
@@ -20,7 +20,7 @@ var (
 	TenBigInt  = big.NewInt(10)
 	ZeroBigInt = big.NewInt(0)
 
-	TickSizeCoeficient = big.NewRat(1, 100)
+	TickSizeCoefficient = big.NewRat(1, 100)
 )
 
 // Order is matching sample ordrer.
@@ -204,33 +204,33 @@ type App struct {
 	OrderBooks map[string]*OrderBook
 	Balances   map[string]sdk.Coins
 
-	DenomTicks map[string]int64
+	DenomMinAmntIncrements map[string]int64
 }
 
 // NewApp returns new instance of an app.
-func NewApp(denomTicks map[string]int64) *App {
+func NewApp(denomMinAmntIncrements map[string]int64) *App {
 	return &App{
-		OrderBooks: make(map[string]*OrderBook),
-		Balances:   make(map[string]sdk.Coins),
-		DenomTicks: denomTicks,
+		OrderBooks:             make(map[string]*OrderBook),
+		Balances:               make(map[string]sdk.Coins),
+		DenomMinAmntIncrements: denomMinAmntIncrements,
 	}
 }
 
 func (app *App) PriceTickSize(sellDenom, buyDenom string) *big.Rat {
-	// 0.01 *sellDenomTick / buyDenomTick
-	return (&big.Rat{}).Mul(TickSizeCoeficient, big.NewRat(app.DenomTicks[sellDenom], app.DenomTicks[buyDenom]))
+	// 0.01 *sellDenomMinAmntIncrement / buyDenomMinAmntIncrement
+	return (&big.Rat{}).Mul(TickSizeCoefficient, big.NewRat(app.DenomMinAmntIncrements[sellDenom], app.DenomMinAmntIncrements[buyDenom]))
 }
 
 // TODO: Don't use Decimal in calculations.
 func (app *App) ValidateOrder(order Order) error {
-	// Verify that ticks are defined for both denoms.
-	sellTick, ok := app.DenomTicks[order.SellDenom]
+	// Verify that min_amount_increment are defined for both denoms.
+	sellMinAmntIncrement, ok := app.DenomMinAmntIncrements[order.SellDenom]
 	if !ok {
-		return fmt.Errorf("unspecified tick for sell denom: %s", order.SellDenom)
+		return fmt.Errorf("unspecified min_amount_increment for sell denom: %s", order.SellDenom)
 	}
-	buyTick, ok := app.DenomTicks[order.BuyDenom]
+	buyMinAmntIncrement, ok := app.DenomMinAmntIncrements[order.BuyDenom]
 	if !ok {
-		return fmt.Errorf("unspecified tick for buy denom: %s", order.BuyDenom)
+		return fmt.Errorf("unspecified min_amount_increment for buy denom: %s", order.BuyDenom)
 	}
 
 	// check that price is multiple of price tick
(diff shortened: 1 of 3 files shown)


wojtek-coreum (commented):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/842)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`x/dex/spec/matching-ticks.md` line 1 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny4QeMT2r5dh4bHbzjr:-Ny4QeMT2r5dh4bHbzjs:bgznjg0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/9b2d0a0ccc83de59d989434da52edeb047e7696d/x/dex/spec/matching-ticks.md#L1)):*
> ```Markdown
> Article about ticks -
> ```

This idea works mathematically for us, but IMO it doesn't solve any problem - it only moves it to a different place.
Now the big question is: How all the issuers of all those millions of tokens should adjust their `min_ammount_increment` values to make trades possible?
___
*[`x/dex/spec/matching-ticks.md` line 22 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny4NOff0nR3uM95MAOK:-Ny4NOff0nR3uM95MAOL:bh5pyef) ([raw file](https://github.com/CoreumFoundation/coreum/blob/9b2d0a0ccc83de59d989434da52edeb047e7696d/x/dex/spec/matching-ticks.md#L22)):*
> ```Markdown
> In this case, it could be 26666664 since 26666664*0.375=9_999_999 so both numbers are integers.
> But such amount reduces order1 to 40_000_001 AAA (price 0.375) -> 15_000_000.375 BBB.
> As you can see it results into decimal number of BBB in order1 which will cause issues during matching of the next order.
> ```

I don't get why it "will cause issues during matching of the next order". During next matching you may apply the same algorithm to find integer amounts, no?
___
*[`x/dex/spec/matching-ticks.md` line 66 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny4Pd7j4ABVDzDXHXub:-Ny4Pd7j4ABVDzDXHXuc:b-ecrb0v) ([raw file](https://github.com/CoreumFoundation/coreum/blob/9b2d0a0ccc83de59d989434da52edeb047e7696d/x/dex/spec/matching-ticks.md#L66)):*
> ```Markdown
> ```
> 
> Empirically (by experimenting with different assets) `0.01`
> ```

but why do you need this `tick_size_coefficient` at all? Why `1.0` isn't okay?

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/842)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @dzmitryhil and @wojtek-coreum)

___
*[`x/dex/spec/matching-ticks.md` line 1 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny4QeMT2r5dh4bHbzjr:-Ny5EQF50kXwIgr3Wz3c:b-e1elz5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/9b2d0a0ccc83de59d989434da52edeb047e7696d/x/dex/spec/matching-ticks.md#L1)):*
> How all the issuers of all those millions of tokens should adjust their `min_ammount_increment` values to make trades possible

well we discussed this that we consider having 2 approaches for this
1. issuer can adjust it. For sure with some limitations and maybe delay
2. for tokens without issuer  (eg. IBC, Core)  we will have some reasonable default value and feature to override it via gov voting
___
*[`x/dex/spec/matching-ticks.md` line 22 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny4NOff0nR3uM95MAOK:-Ny5FdhE98_p0GgTuR00:b-f6yzfi) ([raw file](https://github.com/CoreumFoundation/coreum/blob/9b2d0a0ccc83de59d989434da52edeb047e7696d/x/dex/spec/matching-ticks.md#L22)):*
rephrased it

> During next matching you may apply the same algorithm to find integer amounts, no?

yes correct, but with such approach when we match as much as possible cancelled part is  probably unpredictable. But we are experimenting with it also because maybe it is the best approach.

The purpose of min_amount_increment is to limit cancelled part and make it more controlled.
Also experimenting with it
___
*[`x/dex/spec/matching-ticks.md` line 66 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny4Pd7j4ABVDzDXHXub:-Ny5GZlJ5-iyQbcuXAXn:b-fq8ogj) ([raw file](https://github.com/CoreumFoundation/coreum/blob/9b2d0a0ccc83de59d989434da52edeb047e7696d/x/dex/spec/matching-ticks.md#L66)):*

but why do you need this `tick_size_coefficient` at all? Why `1.0` isn't okay?
</blockquote></details>

1.0 will not work because in this case assets which has same `min_amount_increment` e.g CORE/SARA could only be traded with integer price.

If we can find a way to make  `tick_size_coefficient` = 1.0 then algorithm would work perfectly.

Basically the problem is not tick_size_coefficient. Main rule which we break here is:
 tick_size(A->B) * tick_size(B->A) == 1.0

if we can stick to this rule then matching would be exact

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/842)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`x/dex/spec/matching-ticks.md` line 87 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny5QCkC2eKLA8_E16or:-Ny5QCkC2eKLA8_E16os:b-ol6wpt) ([raw file](https://github.com/CoreumFoundation/coreum/blob/f2c3d520e7f69396610e5fbb120a565af9166be3/x/dex/spec/matching-ticks.md#L87)):*
> ```Markdown
> | min_amount_increment | 1(B)            | 10(B)          | 100(B)        | 1_000(B) | 10_000(B) | 100_000(B) |
> |----------------------|-----------------|----------------|---------------|----------|-----------|------------|
> | 1(A)                 | 0.01            | 0.1            | 1             | 10       | 100       | 1000       |
> ```

Let's take the first example. How may I trade 1A to 1B without rounding by using tick size of 0.01 specified here? Could you provide full flow?

wojtek-coreum (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/842)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @dzmitryhil and @ysv)

___
*[`x/dex/spec/matching-ticks.md` line 87 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/842#-Ny5QCkC2eKLA8_E16or:-Ny5VaW55uaeR2LgryJL:b2mu1yw) ([raw file](https://github.com/CoreumFoundation/coreum/blob/f2c3d520e7f69396610e5fbb120a565af9166be3/x/dex/spec/matching-ticks.md#L87)):*

Let's take the first example. How may I trade 1A to 1B without rounding by using tick size of 0.01 specified here? Could you provide full flow?
</blockquote></details>

What happen in case someone wants to trade those tokens?

PR #859, opened 2024-05-29
Add significant_amount/max_execution_quantity matching spec and POC

dzmitryhil:
# Description

Add significant_amount/max_execution_quantity matching spec and POC
# Reviewers checklist:

# Authors checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)


wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDL_-I6MDi6X0XIWgQ:bmtvjg2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

Here is the correct proof:

Qa - quantity of token A to trade (always integer)
P = pn / pd (price - real number, pn and pd are always integers)
Qa' - final quantity of token A to be traded (always integer)
Qb' - final quantity of token B to be traded (always integer)

By definition of the price:

Qb' = Qa' * P

Thesis:

Qa' = floor(Qa / pd) * pd
Qb' = Qa' * P  

1. Qb' is always an integer
2. Qb' is the maximum possible integer for any Qa' <= Qa

Proof of (1):

Qa' = floor(Qa / pd) * pd - from thesis
Qb' = Qa' * P - by definition
Qb' = Qa' * pn / pd - by definition
Qb' = floor(Qa / pd) * pd * pn / pd - thesis and definitions combined
Qb' = floor(Qa / pd) * pn - after simplification

Because:
- floor(Qa / pd) is an integer
- pn is an integer

then Qb' is an integer

(1) is proven

Proof of (2) **not being valid**:
To proof that it's not valid it's enough to provide one example where thesis does not hold.

Assumptions:
- Qa = 1234
- P = 0.5, pn = 5, pd = 10

After putting numbers into equation:

Qb' = floor(Qa / pd) * pn
Qb' = floor(1234 /  10) * 5
Qb' = 615

Which means that thesis is false because by simply multiplying 1234 * 0.5 we get 617 which is an integer greater than 615.

How to fix this? I think (but don't know yet how to prove it) is to add one more assumption saying that
P = pn / pd is an irreducible fraction, meaning that pn and pd are integers that have no other common divisors than 1.

After this fix, P must be modified accordingly:

P = 0.5, **pn = 1**, **pd = 2**

by applying those numbers again:

Qb' = floor(Qa / pd) * pn
Qb' = floor(1234 /  2) * 1
Qb' = 617

Obviously it requires a proof for any P.
</blockquote></details>

Actually the proof is very simple:
- we want `floor(Qa / pd)` to be the highest possible number
- so `pd` must be the smallest possible number
- the smallest possible `pd` comes from the irreducible fraction

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDMKBICvtpM2YBQEha:bk7t78i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

Actually the proof is very simple:
- we want `floor(Qa / pd)` to be the highest possible number
- so `pd` must be the smallest possible number
- the smallest possible `pd` comes from the irreducible fraction
</blockquote></details>

There is also an interesting dependency:

if we define:

N = floor(Qa / pd)

then:
- Qa' = N * pd
- Qb' = N * pn

It means, that the price P itself is never used in the formulas.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDN1H_3U4nbX-JAurl:b-vb26zx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

There is also an interesting dependency:

if we define:

N = floor(Qa / pd)

then:
- Qa' = N * pd
- Qb' = N * pn

It means, that the price P itself is never used in the formulas.
</blockquote></details>

It means that once pn and pd are known there are no floating point operations at all. because even `floor(Qa / pd)` is just `Qa / pd` if both numbers are of integer type.

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDItjk2GL59C09guAS:b170g4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

It means that once pn and pd are known there are no floating point operations at all. because even `floor(Qa / pd)` is just `Qa / pd` if both numbers are of integer type.
</blockquote></details>

Good point Wojtek

I agree. I think we should "reduce" nominator and denominator before calculating this.

However, as I can see it already works like this in implementation because big.Rat type by default reduces nominator and denominator. So probably we just need to add this into doc to be the same as implementation.


As for formula I don't like this floor operation here. And there is a way to find biggest divisible integer without floor and division.

In our case we should replace `floor(qa/ pd)` to `qa - (qa mod pd)` where mod is modulus operation.
so in this case it will be: `1234 - (1234 mod 2) = 1234 - 0 = 1234`
This way we don't need floats at all

Also I'm ok to do this change in the next version

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDQTfT-em-Qw1agjFb:b-4bjl5q) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

Good point Wojtek

I agree. I think we should "reduce" nominator and denominator before calculating this.

However, as I can see it already works like this in implementation because big.Rat type by default reduces nominator and denominator. So probably we just need to add this into doc to be the same as implementation.


As for formula I don't like this floor operation here. And there is a way to find biggest divisible integer without floor and division.

In our case we should replace `floor(qa/ pd)` to `qa - (qa mod pd)` where mod is modulus operation.
so in this case it will be: `1234 - (1234 mod 2) = 1234 - 0 = 1234`
This way we don't need floats at all

Also I'm ok to do this change in the next version
</blockquote></details>

eventually that is an idiomatic way of finding greatest divisible int in math `qa - (qa mod pd)`

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDVFDtBGvKULHQ4xIG:b-9wy0o1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*
> as I can see it already works like this in implementation because big.Ra

As far as I understand now, we shouldn't use 'big.Rat' at all, because everything might be done using pure integers

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDWZVB2wngZVRd66y7:bgs6mk7) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*
> In our case we should replace `floor(qa/ pd)` to `qa - (qa mod pd)` where mod is modulus operation.

It's just `Qa / pd` if both numbers are `int`s. In this case the CPU does the flooring for you

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzDWsxX2tPIK97N6YCx:b-7lf6g6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*
> As for formula I don't like this floor operation here

By using `floor` I'm not saying that `math.Floor` must be used. `floor` is simply a defined mathematical operations saying that you round the number down.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzOUYHN3HSjizVgjEtV:b2iad50) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

Yes, `big.Rat` reduces fractions by default:

```go
func main() {
	sdkDec := sdk.MustNewDecFromStr("0.5")
	rat := (&big.Rat{}).SetFrac(sdkDec.BigInt(), DecPrecisionReuse)
	
	fmt.Println(rat.String())
	fmt.Println(rat.Num())
	fmt.Println(rat.Denom())
}
```

It prints:

```
1/2
1
2
```
</blockquote></details>

From what I see the final proof is very similar to the one which is in the spec, at least I don't see a big difference. 
What is regarding calculation, we already use big.Int only, and big.Rat to compare the prices and keep the price. But that can be easily changed to any different type.

wojtek-coreum (changes requested):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzRfQTs9Gwom52UJe1h:b2i8fus) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*
> From what I see the final proof is very similar to the one which is in the spec

I don't see any proof in the spec, only thesis is there.

Also a missing part in my proof is why 

Qa' = floor(Qa / pd) * pd

is the right one to compute the maximum possible `Qa'`, here is the missing proof

Qb' = Qa' * P

we want Qb' to be an integer

Qb' = Qa' * pn / pd

To make `Qb'` an integer `pd` must be reduced to 1 by being reduced with `pn` or `Qa'`.
Because `pn / pd`, by definition, is an irreducible fraction it means `pn` and `pd` already have no common divider other than 1.
It means that `pd` must be reduced fully by `Qa'` exclusively.
It means that `Qa`' must be a multiple of `pd`.

That's why 

Qa' = floor(Qa / pd) * pd

because this formula generates the biggest `Qa'` divisible by `pd`.

ysv (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 14 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, and @miladz68)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSWKZp6d-9Gdq0qQaG:-NzSWKZp6d-9Gdq0qQaH:bsb7no1) (no related file):*
I reviewed the PR but not very meticulously since we will change some parts while moving to cosmos implementation

___
___
*[`x/dex/keeper/matching_sample.go` line 149 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSIqbe19FXHLgXd_0s:-NzSIqbe19FXHLgXd_0t:b5fu9n8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L149)):*
> ```Go
> 
> // GetRecordByAccountAndOrderID returns order book record by account and orderID.
> func (ob *OrderBook) GetRecordByAccountAndOrderID(account, orderBookID string) (bool, OrderBookRecord) {
> ```

nit: usually order of returned values is opposite
`OrderBookRecord, bool`



_Code quote:_
```Go
bool, OrderBookRecord
```
___
*[`x/dex/keeper/matching_sample.go` line 170 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSJG0l9P4EqZ55Z5jz:-NzSJG0l9P4EqZ55Z5k-:b-6iep8d) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L170)):*
> ```Go
> 	})
> 	for _, record := range recordCopy {
> 		if iterator(record) {
> ```

I think usually an iterators is stopped when yield function returns false
Here you stop when true is returned



_Code quote:_
```Go
if iterator(record) {
```
___
*[`x/dex/keeper/matching_sample.go` line 203 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSJnqcBy4RAilQb1RU:-NzSJnqdA538Ke3tnRFO:b-s8jin2) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L203)):*
> ```Go
> func (ob *OrderBook) Sort() {
> 	sort.Slice(ob.Records, func(i, j int) bool {
> 		return ob.Records[i].Price.LTE(ob.Records[j].Price)
> ```

I vote for sorting here by price and time (if price is equal) from day 0.
Or we will not have time but numeric ID for this d?

Because I was experimenting in my implementation and once prices are equal it behave randomly.
___
*[`x/dex/keeper/matching_sample.go` line 259 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSTZgI1cAgkN-GU2Ox:-NzSTZgI1cAgkN-GU2Oy:b7wctql) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L259)):*
> ```Go
> 	}
> 
> 	newOrderRevPriceRat := BigRatQuo(NewBigRatFromInt64(1), NewBigRatFromSDKDec(order.Price))
> ```

nit: I would name this taker. Because as long as it is not stored in orderbook it remains taker.
Also we use taker name everywhere in specs but never use `new order`.



_Code quote:_
```Go
newOrderRevPriceRat
```
___
*[`x/dex/keeper/matching_sample.go` line 280 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSUlBGBM4S9yQcfyUs:-NzSUlBGBM4S9yQcfyUt:b-aepkym) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L280)):*
> ```Go
> 		)
> 
> 		if BigRatGT(makerOBRecordExpectedReceiveQuantityRat, takerQuantityRat) {
> ```

Not sure if it influences anything but
for me personally more logical way seems to compare visa versa values
`makerQuantity vs takerReceiveQuantity`

Since you always try to maximize makerQuantity
we can discuss 



_Code quote:_
```Go
if BigRatGT(makerOBRecordExpectedReceiveQuantityRat, takerQuantityRat) 
```
___
*[`x/dex/keeper/matching_sample.go` line 282 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSL1tfEXRw8JrJOzB6:-NzSL1tfEXRw8JrJOzB7:bkpooyf) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L282)):*
> ```Go
> 		if BigRatGT(makerOBRecordExpectedReceiveQuantityRat, takerQuantityRat) {
> 			// the rev order remains, the taker is reduced fully
> 			makerOBRecordRevPriceRat := BigRatQuo(NewBigRatFromInt64(1), NewBigRatFromSDKDec(makerOBRecord.Price))
> ```

nit: big.Rat has method Reverse or smth like this. You can use it here. It just swaps nominator & denominator
___
*[`x/dex/keeper/matching_sample.go` line 303 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSLIKb4KPcl7DnZyvl:-NzSLIKb4KPcl7DnZyvm:bd7ydrd) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L303)):*
> ```Go
> 		}
> 
> 		maxExecutionQuantity, revMaxExecutionQuantity, remainder := FindMaxExecutionQuantity(
> ```

I can see that you handle 2 cases here:
1. makerQuantity > takerQuantity
2. makerQuantity <= takerQuantity

But maybe 2 should be split into 2 cases. makerQuantity < takerQuantity and makerQuantity == takerQuantity.

When this could result into less cancelation when we have matching quantities of opposite orders.

Example:
order 1: sell 1000AAA for 0.371
order 2: buy 1000AAA for 0.371

then we can match them exactly without rounding. But if we do FindMaxExecutionQuantity then only part will be matched.

I'm not 100% sure if such a situation could happen.
WDYT ?
___
*[`x/dex/keeper/matching_sample.go` line 418 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSNl8mEBSHuI2nKdfP:-NzSNl8mEBSHuI2nKdfQ:b-sq1gxy) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L418)):*
> ```Go
> 	remainder := BigIntSub(quantity, maxExecutionQuantity)
> 
> 	return maxExecutionQuantity, revMaxExecutionQuantity, remainder
> ```

nit: I suggest to rename reminder to `quantityRemainder`
Initially It was unobvious if it is remainder for quantity or revQuantity



_Code quote:_
```Go
remainder
```
___
*[`x/dex/keeper/matching_sample_test.go` line 459 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSXRmf1vZATJml60er:-NzSXRmf1vZATJml60es:b-kyj0lh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample_test.go#L459)):*
> ```Go
> 		},
> 		{
> 			name:           "taker_order_100%_remainder",
> ```

should we allow such cases to happen ?
I know that it opens more possibilities for trading but having order as 100% reminder is not really good.
Or we will prohibit this on UI somehow ?
___
*[`x/dex/spec/matching.md` line 60 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSQVRn4_h82-QKzRc-:-NzSQVRn4_h82-QKzRc0:b-p7hmf) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L60)):*
> ```Markdown
> the remainder.
> 
> The `max_execution_quantity` is the maximum integer value of the `remaining_quantity` (integer) which gives integer when
> ```

The `max_execution_quantity` is the maximum integer value less than `remaining_quantity` (integer) which gives integer when



_Code quote:_
```Markdown
The `max_execution_quantity` is the maximum integer value of the `remaining_quantity` (integer) which gives integer when
```
___
*[`x/dex/spec/matching.md` line 63 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSR7z11Nz1aX_RK8Tl:-NzSR7z11Nz1aX_RK8Tm:b-q0b2bs) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L63)):*
> ```Markdown
> we multiply it by the execution price. The execution price might be the maker order price, if we try to fill the maker
> order `remaining_quantity`. Or one divided by the maker order price, if we try to fill the taker
> order `remaining_quantity`.
> ```

I didn't get this part.
I think it is always maker price. 

`Or one divided by the maker order price, if we try to fill the taker` this we only use for comparison. And it is 1/taker_price



_Code quote:_
```Markdown
we multiply it by the execution price. The execution price might be the maker order price, if we try to fill the maker
order `remaining_quantity`. Or one divided by the maker order price, if we try to fill the taker
order `remaining_quantity`.
```
___
*[`x/dex/spec/matching.md` line 75 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSRieyDCCP-e8sgGN5:-NzSRieyDCCP-e8sgGN6:bbynjmr) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L75)):*
> ```Markdown
> Based on the formula we can define the max not filled quantity. The max not filled quantity depends on the order we
> fill fully. If the order is taker order, the max remainder is equal to the maker price numerator, if maker, the maker
> price denominator.
> ```

isn't it always maker price nominator ?



_Code quote:_
```Markdown
fill fully. If the order is taker order, the max remainder is equal to the maker price numerator, if maker, the maker
price denominator.
```
___
*[`x/dex/spec/matching.md` line 82 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSRrdX5UwCT-33cx0h:-NzSRrdX5UwCT-33cx0i:b-uvmkf6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L82)):*
> ```Markdown
> 
> ```
> if 1/(taker_price) > rev_order_price:
> ```

1/taker_price > maker_price



_Code quote:_
```Markdown
if 1/(taker_price) > rev_order_price:
```

ysv (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 14 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzSd2iT0iEFpbN_gpBu:b-jsoyws) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

> From what I see the final proof is very similar to the one which is in the spec

I don't see any proof in the spec, only thesis is there.

Also a missing part in my proof is why 

Qa' = floor(Qa / pd) * pd

is the right one to compute the maximum possible `Qa'`, here is the missing proof

Qb' = Qa' * P

we want Qb' to be an integer

Qb' = Qa' * pn / pd

To make `Qb'` an integer `pd` must be reduced to 1 by being reduced with `pn` or `Qa'`.
Because `pn / pd`, by definition, is an irreducible fraction it means `pn` and `pd` already have no common divider other than 1.
It means that `pd` must be reduced fully by `Qa'` exclusively.
It means that `Qa`' must be a multiple of `pd`.

That's why 

Qa' = floor(Qa / pd) * pd

because this formula generates the biggest `Qa'` divisible by `pd`.
</blockquote></details>

I'm ok to add Wojteks description into doc
It sounds correct & simple enough

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 13 unresolved discussions (waiting on @masihyeganeh, @miladz68, @wojtek-coreum, and @ysv)

*[a discussion](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSWKZp6d-9Gdq0qQaG:-NzSjh_p--iesVlGfJjr:b-80tlqk) (no related file):*

I reviewed the PR but not very meticulously since we will change some parts while moving to cosmos implementation
</blockquote></details>

Sure.

___
___
*[`x/dex/keeper/matching_sample.go` line 149 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSIqbe19FXHLgXd_0s:-NzSkEafBmj_9kfUkj94:b-bw8e14) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L149)):*

nit: usually order of returned values is opposite
`OrderBookRecord, bool`
</blockquote></details>

Updated.
___
*[`x/dex/keeper/matching_sample.go` line 170 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSJG0l9P4EqZ55Z5jz:-NzSoAjYDPJOdTqCToU9:by4r0s1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L170)):*

I think usually an iterators is stopped when yield function returns false
Here you stop when true is returned
</blockquote></details>

In cosmos sdk `(stop bool) ` true is stop action. 
___
*[`x/dex/keeper/matching_sample.go` line 203 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSJnqcBy4RAilQb1RU:-NzSoOsL9-jV5_lpjrEA:bfnl62i) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L203)):*

I vote for sorting here by price and time (if price is equal) from day 0.
Or we will not have time but numeric ID for this d?

Because I was experimenting in my implementation and once prices are equal it behave randomly.
</blockquote></details>

Actually, we won't have such a problem in cosmos implementation, since each order in the order book will have its own sequential number. But that's out of scope for the PR.
___
*[`x/dex/keeper/matching_sample.go` line 259 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSTZgI1cAgkN-GU2Ox:-NzSovHh0sjYeHZF3sDV:bs823aa) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L259)):*

nit: I would name this taker. Because as long as it is not stored in orderbook it remains taker.
Also we use taker name everywhere in specs but never use `new order`.
</blockquote></details>

It's done intentionally, since it might be a maker or taker.
___
*[`x/dex/keeper/matching_sample.go` line 280 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSUlBGBM4S9yQcfyUs:-NzSp6lu-Tx7EcVyOkWD:b4vdp4l) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L280)):*

Not sure if it influences anything but
for me personally more logical way seems to compare visa versa values
`makerQuantity vs takerReceiveQuantity`

Since you always try to maximize makerQuantity
we can discuss
</blockquote></details>

Acrtaully any can work, but I compare this way since wrote the aglo in the spec with such order of the comparison. 
___
*[`x/dex/keeper/matching_sample.go` line 282 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSL1tfEXRw8JrJOzB6:-NzSpjnN8l9_E0GwFBoo:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L282)):*

nit: big.Rat has method Reverse or smth like this. You can use it here. It just swaps nominator & denominator
</blockquote></details>

Done
___
*[`x/dex/keeper/matching_sample.go` line 303 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSLIKb4KPcl7DnZyvl:-NzSpqfu0Ph4fuQ2re6y:b-3usdij) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L303)):*

I can see that you handle 2 cases here:
1. makerQuantity > takerQuantity
2. makerQuantity <= takerQuantity

But maybe 2 should be split into 2 cases. makerQuantity < takerQuantity and makerQuantity == takerQuantity.

When this could result into less cancelation when we have matching quantities of opposite orders.

Example:
order 1: sell 1000AAA for 0.371
order 2: buy 1000AAA for 0.371

then we can match them exactly without rounding. But if we do FindMaxExecutionQuantity then only part will be matched.

I'm not 100% sure if such a situation could happen.
WDYT ?
</blockquote></details>

The initial version was as you proposed, but later I found that `<=` or `>=` covers it fully and without the rounding if `quantity * price` is integer. If not any option won't help. 
___
*[`x/dex/keeper/matching_sample.go` line 418 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSNl8mEBSHuI2nKdfP:-NzSr4XY4Ngn3Wh0_G3J:b-xdkjbs) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample.go#L418)):*

nit: I suggest to rename reminder to `quantityRemainder`
Initially It was unobvious if it is remainder for quantity or revQuantity
</blockquote></details>

Make sence, updated. 
___
*[`x/dex/keeper/matching_sample_test.go` line 459 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSXRmf1vZATJml60er:-NzSr99j9UiuYgqB35HE:b-a59y87) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/keeper/matching_sample_test.go#L459)):*

should we allow such cases to happen ?
I know that it opens more possibilities for trading but having order as 100% reminder is not really good.
Or we will prohibit this on UI somehow ?
</blockquote></details>

We can't prohibit it. If you create an order with very low volume, the order with greater volume and price with move numbers might cancel it.
___
*[`x/dex/spec/matching.md` line 60 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSQVRn4_h82-QKzRc-:-NzSs0rH4F28GVSs59wd:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L60)):*

The `max_execution_quantity` is the maximum integer value less than `remaining_quantity` (integer) which gives integer when
</blockquote></details>

Done.
___
*[`x/dex/spec/matching.md` line 63 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSR7z11Nz1aX_RK8Tl:-NzSsAeWEfxsZjn5ZcfQ:b-81jy1w) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L63)):*

I didn't get this part.
I think it is always maker price. 

`Or one divided by the maker order price, if we try to fill the taker` this we only use for comparison. And it is 1/taker_price
</blockquote></details>

It's either maker price or 1/maker price. Check the `Matching algorithm` part. 
___
*[`x/dex/spec/matching.md` line 70 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzCqzav0sxFnX0vHPdr:-NzSuiPs7r5u0hZ4XERQ:b9m63co) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L70)):*

I'm ok to add Wojteks description into doc
It sounds correct & simple enough
</blockquote></details>

OK, updated. 
___
*[`x/dex/spec/matching.md` line 75 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSRieyDCCP-e8sgGN5:-NzT68XR5w53ecAbVvi4:b-un4iua) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L75)):*

isn't it always maker price nominator ?
</blockquote></details>

Actually not. The statement is correct.  Check the `Matching algorithm`.
___
*[`x/dex/spec/matching.md` line 82 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSRrdX5UwCT-33cx0h:-NzT6QoO83io_xokm6WB:b97s23y) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L82)):*

1/taker_price > maker_price
</blockquote></details>

It might be not maker price.

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 3 unresolved discussions (waiting on @masihyeganeh, @miladz68, @wojtek-coreum, and @ysv)

___
*[`x/dex/keeper/matching_sample.go` line 236 at r3](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzTU5erDmuWXiou3grS:-NzTbjyfAj249yBX8wM1:b-rrcnvz) ([raw file](https://github.com/CoreumFoundation/coreum/blob/6da01920a9fbadc26fc42ef355dab205d536246b/x/dex/keeper/matching_sample.go#L236)):*

did u decide to change reversed to inversed everywhere or only here ?
I'm ok in general
but what is the reason for this, because we are already used to naming it reversed ?
</blockquote></details>

Because the `Inverse` is 1/x usually. That's why I've renamed it everywhere and in spec as well. 
___
*[`x/dex/spec/matching.md` line 60 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzSQVRn4_h82-QKzRc-:-NzTcfp77jKpkPEkbRQV:bpkx7mn) ([raw file](https://github.com/CoreumFoundation/coreum/blob/429a900626aa958cd631f53b8f3e066450ebd28d/x/dex/spec/matching.md#L60)):*

still no changes visible
</blockquote></details>

You are right, pushed now.

ysv (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @wojtek-coreum)

wojtek-coreum (changes requested):
> Reviewed 1 of 3 files at r2, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, and @miladz68)

___
*[`x/dex/keeper/matching_sample.go` line 287 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwQUrA0ZhKZ4xe77B:-NzWwQUrA0ZhKZ4xe77C:b-2hjfzf) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L287)):*
> ```Go
> 			)
> 			// taker receives
> 			app.SendCoin(takerOrder.Account, sdk.NewCoin(takerOrder.SellDenom, sdk.NewIntFromBigInt(quantityRemainder)))
> ```

Do it only if reminder is positive
___
*[`x/dex/keeper/matching_sample.go` line 308 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwXsZ4ZYAurUfmGW2:-NzWwXsZ4ZYAurUfmGW3:b-2hjfzf) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L308)):*
> ```Go
> 		// maker receives
> 		app.SendCoin(makerOBRecord.Account, sdk.NewCoin(makerOB.BuyDenom, sdk.NewIntFromBigInt(invMaxExecutionQuantity)))
> 		app.SendCoin(makerOBRecord.Account, sdk.NewCoin(makerOB.SellDenom, sdk.NewIntFromBigInt(quantityRemainder)))
> ```

Do it only if reminder is positive
___
*[`x/dex/keeper/matching_sample.go` line 314 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwqZW4R4cJH2o8Huo:-NzWwqZW4R4cJH2o8Hup:bjpmev6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L314)):*
> ```Go
> 		makerOB.RemoveRecord(makerOBRecord)
> 		// update state
> 		takerOrder.Quantity = takerOrder.Quantity.Sub(sdk.NewIntFromBigInt(invMaxExecutionQuantity))
> ```

I think it would be better to have a record in the order book where original Quantity is untouched and only RemainingQuantity is set to updated value

Also if you do it and create `takerOBRecord` object, the code might be deduplicated greatly
___
*[`x/dex/keeper/matching_sample.go` line 380 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWxryO533UxLr6Bjf6:-NzWxryO533UxLr6Bjf7:b6ya5jg) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L380)):*
> ```Go
> 	_, remainder := BigRatToIntWithRemainder(BigRatQuo(priceRat, tickSize))
> 	if !BigIntEqZero(remainder) {
> 		panic(
> ```

Don't panic but return error. Unit tests should verify that error is returned.
___
*[`x/dex/keeper/matching_sample.go` line 413 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWzMV53CweNah2kiLE:-NzWzMV53CweNah2kiLF:b2rdclr) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L413)):*
> ```Go
> 	priceDenom := price.Denom()
> 	// truncate(quantity / priceDenom) * priceDenom
> 	maxExecutionQuantity := BigIntMul(BigIntQuo(quantity, priceDenom), priceDenom)
> ```

It might be simplified:

1. N = floor(quantity / price.Denom())
2. maxExecutionQuantity = N * price.Denom()
3. invMaxExecutionQuantity = N * price.Num()

By doing it this way you don't need Rat operations at all.
___
*[`x/dex/spec/matching.md` line 72 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWttmvDq1a6MoZUARK:-NzWttmw70qCEM0fgXtP:b74r1d9) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/spec/matching.md#L72)):*
> ```Markdown
> Qa' - final quantity of token A to be traded (integer)
> Qb' - final quantity of token B to be traded (integer)
> P = pn / pd, where pn is price numerator (integer) and pd - price denominator (integer)
> ```

please add a note saying that pd/pn is an irreducible fraction
___
*[`x/dex/spec/matching.md` line 91 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWuIwc9IRci2BoGsIL:-NzWuIwc9IRci2BoGsIM:b-2w1fb) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/spec/matching.md#L91)):*
> ```Markdown
> 
> ```
> Qa' = floor(Qa / pd) * pd
> ```

Please also put the final formula for `Qb'` to make it clear for everyone:

Qb' = floor(Qa / pd) * pn

ysv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 7 unresolved discussions (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/keeper/matching_sample.go` line 314 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwqZW4R4cJH2o8Huo:-NzX6Duo9yVKxVIxGxxj:b-eyjo56) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L314)):*

I think it would be better to have a record in the order book where original Quantity is untouched and only RemainingQuantity is set to updated value

Also if you do it and create `takerOBRecord` object, the code might be deduplicated greatly
</blockquote></details>

in general idea makes sense to deduplicated code
but I'm not sure `takerOBRecord` is reasonable since taker can't be in orderbook
___
*[`x/dex/keeper/matching_sample.go` line 380 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWxryO533UxLr6Bjf6:-NzX6ceg8DR8Sn1H5fhM:b-7tdf7d) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L380)):*

Don't panic but return error. Unit tests should verify that error is returned.
</blockquote></details>

agree

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 7 unresolved discussions (waiting on @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/keeper/matching_sample.go` line 287 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwQUrA0ZhKZ4xe77B:-NzX7Z82EPJWkLeHVgsc:bdgug21) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L287)):*

Do it only if reminder is positive
</blockquote></details>

The `SendCoin` checks it inside, otherwise, you will need to check it before almost any `Send`.
___
*[`x/dex/keeper/matching_sample.go` line 308 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwXsZ4ZYAurUfmGW2:-NzX7sD6BgNdCZGD16ka:bdgug21) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L308)):*

Do it only if reminder is positive
</blockquote></details>

The `SendCoin` checks it inside, otherwise, you will need to check it before almost any `Send`.
___
*[`x/dex/keeper/matching_sample.go` line 314 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwqZW4R4cJH2o8Huo:-NzX7z0P86C_kLbRm9c8:b-bc1d8s) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L314)):*

in general idea makes sense to deduplicated code
but I'm not sure `takerOBRecord` is reasonable since taker can't be in orderbook
</blockquote></details>

I had such an idea as well, but found that the code was not clear. We can try to return to the question in the final implementation with the cosmos. 
___
*[`x/dex/keeper/matching_sample.go` line 380 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWxryO533UxLr6Bjf6:-NzX8lL32n4PbuBTF4F7:b-pp3vm6) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L380)):*

agree
</blockquote></details>

OK, update in all the places where I used the panic for better code readability.
___
*[`x/dex/keeper/matching_sample.go` line 413 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWzMV53CweNah2kiLE:-NzXBNfE4NmDn_Hek0IA:bkefpc3) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L413)):*

It might be simplified:

1. N = floor(quantity / price.Denom())
2. maxExecutionQuantity = N * price.Denom()
3. invMaxExecutionQuantity = N * price.Num()

By doing it this way you don't need Rat operations at all.
</blockquote></details>

Correct, and this is how it was done in the initial calculation. But here for simplicity of understanding, I have reproduced the full formula. But agree that it's better to leave the comments and use just ints.
___
*[`x/dex/spec/matching.md` line 72 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWttmvDq1a6MoZUARK:-NzXDozY8eRWc0S_FRCq:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/spec/matching.md#L72)):*

please add a note saying that pd/pn is an irreducible fraction
</blockquote></details>

Done.
___
*[`x/dex/spec/matching.md` line 91 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWuIwc9IRci2BoGsIL:-NzXDpCf1dhJvmzrQ_el:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/spec/matching.md#L91)):*

Please also put the final formula for `Qb'` to make it clear for everyone:

Qb' = floor(Qa / pd) * pn
</blockquote></details>

Done.

wojtek-coreum (changes requested):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 1 unresolved discussion (waiting on @dzmitryhil, @masihyeganeh, @miladz68, and @ysv)

___
*[`x/dex/keeper/matching_sample.go` line 314 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwqZW4R4cJH2o8Huo:-NzXJEyjBqM9_cnCDD5h:b1kh9bx) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L314)):*

I had such an idea as well, but found that the code was not clear. We can try to return to the question in the final implementation with the cosmos.
</blockquote></details>

@ysv `takerOrder`  goes to the order book a few lines later in the code if it is not fully reduced. If we doesn't preserve the original `Quantity` there, then why do we do it in other places by updating `RemainingQuantity` instead of `Quantity`? If you want to differentiate between `Quantity` and `RemainingQuantity` please do it everywhere or remove it from everywhere.
___
*[`x/dex/keeper/matching_sample.go` line 413 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWzMV53CweNah2kiLE:-NzXJxyg6RMV5v68C7fJ:bnz1kn4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L413)):*

Correct, and this is how it was done in the initial calculation. But here for simplicity of understanding, I have reproduced the full formula. But agree that it's better to leave the comments and use just ints.
</blockquote></details>

That's why we follow math first to find the simplest formula to implement. It is much better now, because it clearly shows that we operate on integers, not rational numbers

dzmitryhil (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: all files reviewed, 1 unresolved discussion (waiting on @masihyeganeh, @miladz68, and @wojtek-coreum)

___
*[`x/dex/keeper/matching_sample.go` line 314 at r4](https://reviewable.io/reviews/CoreumFoundation/coreum/859#-NzWwqZW4R4cJH2o8Huo:-NzXOOXJDgYHcDWNvZYN:b-6af2x8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/fc2fb46999657f86f324dab00af3ebb5e10703b7/x/dex/keeper/matching_sample.go#L314)):*

@ysv `takerOrder`  goes to the order book a few lines later in the code if it is not fully reduced. If we doesn't preserve the original `Quantity` there, then why do we do it in other places by updating `RemainingQuantity` instead of `Quantity`? If you want to differentiate between `Quantity` and `RemainingQuantity` please do it everywhere or remove it from everywhere.
</blockquote></details>

The current structures might be significantly different to the cosmos related, that's why I propose not to waste time on this optimization.

wojtek-coreum (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @masihyeganeh and @miladz68)

dzmitryhil (commented):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @masihyeganeh and @miladz68)

ysv (approved):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/859)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @masihyeganeh and @miladz68)

PR #1162, opened 2025-06-03
Integration test export

TxCorpi0x:
# Description

The current change is dependent on [PR#482](https://github.com/CoreumFoundation/crust/pull/482) from crust. 

- [x] Modules hash calculation 
- [ ] Export test success

# Reviewers checklist:

# Authors checklist

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)

### DIFF
--- a/build/coreum/tests.go
+++ b/build/coreum/tests.go
@@ -134,6 +134,11 @@ func runIntegrationTests(
 
 	for _, testDir := range testDirs {
 		var envs []string
+		// the export tests needs to stop znet to export the genesis file
+		// and within the export test we need to access to the binary (to run export command),
+		// and full node home directory(full node is needed to be able to get the stores at certain height)
+		// so it is a special case and to prevent hardcoded path in the export test we set the environment variables
+		// TODO: investigate if we can avoid this environment variables.
 		if testDir == TestExport {
 			config := znet.NewConfig(znetConfig, infra.NewSpec(znetConfig))
 			absRootDir, err := filepath.Abs(znetConfig.RootDir)
--- a/integration-tests/export/export_test.go
+++ b/integration-tests/export/export_test.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/samber/lo"
 	"github.com/stretchr/testify/require"
 
 	sdkstore "cosmossdk.io/core/store"
@@ -208,14 +209,14 @@ func syncAppsHeights(t *testing.T, requireT *require.Assertions, exportedApp *ap
 	require.NoError(t, err)
 }
 
-func checkModuleStoreMismatches(t *testing.T, requireT *require.Assertions, nodeApp *app.App, newApp *app.App, height int64) {
+func checkModuleStoreMismatches(t *testing.T, requireT *require.Assertions, exportedApp *app.App, initiatedApp *app.App, height int64) {
 	var mismatches []string
 
 	// ensure the app contexts are created for the specified height
-	nodeAppCtx := nodeApp.NewUncachedContext(false, cmtproto.Header{Height: height})
-	newAppCtx := newApp.NewUncachedContext(false, cmtproto.Header{Height: height})
+	exportedAppCtx := exportedApp.NewUncachedContext(false, cmtproto.Header{Height: height})
+	initiatedAppCtx := initiatedApp.NewUncachedContext(false, cmtproto.Header{Height: height})
 
-	for _, moduleName := range nodeApp.ModuleManager.ModuleNames() {
+	for _, moduleName := range exportedApp.ModuleManager.ModuleNames() {
 		// skip the module if it is in the ignored list
 		if _, ok := ignoredModuleNames[moduleName]; ok {
 			continue
@@ -242,10 +243,10 @@ func checkModuleStoreMismatches(t *testing.T, requireT *require.Assertions, node
 					t.Logf("Recovered from panic while opening KVStore for %s: %v", storeName, r)
 				}
 			}()
-			nodeAppKvStoreService := runtime.NewKVStoreService(nodeApp.GetKey(storeName))
-			nodeAppKvStore = nodeAppKvStoreService.OpenKVStore(nodeAppCtx)
-			newAppKvStoreService := runtime.NewKVStoreService(newApp.GetKey(storeName))
-			newAppKvStore = newAppKvStoreService.OpenKVStore(newAppCtx)
+			nodeAppKvStoreService := runtime.NewKVStoreService(exportedApp.GetKey(storeName))
+			nodeAppKvStore = nodeAppKvStoreService.OpenKVStore(exportedAppCtx)
+			newAppKvStoreService := runtime.NewKVStoreService(initiatedApp.GetKey(storeName))
+			newAppKvStore = newAppKvStoreService.OpenKVStore(initiatedAppCtx)
 		}()
 
 		if nodeAppKvStore == nil {
@@ -296,14 +297,9 @@ func compareKVStores(exportedAppStore, initiatedAppStore sdkstore.KVStore, ignor
 	// Check for keys in exportedMap not in initiatedMap or with different values
 	for k, v := range exportedMap {
 		// Skip keys with any of the ignorePrefixes
-		ignored := false
-		for _, prefix := range ignorePrefixes {
-			if bytes.HasPrefix([]byte(k), prefix) {
-				ignored = true
-				break
-			}
-		}
-		if ignored {
+		if lo.ContainsBy(ignorePrefixes, func(prefix []byte) bool {
+			return bytes.HasPrefix([]byte(k), prefix)
+		}) {
 			continue
 		}
 


miladz68 (changes requested):
> Reviewed 5 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: 5 of 8 files reviewed, 1 unresolved discussion (waiting on @masihyeganeh and @ysv)

___
*[`Makefile` line 76 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-ORu0goy2H5byqfgrKbV:-ORu0goy2H5byqfgrKbW:b-nse6wv) ([raw file](https://github.com/CoreumFoundation/coreum/blob/f4d47319dd2d87b9939c0339b42fa3c537653e98/Makefile#L76)):*
> ```Makefile
> 
> .PHONY: integration-tests-export
> integration-tests-export:
> ```

this export does not contain any integration tests and is less likely to catch bugs. the export should be an step in normal integration test and we should run only after the upgrade test (since upgrade test contains all other tests)

masihyeganeh (commented):
> Reviewed 6 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @TxCorpi0x and @ysv)

ysv (changes requested):
> Reviewed 17 of 19 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: 17 of 19 files reviewed, 7 unresolved discussions (waiting on @masihyeganeh and @miladz68)

___
*[`Makefile` line 76 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-ORu0goy2H5byqfgrKbV:-OT746UX4gE9BfomWhgc:b-6b9qtd) ([raw file](https://github.com/CoreumFoundation/coreum/blob/f4d47319dd2d87b9939c0339b42fa3c537653e98/Makefile#L76)):*

this export does not contain any integration tests and is less likely to catch bugs. the export should be an step in normal integration test and we should run only after the upgrade test (since upgrade test contains all other tests)
</blockquote></details>

I agree, lets run it after upgrade. Maybe also makes sense to be able to run it after modules & IBC integration tests
___
*[`build/coreum/tests.go` line 144 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OT75PTI1uwzBoEAGECp:-OT75PTJ6W2UOimRfLjm:b-juxwnm) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/build/coreum/tests.go#L144)):*
> ```Go
> 			}
> 
> 			envs = []string{
> ```

could u pls explain why we need this ?
___
*[`x/dex/keeper/keeper.go` line 460 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBy1V41rxWZKgMmecO:-OTBy1V41rxWZKgMmecP:b-xjfxwo) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/dex/keeper/keeper.go#L460)):*
> ```Go
> 	keys := make([][]byte, 0)
> 	for ; iterator.Valid(); iterator.Next() {
> 		keys = append(keys, iterator.Key())
> ```

so we had this obvious bug here, right ?
___
*[`integration-tests/export/export_test.go` line 46 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBvdwV3VU0X0NlWwvt:-OTBvdwV3VU0X0NlWwvu:byb1n2u) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L46)):*
> ```Go
> 
> func init() {
> 	flag.BoolVar(&runUnsafe, "run-unsafe", false, "run unsafe tests for example ones related to governance")
> ```

but do we really need/use this run-unsafe option for genesis import-export tests ?
___
*[`x/delay/keeper/keeper.go` line 103 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBx9LUAp0p2nKnX8z8:-OTBx9LV7RBmQQvuno6B:b-tcy4zo) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/delay/keeper/keeper.go#L103)):*
> ```Go
> 	}
> 	if val {
> 		if skipDuplicate {
> ```

I'm not sure that is the proper way to fix this issue. 
ID of delayed execution should be completely unique IMO

In which case can we have duplicate IDs ?
___
*[`x/delay/keeper/keeper.go` line 103 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTByLKo1I2SdKeio8nf:-OTByLKo1I2SdKeio8ng:bmwpx9e) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/delay/keeper/keeper.go#L103)):*
> ```Go
> 	}
> 	if val {
> 		if skipDuplicate {
> ```

Also, as far as I recall there is a similar module in osmosis chain which handles this functionality.
Maybe we can check how they deal with duplicate ids there
___
*[`x/asset/ft/genesis.go` line 76 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBymAD71pALUgyajnw:-OTBymAD71pALUgyajnx:b-ea8lxn) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/asset/ft/genesis.go#L76)):*
> ```Go
> 	// Init DEX locked balances
> 	for _, dexLockedBalance := range genState.DEXLockedBalances {
> 		// TODO: investigate why two part denom validation is not working
> ```

so we still have this issue present ?
Do we need a separate task to fix this one ?

TxCorpi0x (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: 16 of 19 files reviewed, 19 unresolved discussions (waiting on @masihyeganeh, @miladz68, and @ysv)

___
*[`Makefile` line 76 at r1](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-ORu0goy2H5byqfgrKbV:-OTkpxxt3EmAehXe83qT:b2hrw4) ([raw file](https://github.com/CoreumFoundation/coreum/blob/f4d47319dd2d87b9939c0339b42fa3c537653e98/Makefile#L76)):*

I agree, lets run it after upgrade. Maybe also makes sense to be able to run it after modules & IBC integration tests
</blockquote></details>

It was a temporary change for local debugging, it is removed now.
___
*[`build/coreum/tests.go` line 144 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OT75PTI1uwzBoEAGECp:-OTkq4gvAYgkqvmL4UPR:b-n4bmk1) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/build/coreum/tests.go#L144)):*

could u pls explain why we need this ?
</blockquote></details>

The export command should use CLI of the cored binary to issue the export. the root path of the crust as well as binary is not accessible from inside the integration tests.
so there are two ways:
1. assume the home is always home directory of user
2. This way and pass the znet home and cored path to the integration tests using environment variables
___
*[`integration-tests/export/export_test.go` line 21 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCH15F3DPZAHzUotbh:-OTkqnZY75cB02-HxbXD:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L21)):*

sdkstore is better name here because core is similar to cored
</blockquote></details>

Done
___
*[`integration-tests/export/export_test.go` line 46 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBvdwV3VU0X0NlWwvt:-OTkqp1A8Z9KQNw30ijo:b9hvhjo) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L46)):*

but do we really need/use this run-unsafe option for genesis import-export tests ?
</blockquote></details>

The flag is passed by integration tests, this code is to prevent panic because of not handling of the flag.
___
*[`integration-tests/export/export_test.go` line 77 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCLWuk-xpYRvrRLZYj:-OU-LqZECxr6YVZVqYxz:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L77)):*

worth adding comment of what is happening inside this func
</blockquote></details>

Done
___
*[`integration-tests/export/export_test.go` line 89 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCMD_BEKxlSDDtP_KY:-OTkr2J49rSSARox8ScR:b11lf5n) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L89)):*

why do we have cored-05-full hardcoded here ?
Does it mean that we will not be able to run export tests for other profiles ?
</blockquote></details>

Yes, export is possible for a full node (not pruned state)
___
*[`integration-tests/export/export_test.go` line 97 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCNj875a7HCZDCVtaE:-OTkr8pe93qLpKNg0FIm:bqtw2qn) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L97)):*

Ideally, would be nice to not have all this bash commands calls inside cored but I'm not sure it is achievable

Have to think about it
</blockquote></details>

May you please elaborate, cored means the cmd package of the coreum code?
___
*[`integration-tests/export/export_test.go` line 104 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCNnL87Tx26UMB_yk7:-OU-Lo5wDey1IwWwAs8_:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L104)):*

do you call shell `cp` here ?
Should we use golang file copy instead ?
</blockquote></details>

Done
___
*[`integration-tests/export/export_test.go` line 114 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCLO6B89id9i7YSOED:-OTkrySdDYlPqxSkxNEK:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L114)):*

should be removed/uncommented ?
</blockquote></details>

Done.
___
*[`integration-tests/export/export_test.go` line 126 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCIV2PBjEJ4rGOiOOJ:-OU-Lgzc-rE3DwR0TOkM:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L126)):*

could you pls clarify what is the difference between nodeApp  vs newApp ?
I think it is worth adding comment for it.
Also newApp name doesn't seem like the best, maybe it should be simApp ?
</blockquote></details>

Done
___
*[`integration-tests/export/export_test.go` line 160 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCMPePEhRxyvVjDXLa:-OU-LiGa2cypkU_WytgQ:b-otj2ev) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L160)):*

I see clear separation of this function into 2 (possibly even 3) separate funcs:
1. preparation of files and directories
2. app initialization and comparison itself

I propose we split it
</blockquote></details>

Splitted into multiple functions
___
*[`integration-tests/export/export_test.go` line 288 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCFBNlAWkUcCmtgmhw:-OTkwo9tDzvAviz-7L68:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L288)):*

nit: these 2 variable names are too similar `ignorePrefixes` & `ignoredPrefixes`

This causes lots of confusing while trying to understand code
</blockquote></details>

Done
___
*[`integration-tests/export/export_test.go` line 294 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCFyN4AfJx7VcxZ8fh:-OTkvK6AELeo1SO1u9fJ:b-dcm9gf) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L294)):*

IMO it is worth moving this to a separate non-anonymous func
</blockquote></details>

It is intentional to handle the panic from this block only, but if you preferm we can move all of the closure code to separate funtion. i thought it would be more readable.
```
nodeAppKvStoreService := runtime.NewKVStoreService(nodeApp.GetKey(storeName))
			nodeAppKvStore = nodeAppKvStoreService.OpenKVStore(nodeAppCtx)
			newAppKvStoreService := runtime.NewKVStoreService(newApp.GetKey(storeName))
			newAppKvStore = newAppKvStoreService.OpenKVStore(newAppCtx)
```
 
___
*[`integration-tests/export/export_test.go` line 315 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCGNc39bd21Znp0tkC:-OTksbywBfLy4cJZheaQ:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L315)):*

simplified statement
```
if nodeAppKvStore == nil {
			if newAppKvStore != nil {
mismatches = append(mismatches, fmt.Sprintf("KVStore %s not found in node app", storeName))
				
			}
                  continue
		}
```
</blockquote></details>

Done.
___
*[`integration-tests/export/export_test.go` line 325 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCH9PyC4c5uOBb8i63:-OTkuHkL8iwojJm9caE4:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L325)):*

since we already pass `testing.T` into this func maybe it is worth using it to assert for errors/mismatches absence.

Also this way you can avoid returning of error from this func
</blockquote></details>

Done
___
*[`x/asset/ft/genesis.go` line 76 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBymAD71pALUgyajnw:-OU-Lw8GADa7AHMffYOe:bo8cg2q) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/asset/ft/genesis.go#L76)):*

so we still have this issue present ?
Do we need a separate task to fix this one ?
</blockquote></details>

I will read and implement proper validation in this PR 
___
*[`x/delay/keeper/keeper.go` line 103 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBx9LUAp0p2nKnX8z8:-OTkuTcD0dNwnwZ8l6Lg:bih91oq) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/delay/keeper/keeper.go#L103)):*

I'm not sure that is the proper way to fix this issue. 
ID of delayed execution should be completely unique IMO

In which case can we have duplicate IDs ?
</blockquote></details>

In the init genesis, for the ones that already present in the dalay module state, those will be added by delay module initgenesis as well as dex module initgenesis `AfterBlock` call.
___
*[`x/delay/keeper/keeper.go` line 103 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTByLKo1I2SdKeio8nf:-OTkv3L29WhNEGzFlbC8:b-2k3td5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/delay/keeper/keeper.go#L103)):*

Also, as far as I recall there is a similar module in osmosis chain which handles this functionality.
Maybe we can check how they deal with duplicate ids there
</blockquote></details>

TODO: I will have a look into that
___
*[`x/dex/keeper/keeper.go` line 460 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBy1V41rxWZKgMmecO:-OTkuy8kDcOgq1bUXG2-:b-ok5syy) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/dex/keeper/keeper.go#L460)):*

so we had this obvious bug here, right ?
</blockquote></details>

Yes, it exports values into keys instead.

ysv (changes requested):
> Reviewed 1 of 1 files at r3, 2 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: all files reviewed, 12 unresolved discussions (waiting on @masihyeganeh, @miladz68, and @TxCorpi0x)

___
*[`build/coreum/tests.go` line 144 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OT75PTI1uwzBoEAGECp:-OU5zSjACTWeSGnpBx6h:bgx047t) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/build/coreum/tests.go#L144)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

The export command should use CLI of the cored binary to issue the export. the root path of the crust as well as binary is not accessible from inside the integration tests.
so there are two ways:
1. assume the home is always home directory of user
2. This way and pass the znet home and cored path to the integration tests using environment variables
</blockquote></details>

I see, makes sense.
I'm not really sure if there is a way to quickly resolve this without significant changes so
Pls leave TODO here explaining why this is needed and that we want to avoid calling binaries from integration-tests
___
*[`integration-tests/export/export_test.go` line 46 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBvdwV3VU0X0NlWwvt:-OU5MiaH3794vB1sG4Ta:bxkr9m8) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L46)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

The flag is passed by integration tests, this code is to prevent panic because of not handling of the flag.
</blockquote></details>

and how do we get this handled in other integration-test pkgs (modules, upgrade) ? Because we also have runUnsafe there ?
___
*[`integration-tests/export/export_test.go` line 89 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCMD_BEKxlSDDtP_KY:-OU5N5x_-FMR-NdvCXg0:b-hutzse) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L89)):*
> I really don't like this hardcoded value. Is there any way around it?

Same here. We should not hardcode it somehow, any idea ?
___
*[`integration-tests/export/export_test.go` line 97 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCNj875a7HCZDCVtaE:-OU5NG8f6KcGQo45Yfkd:b-1s66q0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L97)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

May you please elaborate, cored means the cmd package of the coreum code?
</blockquote></details>

I mean would be nice to not call bash commands (e.g `cored`, cp etc) from integration tests 
___
*[`integration-tests/export/export_test.go` line 126 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCIV2PBjEJ4rGOiOOJ:-OU5O90w5A-cgzyA9a3M:b-xse975) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L126)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

Done
</blockquote></details>

I can still see `newApp *app.App` & `newAppCtx` names used
___
*[`integration-tests/export/export_test.go` line 160 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCMPePEhRxyvVjDXLa:-OU5OOb08bRDk0-HjjWu:bnt655) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L160)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

Splitted into multiple functions
</blockquote></details>

looks better :thumbsup: 
___
*[`x/delay/keeper/keeper.go` line 103 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBx9LUAp0p2nKnX8z8:-OU5ZH-G5Fn4Fw88AeP0:b-xzon3z) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/delay/keeper/keeper.go#L103)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

In the init genesis, for the ones that already present in the dalay module state, those will be added by delay module initgenesis as well as dex module initgenesis `AfterBlock` call.
</blockquote></details>

I see, so the same method is called inside DEX module when we:
- place order via API (delay should be registered)
- export genesis (delay shouldn't be registered)

Shall we handle it on DEX level instead ?
___
*[`x/dex/keeper/keeper.go` line 460 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBy1V41rxWZKgMmecO:-OU5Pp8z8GKJM_CZ8jQB:bzqm1q) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/dex/keeper/keeper.go#L460)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

Yes, it exports values into keys instead.
</blockquote></details>

good catch !
___
*[`integration-tests/export/export_test.go` line 241 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OU5UFO89b3w92sx1cGK:-OU5UFO9BIzR1M_FiKuH:brpfj0p) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c6874a9c5bd5161fb8a2967981e5bc323c771a89/integration-tests/export/export_test.go#L241)):*
> ```Go
> 		func() {
> 			defer func() {
> 				if r := recover(); r != nil {
> ```

is it safe to recover here or test should fail ?
does this happen regularly/randomly ?
___
*[`integration-tests/export/export_test.go` line 308 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OU5SwEr0YvT_87kl6OC:-OU5SwEs8F5PyGVL9SHY:btxi7mu) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c6874a9c5bd5161fb8a2967981e5bc323c771a89/integration-tests/export/export_test.go#L308)):*
> ```Go
> 		if ignored {
> 			continue
> 		}
> ```

nit: we can use lo.Contains in both places to make this code more readable

```
if lo.Contains(ignoredPrefixes, prefix, func()...) {
  continue
}
```



_Code quote:_
```Go
		ignored := false
		for _, prefix := range ignorePrefixes {
			if bytes.HasPrefix([]byte(k), prefix) {
				ignored = true
				break
			}
		}
		if ignored {
			continue
		}
```

TxCorpi0x (commented):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: all files reviewed, 11 unresolved discussions (waiting on @masihyeganeh, @miladz68, and @ysv)

___
*[`integration-tests/export/export_test.go` line 46 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBvdwV3VU0X0NlWwvt:-OU9RAQ845SZ1ruo7H3L:b-se4qi5) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L46)):*

and how do we get this handled in other integration-test pkgs (modules, upgrade) ? Because we also have runUnsafe there ?
</blockquote></details>

The rest of the integration tests use `NewChainsTestingContext` and in the init function of the parent package the flag is handled.
but in export (because the chain is stopped and there is no grpc/rpc/rest available) we can not use normal integration tests objects and types.
___
*[`integration-tests/export/export_test.go` line 89 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCMD_BEKxlSDDtP_KY:-OU9RdkEApW999Kgy9aX:b-7f1lft) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L89)):*

> I really don't like this hardcoded value. Is there any way around it?

Same here. We should not hardcode it somehow, any idea ?
</blockquote></details>

Agreed, It would be great to have a link (env or any better way) paths from crust to be usable within tests/integration-tests
___
*[`integration-tests/export/export_test.go` line 97 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCNj875a7HCZDCVtaE:-OU9Ry2jDuqJyvwuC5-g:b9z7es0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L97)):*

I mean would be nice to not call bash commands (e.g `cored`, cp etc) from integration tests
</blockquote></details>

that is not possible normally (it needs coreum app cmd to gets modified and have an exported function to be used here).
and also it is better to use bash because this is the normal way of export (there is no grpc handler) and the chain/node is stopped and should be stopped for the command to gets executed.
___
*[`integration-tests/export/export_test.go` line 126 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCIV2PBjEJ4rGOiOOJ:-OU9SebKEeeYP575AhQw:b-896fix) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L126)):*

I can still see `newApp *app.App` & `newAppCtx` names used
</blockquote></details>

Done.
___
*[`integration-tests/export/export_test.go` line 241 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OU5UFO89b3w92sx1cGK:-OU9Sh339FGv7OK9X568:b-8cdiwh) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c6874a9c5bd5161fb8a2967981e5bc323c771a89/integration-tests/export/export_test.go#L241)):*

is it safe to recover here or test should fail ?
does this happen regularly/randomly ?
</blockquote></details>

It happens in OpenKVStore which does not return any error, and it panics. normally we don't call this method in the blockchain because we know which stores we have initialized in the app module.
___
*[`integration-tests/export/export_test.go` line 308 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OU5SwEr0YvT_87kl6OC:-OU9VIcKCHHwaZGGhir-:bb6un1j) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c6874a9c5bd5161fb8a2967981e5bc323c771a89/integration-tests/export/export_test.go#L308)):*

nit: we can use lo.Contains in both places to make this code more readable

```
if lo.Contains(ignoredPrefixes, prefix, func()...) {
  continue
}
```
</blockquote></details>

Done
___
*[`proto/coreum/feemodel/v1/genesis.proto` line 22 at r5](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OU4Uy5v2Ww1-05ayKrH:-OU9VmZ-6MXjZyYrGlHW:bnkd3q0) ([raw file](https://github.com/CoreumFoundation/coreum/blob/c6874a9c5bd5161fb8a2967981e5bc323c771a89/proto/coreum/feemodel/v1/genesis.proto#L22)):*
<details><summary><i>Previously, masihyeganeh (Masih Yeganeh) wrote…</i></summary><blockquote>

Would you please tell me what these EMAs are?
</blockquote></details>

Exponential Moving Average (EMA) calculated over a period of time.
https://www.investopedia.com/terms/e/ema.asp#:~:text=The%20EMA%20is%20a%20moving,divergences%20from%20the%20historical%20average.
___
*[`x/delay/keeper/keeper.go` line 103 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTBx9LUAp0p2nKnX8z8:-OU9UyCwAg0T0X3Z-odd:b-bp51xt) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/x/delay/keeper/keeper.go#L103)):*

I see, so the same method is called inside DEX module when we:
- place order via API (delay should be registered)
- export genesis (delay shouldn't be registered)

Shall we handle it on DEX level instead ?
</blockquote></details>

Sure,
TODO: Consider moving the change to Dex module itself

masihyeganeh (approved):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: all files reviewed, 11 unresolved discussions (waiting on @miladz68, @TxCorpi0x, and @ysv)

___
*[`integration-tests/export/export_test.go` line 89 at r2](https://reviewable.io/reviews/CoreumFoundation/coreum/1162#-OTCMD_BEKxlSDDtP_KY:-OUEZ6m7ATWNBL_XxLOa:be6hqt9) ([raw file](https://github.com/CoreumFoundation/coreum/blob/d3eafc5b4a7a858a3af5d366769a7d7c87b90408/integration-tests/export/export_test.go#L89)):*
<details><summary><i>Previously, TxCorpi0x wrote…</i></summary><blockquote>

Agreed, It would be great to have a link (env or any better way) paths from crust to be usable within tests/integration-tests
</blockquote></details>

Why it is `05`? Can you walk me through it?

masihyeganeh (approved):
> *[Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1162)* status: all files reviewed, 10 unresolved discussions (waiting on @miladz68, @TxCorpi0x, and @ysv)

TxCorpi0x:
The export tests and fixes will be moved to another separate PRs

PR #1177, opened 2026-01-17
feat: enforce 5% minimum validator commission rate

trizzle5k:
This PR introduces a protocol-level requirement for a minimum validator commission rate of 5%. This change is designed to ensure network security and validator sustainability by preventing a "race to the bottom" in commission rates.

Changes:

Proto Definition: Updated StakingParams in params.proto to include min_commission_rate using cosmossdk.io/math.LegacyDec.

Validation: Added validateMinCommissionRate in params.go to enforce the 0.05 (5%) floor during parameter updates and ValidateBasic checks.

Defaults: Updated DefaultStakingParams to set the initial floor to 5%.

Note on Code Generation: I have manually updated the .proto and logic files. However, make generate failed in my environment due to a compatibility issue with github.com/bytedance/sonic (GoMapIterator error). I am submitting the source changes so the CI or a maintainer with a compatible build environment can sync the generated .pb.go files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/CoreumFoundation/coreum/1177)

### DIFF
--- a/proto/coreum/customparams/v1/params.proto
+++ b/proto/coreum/customparams/v1/params.proto
@@ -1,23 +1,2 @@
-syntax = "proto3";
-package coreum.customparams.v1;
 
-import "gogoproto/gogo.proto";
 
-option go_package = "github.com/CoreumFoundation/coreum/v6/x/customparams/types";
-
-// StakingParams defines the set of additional staking params for the staking module wrapper.
-message StakingParams {
-  // min_self_delegation is the validators global self declared minimum for delegation.
-  string min_self_delegation = 1 [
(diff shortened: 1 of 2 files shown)


omaidf:
The `CreateValidator` wrapper function only enforces the minimum self delegation requirement but completely ignores the `MinCommissionRate` parameter. This is a critical business logic gap - validators can still create validators with 0% commission rate, defeating the entire purpose of the PR which is to enforce a 5% minimum commission rate.

The code retrieves the custom params including `MinCommissionRate` at line 35, but only uses `MinSelfDelegation` for validation. The commission rate from `msg.Commission.Rate` is never checked against `params.MinCommissionRate`.

